PR #15, opened 2022-02-07
Docker: Local development setup

ababic:
With the exception of swapping out dependency management to Poetry, this is pretty much exactly how things are done on Etna.

Commit messages should be self-explanatory. Please do shout if you're unsure about anything.


lozette (commented):
This works well for me locally, however the CI tests aren't passing and I'm not sure why. Happy to pair on this if you need me to.

PR #33, opened 2022-02-15
Import Structured Search template from flask app

LornaHarwood:
This template is linked from the Basic Search Form template and was orginally missed from the previous template import work which should have included this import.


Floppy (approved):
I assume there's wiring up to do, but for the import, this looks fine to me.

PR #34, opened 2022-02-15
Import Results Template

LornaHarwood:
Import Results template and all relating templates

### DIFF
--- a/ds_judgements_public_ui/templates/includes/pagination.html
+++ b/ds_judgements_public_ui/templates/includes/pagination.html
@@ -1,0 +1,74 @@
+<nav class="pagination" aria-label="Results pagination">
+  <h3>Results pagination</h3>
+  <ul class="pagination__list">
+    <li class="pagination__list-item">
+      <span class="pagination__page-chevron-previous--disabled">
+        <span>
+          Previous page
+        </span>
+      </span>
+    </li>
+    <li class="pagination__list-pages">
+      <ol>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link-current" href="#" aria-label="Current page, Page 1"
+            aria-current="true"><span>Current page </span>1
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 2">
+            <span>Page </span>2
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 3">
+            <span>Page </span>3
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 4">
+            <span>Page </span>4
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 5">
+            <span>Page </span>5
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 6">
+            <span>Page </span>6
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 7">
+            <span>Page </span>7
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 8">
+            <span>Page </span>8
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 9">
+            <span>Page </span>9
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 10">
+            <span>Page </span>10
+          </a>
+        </li>
+        <li class="pagination__list-item">
+          <a class="pagination__page-link" href="#" aria-label="Go to page 11">
+            <span>Page </span>11
+          </a>
+        </li>
+      </ol>
+    </li>
+    <li class="pagination__list-item">
+      <a class="pagination__page-chevron-next" href="#"><span>Next page</span></a>
+    </li>
+  </ul>
+</nav>
--- a/ds_judgements_public_ui/templates/includes/results_applied_filters.html
+++ b/ds_judgements_public_ui/templates/includes/results_applied_filters.html
@@ -1,0 +1,38 @@
+{# {% if form.search_term.data or (form.collections.data != 'all') or (form.courts.data != 'all') %} #}
+  <h2 class="results-search-component__sub-header">Filters include</h2>
+  <ul class="results-search-component__removable-options">
+    {# {% if form.search_term.data %} #}
+      <li>
+        <span class="results-search-component__removable-options-key">Search term:</span>
+        <span class="results-search-component__removable-options-value">
+          <span class="results-search-component__removable-options-value-text">{# {{ form.search_term.data }} #}</span>
+          <a class="results-search-component__removable-options-link" href="{# {{ url_for('results', search_term='', collections=form.collections.data, courts=form.courts.data) }} #}">
+            <span class="results-search-component__removable-options-sr-only">Remove</span>
+          </a>
+        </span>
+      </li>
+    {# {% endif %} #}
+    {# {% if form.collections.data != 'all' %} #}
+      <li>
+        <span class="results-search-component__removable-options-key">Collection:</span>
+        <span class="results-search-component__removable-options-value">
+          <span class="results-search-component__removable-options-value-text">{# {{ form.collections.data }} #}</span>
+          <a class="results-search-component__removable-options-link" href="{# {{ url_for('results', search_term=form.search_term.data, collections='all', courts=form.courts.data) }} #}">
+            <span class="results-search-component__removable-options-sr-only">Remove</span>
+          </a>
+        </span>
+      </li>
+    {# {% endif %} #}
+    {# {% if form.courts.data != 'all' %} #}
+      <li>
+        <span class="results-search-component__removable-options-key">Courts:</span>
+        <span class="results-search-component__removable-options-value">
+          <span class="results-search-component__removable-options-value-text">{# {{ form.courts.data }} #}</span>
+          <a class="results-search-component__removable-options-link" href="{# {{ url_for('results', search_term=form.search_term.data, collections=form.collections.data, courts='all') }} #}">
+            <span class="results-search-component__removable-options-sr-only">Remove</span>
+          </a>
+        </span>
+      </li>
+    {# {% endif %} #}
+  </ul>
+{# {% endif %} #}
--- a/ds_judgements_public_ui/templates/includes/results_search_component.html
+++ b/ds_judgements_public_ui/templates/includes/results_search_component.html
@@ -1,0 +1,11 @@
+<div class="results-search-component">
+  <form action="{% url 'results' %}" id="analytics-results-filters">
+
+    <div class="results-search-component__container">
+      {% include 'includes/results_search_field.html' %}
+      {% include 'includes/results_applied_filters.html' %}
+    </div>
+
+    {% include 'includes/results_filters.html' %}
+  </form>
+</div>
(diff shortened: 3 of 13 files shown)


config/urls.py
```diff
@@ -22,6 +22,11 @@
         TemplateView.as_view(template_name="pages/sources.html"),
         name="sources",
     ),
+    path(
+        "results",
+        TemplateView.as_view(template_name="pages/results.html"),
+        name="results",
+    ),
```
lozette, line 29: I think because this route deals with a page built from data, that it should be in the `judgments` app (in `urls.py` in that app). Add a dummy method to `views.py` to hook the route onto.

In urls.py:
`path("results", views.results, name="results"),`

In views.py:
```
def results():
    pass
```

lozette (changes requested):

lozette (dismissed):

judgments/urls.py
```diff
@@ -4,8 +4,9 @@
 
 urlpatterns = [
     path("", views.index, name="index"),
-    path("<page>", views.index, name="index"),
+    path("(?P<page>\d)", views.index, name="index"),
```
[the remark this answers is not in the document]
LornaHarwood: Thanks!

ds_judgements_public_ui/templates/includes/results_search_component.html
```diff
@@ -0,0 +1,11 @@
...
+  <form action="{% url 'results' %}" id="analytics-results-filters">
+
+    <div class="results-search-component__container">
+      {% include 'includes/results_search_field.html' %}
+      {% include 'includes/results_applied_filters.html' %}
+    </div>
+
+    {% include 'includes/results_filters.html' %}
+  </form>
+</div>
```
ababic, line 11: Hi @LornaHarwood. We have `crispy-forms` installed on this project, which should allow us to avoid a lot of this custom template work. We should be aiming to do as much of this in python as possible: 
https://django-crispy-forms.readthedocs.io/en/latest/layouts.html

ds_judgements_public_ui/templates/includes/pagination.html
```diff
@@ -0,0 +1,74 @@
...
+            <span>Page </span>11
+          </a>
+        </li>
+      </ol>
+    </li>
+    <li class="pagination__list-item">
+      <a class="pagination__page-chevron-next" href="#"><span>Next page</span></a>
+    </li>
+  </ul>
+</nav>
```
ababic, line 74: Should this output be dynamic, based on what the API returns?

ds_judgements_public_ui/templates/includes/results_applied_filters.html
```diff
@@ -0,0 +1,38 @@
+{# {% if form.search_term.data or (form.collections.data != 'all') or (form.courts.data != 'all') %} #}
+  <h2 class="results-search-component__sub-header">Filters include</h2>
+  <ul class="results-search-component__removable-options">
+    {# {% if form.search_term.data %} #}
+      <li>
+        <span class="results-search-component__removable-options-key">Search term:</span>
+        <span class="results-search-component__removable-options-value">
+          <span class="results-search-component__removable-options-value-text">{# {{ form.search_term.data }} #}</span>
+          <a class="results-search-component__removable-options-link" href="{# {{ url_for('results', search_term='', collections=form.collections.data, courts=form.courts.data) }} #}">
```
ababic, line 9: I've got a little template tag I use for this kind of thing:

```
from django import template
from django.db.models import Model
from django.http.request import QueryDict

register = template.Library()

MODE_ADD = "__add"
MODE_REMOVE = "__remove"
MODE_TOGGLE = "__toggle"


@register.simple_tag(takes_context=True)
def querystring_modify(
    context, base=None, remove_blanks=False, remove_utm=True, **kwargs
):
    """
    Renders a URL and IRI encoded querystring (e.g. "q=Hello%20World&amp;category=1") that is safe to include in links.
    The querystring for the current request (``request.GET``) is used as a base by default, or an alternative
    ``QueryDict``, ``dict`` or querystring value can be provided as the first argument. The base value can be modified
    by providing any number of additional key/value pairs. ``None`` values are discounted automatically, and blank
    values can be optionally discounted by specifying ``remove_blanks=True``.

    When specifying key/value pairs, any keys that do not already exist in the base value will be added, and those
    that do will have their value replaced. Specifying a value of ``None`` for an existing item will result in it being
    discounted. For example, if the querystring for the current request were "foo=ORIGINALFOOVAL&bar=ORIGINALBARVAL",
    but you wanted to:

    * Change the value of "foo" to "NEWFOOVAL"
    * Remove "bar"
    * Add a new "baz" item with the value `1`

    You could do so using the following:

    ```
    {% load querystring_modify %}
    {% querystring_modify foo="NEWFOOVAL" bar=None baz=1 %}```
    ```

    The output of the above would be "?foo=NEWFOOVAL&amp;baz=1".

    Values can be strings, booleans, integers or references to other variables in the current context.
    For example, if the tag were being used to generate pagination links, where the page number
    was a variable named ``page_num``, you could reference that value like so:

    ```
    {% load querystring_modify %}
    {% querystring_modify p=page_num %}
    ```

    You can also specify more than one value for a key by providing an iterable as a value. For example, if the context
    contained a variable ``tag_list``, which was list of 'tag' values (```['tag1', 'tag2', 'tag3']```), you include all
    of those values by referencing the list value. For example:

    ```
    {% load querystring_modify %}
    {% querystring_modify tags=tag_list %}
    ```

    The output of the above would be "?tags=tag1&amp;tags=tag2&amp;tags=tag3" (plus whatever other values were in the
    base value).

    And finally, if there are potentially multiple values for a specific parameter,
    and you only want to add or remove a single one, you can add '__add' or '__remove'
    to the variable name in your list of modifier kwargs.

    For example, if the querystring was "tags=tag1&amp;tags=tag2&amp;tags=tag3", and you wanted to remove 'tag2', you
    could do:

    ```
    {% load querystring_modify %}
    {% querystring_modify tags__remove='tag2' %}
    ```

    Which would output: "?tags=tag1&amp;tags=tag3"

    Or, if you wanted to add a new 'tagNew' value to that same parameter, you could do:

    ```
    {% load querystring_modify %}
    {% querystring_modify tags__add='tagNew' %}
    ```

    Which would output: "?tags=tag1&amp;tags=tag2&amp;tags=tag3&amp;tags=tagNew"

    Similarly, if you're unsure whether the value is present or not, and would like to remove it if it is,
    or add it if it is not, you can also use the '__toggle' option. For example:

    ```
    {% load querystring_modify %}
    {% querystring_modify tags__toggle='tagNew' %}
    ```

    You can add as many modifiers to the same tag as you need to, with any
    combination of modifiers at once. For example, the following is perfectly valid:

    ```
    {% load querystring_modify %}
    {% querystring_modify page=None tags__add="newTag" tags__remove="oldTag" %}
    ```

    Modifiers always fail gracefully if the value you're trying to add is already
    present, or a value you're trying to remove is not, or the named parameter isn't
    present at all.
    """
    querydict = get_base_querydict(context, base)
    for key, value in kwargs.items():

        if isinstance(value, Model):
            value = str(value.pk)
        elif not hasattr(value, "__iter__"):
            value = str(value)

        if key.endswith(MODE_TOGGLE):
            key = key[: -len(MODE_TOGGLE)]
            values = set(querydict.get_list(key))
            if value in values:
                values.remove(value)
            else:
                values.add(value)
            querydict.set_list(key, list(values))

        elif key.endswith(MODE_ADD):
            key = key[: -len(MODE_ADD)]
            values = set(querydict.get_list(key))
            if value not in values:
                values.add(value)
                querydict.set_list(key, list(values))

        elif key.endswith(MODE_REMOVE):
            key = key[: -len(MODE_REMOVE)]
            values = set(querydict.get_list(key))
            if value in values:
                values.remove(value)
                querydict.set_list(key, list(values))

        elif value is None:
            querydict.pop(key, None)
        else:
            if isinstance(value, (str, bytes)):
                querydict[key] = value
            elif hasattr(value, "__iter__"):
                querydict.setlist(key, list(value))

    clean_querydict(querydict, remove_blanks, remove_utm)

    return f"?{querydict.urlencode()}"


def get_base_querydict(context, base):
    if base is None and "request" in context:
        return context["request"].GET.copy()
    if isinstance(base, QueryDict):
        return base.copy()
    if isinstance(base, dict):
        return QueryDict.fromkeys(base, mutable=True)
    if isinstance(base, str):
        return QueryDict(base, mutable=True)
    # request not present or base value unsupported
    return QueryDict("", mutable=True)


def clean_querydict(querydict, remove_blanks=False, remove_utm=True):
    remove_vals = {None}
    if remove_blanks:
        remove_vals.add("")

    if remove_utm:
        for key in tuple(querydict.keys()):
            if key.lower().startswith("utm_"):
                querydict.pop(key)

    for key, values in querydict.lists():
        cleaned_values = [v for v in values if v not in remove_vals]
        if cleaned_values:
            querydict.setlist(key, cleaned_values)
        else:
            del querydict[key]
```

ds_judgements_public_ui/templates/includes/pagination.html
ababic: Django has some built-in mechanisms for pagination, which we could use here (with a bit of bending). Happy to pair program some time next week if you'd like to explore?

ds_judgements_public_ui/templates/includes/results_search_component.html
ababic: On second thoughts, maybe there's no way to avoid dedicated templates for the level of customisation being made from a UI perspective - if it's only applicable to the search form, and the fields/filters are pretty stable, shifting over to `crispy-forms` might not be worth it.

ds_judgements_public_ui/templates/includes/pagination.html
gtvj: Yes, @ababic - you're right. It should be dynamic to reflect the current page. @LornaHarwood - when making this dynamic we need to ensure we're keeping the `aria-label` and `aria-current` properties and values represented in the static version.

ds_judgements_public_ui/templates/includes/results_search_component.html
gtvj: I don't have a particularly strong view on whether the custom template approach we've used in Flask could be achieved in a more streamlined way using `crispy-forms` in Django but I'd probably opt for the approach that will be most time efficient and easiest for people to work with in the short term.

ds_judgements_public_ui/templates/includes/results_applied_filters.html
ababic: This would ideally live in something like `core/templatetags/querystring_modify.py` (`core` being a new app for generic things like this, that don't really fit into a single app.) The `core` app would need a blank `__init__.py` to make it an 'app', and `core` would need to be added to the `INSTALLED_APPS` setting in order for the templatetags to be loaded.

judgments/tests.py
```diff
@@ -1,13 +1,17 @@
+from unittest import skip
+
 from django.test import TestCase
 
 
 class TestJudgment(TestCase):
+    @skip
```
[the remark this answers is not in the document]
Floppy: If that's the case, that's fine, btw, we can wire it back up in a followup.

Floppy (approved):

judgments/tests.py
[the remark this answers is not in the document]
lozette: We don't have any dynamic content yet so I thought it best to advise we skip these until we have it all working "for real"

ds_judgements_public_ui/templates/includes/results_search_component.html
Floppy: We could take either approach, but let's treat it as a separate "swap to crispy forms" task later on, if we want to.

ds_judgements_public_ui/templates/includes/pagination.html
Floppy: The dynamic things will be added in later, this PR is just bringing the template across to where we can start implementing that. @ababic the notes you've added here will be very helpful when we do that, thanks 👍🏻

ds_judgements_public_ui/templates/includes/results_applied_filters.html
Floppy: Sounds like a good refactoring for a followup PR sometime. Nice one.

PR #36, opened 2022-02-16
Update Judgment name

lozette:
A simple proof-of-concept form to update a Judgment's name in its metadata.

This commit adds an admin app (`judgment-admin`) with an edit & update method
for Judgments. A judgment can be edited by visiting `/judgments_admin/{uri}`.
The edit form only exposes the metadata name  (Xpath: `//akn:FRBRname/@value`)
for editing.

When the form is submitted, the judgment XML is updated, and the XML document
is sent to Marklogic via a PUT method in the API client.

### DIFF
--- a/judgments/api_client.py
+++ b/judgments/api_client.py
@@ -96,43 +96,39 @@ class MarklogicApiClient:
         self,
         method: str,
         path: str,
-        get_multipart: False,
+        headers: Dict[str, Any],
         data: Dict[str, Any] = None,
         body=None,
     ) -> requests.Response:
         kwargs = self.prepare_request_kwargs(method, path, body, data)
-        if get_multipart:
-            headers = {"Accept": "multipart/mixed"}
-        else:
-            headers = {"Accept": "text/xml"}
-        if method == "PUT":
-            headers["Content-Type"] = "application/xml"
         self.session.headers = headers
         response = self.session.request(method, **kwargs)
         # Raise relevant exception for an erroneous response
         self._raise_for_status(response)
         return response
 
-    def GET(self, path: str, get_multipart: bool, **data: Any) -> requests.Response:
-        return self.make_request("GET", path, get_multipart, data)
+    def GET(self, path: str, headers: Dict[str, Any], **data: Any) -> requests.Response:
+        return self.make_request("GET", path, headers, data)
 
-    def POST(self, path: str, **data: Any) -> requests.Response:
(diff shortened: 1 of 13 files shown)


judgments/api_client.py
```diff
@@ -86,16 +88,25 @@ def prepare_request_kwargs(
...
+        body=None,
     ) -> requests.Response:
-        kwargs = self.prepare_request_kwargs(method, path, data)
+        kwargs = self.prepare_request_kwargs(method, path, body, data)
         if get_multipart:
             headers = {"Accept": "multipart/mixed"}
         else:
             headers = {"Accept": "text/xml"}
+        if method == "PUT":
+            headers["Content-Type"] = "application/xml"
```
ababic, line 109: @lozette If we find ourselves setting headers for each type of request, it might be better to tackle this in a different way. For example, maybe `make_request()` could accept a `headers` keyword argument, and the individual methods specify the extra headers they need?

judgments_admin/views.py
```diff
@@ -0,0 +1,48 @@
...
+        name = xml.xpath(
+            "//akn:FRBRname",
+            namespaces=akn_namespace,
+        )[0]
+        name.set("value", request.POST["metadata_name"])
+        api_client.save_judgment_xml(request.POST["uri"], xml)
+        context["metadata_name"] = xml.xpath(
+            "//akn:FRBRname/@value",
+            namespaces=akn_namespace,
+        )[0]
```
ababic, line 43: It's early days, so this is probably fine... Just wondered if it's worth thinking about what it the responsibility of the api client, and what is the responsibility of the view. Maybe there's no need for the views to know anything about XML or the underlying structure? If that's a detail of how the client interacts with marklogic, maybe that should be taking care of value prep?

ababic:
@lozette Just wanted to also check that we have tickets in place for exploring how authentication will work for these views? From experience, it's usually best to start by assuming that only `superusers` in Django (`request.user.is_superuser is True`) should be able to use functionality, then figure out how to relax that requirement at a later time.

ababic:
@lozette More info here on protecting views based on the user: https://docs.djangoproject.com/en/4.0/topics/auth/default/#limiting-access-to-logged-in-users-that-pass-a-test

Floppy:
@ababic this will, in time, be split off into a separate microservice app that's fully auth-protected 👍🏻

Floppy (approved):
LGTM!

PR #42, opened 2022-02-24
Bring styles and javascript into django

gtvj:
This PR relates to the Trello Ticket "[Bring all SASS/CSS and JavaScript into Django](https://trello.com/c/cSdFrchX)" 

The tasks are: 

- [x] Bring SASS/CSS across
- [x] Populate title tag (using a configuration value for the service name)
- [x] Fix breadcrumbs
- [x] Introduce WebFonts that are used in the Flask app
- [x] Introduce level 1 heading in search results page
- [x] Add jQuery as dev dependency for Node
- [x] Add service name to heading on home page
- [x] Add updated header to HTML for static pages (Terms of use and Open Justice License)


lozette:
@gtvj A tip: run `pre-commit run --all-files` to run the linter locally, it should auto-fix any errors!

gtvj:
The current approach to page titles and breadcrumbs feels pretty sub-optimal (just thinking from the longer term maintenance perspective). It feels like: 

- Page titles should be passed through to the template from the view, and the value should be shared by both the `<h1>` and interpolated into the `<title>` (with "- Find case law" appended)
- "Find case law" should be a configuration value - it's currently replicated across multiple templates twelve times
- The `page-header__breadcrumb` component should be extracted to an include that can be shared across templates (that makes use of the "Find case law" configuration value and Page title, conditionally showing the third item). 

I did consider having a go at this but found `TemplateView.as_view` wasn't keen on taking variables.

Would be good to know what @lozette @LewisDaleUK @Floppy @LornaHarwood think about these proposals (they could be picked up in a separate PR / Trello ticket if everyone agrees).

lozette:
@gtvj Happy for have those bits put into a new card! We definitely need to configure some general site settings.

lozette (approved):

gtvj:
Thanks @lozette - I've created a card in Trello here: https://trello.com/c/hq8jIu50

PR #43, opened 2022-02-24
Add pagination

lozette:
Trello: https://trello.com/c/poU0fDFw/128-dev-pagination

Add a pagination method to the Judgment view. Because we're not using standard Django models, we have had to roll our own paginator.

Use this paginator in the index & search views.

### DIFF
--- a/ds_judgements_public_ui/templates/includes/pagination.html
+++ b/ds_judgements_public_ui/templates/includes/pagination.html
@@ -2,73 +2,49 @@
   <h3>Results pagination</h3>
   <ul class="pagination__list">
     <li class="pagination__list-item">
-      <span class="pagination__page-chevron-previous--disabled">
-        <span>
-          Previous page
-        </span>
+      {% if context.paginator.has_prev_page %}
+        <a class="pagination__page-chevron-previous"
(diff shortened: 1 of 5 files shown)


ds_judgements_public_ui/templates/includes/pagination.html
```diff
@@ -2,73 +2,49 @@
   <h3>Results pagination</h3>
   <ul class="pagination__list">
     <li class="pagination__list-item">
-      <span class="pagination__page-chevron-previous--disabled">
-        <span>
-          Previous page
-        </span>
+      {% if context.paginator.prev_page %}
+        <a class="pagination__page-chevron-previous"
+           href="{{request.path}}?{% if context.query %}query={{context.query}}&{% endif %}page={{context.paginator.current_page|add:'-1'}}">
```
LewisDaleUK, line 7: I imagine that the `current_page|add` syntax will be used every time we use the Paginator - it might be worth modifying the paginator output so it has a `prev_page` index we can just output? Maybe change the current `prev_page` to `has_prev_page` and then `prev_page=current_page-1`? (and the same again for next page)

lozette: Addressed!

LewisDaleUK (approved):

PR #44, opened 2022-02-24
Apply BasicAuth to Application

LornaHarwood:
Used wsgi protocol package [https://pypi.org/project/wsgi-basic-auth/](https://pypi.org/project/wsgi-basic-auth/)

Set environment variable WSGI_AUTH_CREDENTIALS=foo:bar with username and password credentials.


LewisDaleUK (approved):
Nice find with that middleware library 👍🏻

PR #58, opened 2022-03-01
(Spike) Search with XQuery

lozette:
Add an `advanced search` endpoint to search Marklogic with XQuery

Example search: `http://localhost:3000/judgments/advanced_search?query=&court=uksc&party=&judge=&page=1&order=-date&to=2020-01-01`

`order` can be `date` or `-date` (ascending & descending). If no order is passed the results come back in a rank order.
`from` and `to` should be in `YYYY-mm-dd` format

Note there's an update to the README showing how to add indexes to Marklogic, which is needed for the sorting & date search to work.

### DIFF
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -135,7 +135,7 @@ def results(request):
         query = params.get("query")
         page = params.get("page") if params.get("page") else "1"
         if query:
-            results = api_client.search_judgments(query, page)
+            results = api_client.basic_search(query, page)
             if type(results) == str:  # Mocked WebLogic response
                 xml_results = xmltodict.parse(results)
                 total = xml_results["search:response"]["@total"]
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -67,7 +67,7 @@ def detail(request, judgment_uri):
     return HttpResponse(template.render({"xml": judgment_xml}, request))
 
 
-def eval(request):
+def advanced_search(request):
     params = request.GET
     query = params.get("query")
     court = params.get("court")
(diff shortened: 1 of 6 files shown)


judgments/views.py
```diff
@@ -67,6 +67,45 @@ def detail(request, judgment_uri):
...
+        model = SearchResults.create_from_string(multipart_data.parts[0].text)
+
+        context["search_results"] = [
+            SearchResult.create_from_node(result) for result in model.results
+        ]
+        context["total"] = model.total
+        context["paginator"] = paginator(int(page), model.total)
+        context[
+            "query_string"
+        ] = f'query={str(query or "")}&court={str(court or "")}&party={str(party or "")}&judge={str(judge or "")}'
```
LewisDaleUK, line 102: This is a minor comment and can come as a later refactor, but to make this a bit easier to extend I'd probably put the various parameters into a dict at the top, like this:
```python
search_params = {
  "query": params.get("query"),
  "court": params.get("court"),
  "party": params.get("party"),
  "judge": params.get("judge") 
}
```

Then we can combine `join` and Python's list comprehension to build out the query string:

```python
context["query_string"] = "&".join([f'{key}={query_params[key]}' for key in query_params if query_params[key] is not None])
```

That way, when we want to add new search parameters we can just add them to the `query_params` dict.


I don't think this needs changing now, just wanted to leave this comment so we can refer back to it later if & when we're doing some refactoring

LewisDaleUK (approved):

PR #62, opened 2022-03-03
Cookie consent implementation

gtvj:
This PR brings across the cookie consent approach used for other TNA sites. 

- [x] Introduce front end (HTML, CSS and JavaScript) - clearly indicating this code is from a separate repository
- [x] Introduce the back end logic so that the banner and other relevant code is *not included* in the HTML payload if the user has not consented (this will need input from back end developer in Digital Services who can use the Discovery implementation as a reference. I've suggested pairing between @LewisDaleUK and a TNA back end developer - and I'm very happy to be part of this too)

- [ ] Deploy to a testing site on `nationalarchives.gov.uk` to test implementation is working as expected

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -137,6 +137,7 @@ MIDDLEWARE = [
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.common.BrokenLinkEmailsMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "judgments.middleware.CookieConsentMiddleware",
 ]
 
 # STATIC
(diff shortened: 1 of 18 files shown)


gtvj:
> Hi @gtvj
> 
> Thanks for implementing this.
> 
> The process you've done is correct. However on other environments, the built JS and CSS file from `ds-cookie-consent` was imported into straight the environment, instead of the source files. I think using the source files has its benefits (as there are usually environment specific CSS issues that need addressing, for example) but it would then make it hard to update cookie consent across all environments in the future if we've introduced environment specific code.
> 
> When I run the project the CSS looks slightly off, but I'm guessing this is because this repo isn't using `Roboto Mono` or bootstrap?
> 
> image

Thanks @AshTNA - in answer to your comments:

* I did consider bringing across the compiled JavaScript but I was a little reluctant to include the compiled JS straight from another repository. For a system like caselaw it seems particularly important that the source can easily be scrutinised (which is very difficult when working with compiled JS)
* You're right - things look a little different because caselaw isn't using Roboto Mono or bootstrap

gtvj:
This work has moved over to a new PR #155

PR #77, opened 2022-03-11
Post data integration CSS fixes and cross-browser testing

gtvj:
This PR is focussed on:

* on fixing issues that have emerged following the integration with MarkLogic
* performing cross-browser testing of the application pages against the browsers and devices listed in [The Service Manual](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2021).

The PR does not delve into Judgment text because (1) the parser is in the process of being updated, and (2) only one Judgment is available via local development at present

## Fixes

- [x] Fix logo alignment 
- [x] Fix phase banner focus state contrast
- [x] Fix breadcrumb focus state contrast
- [x] Add JavaScript to manage keyboard events on ARIA buttons
- [x] Fix JavaScript error in IE11 console (WebPack/Babel are including an arrow function in the compiled `app.js`)
- [x] Fix results fieldset rendering in IE11
- [x] Fix structured search panel rendering in IE11
- [x] Cross browser fixes (see table below)

## Cross browser testing

OS | Browser | Tested and fixes applied
-- | -- | --
Windows | Internet Explorer 11 | Tested 
  | Edge (latest versions) |  Tested in Edge 97 and 98
  | Google Chrome (latest versions) | Tested in Chrome 99 and 98
  | Mozilla Firefox (latest versions) | Tested in Firefox 97 and 96
macOS | Safari 12 and later | Tested in 12, 13, 14, and 15
  | Google Chrome (latest versions) | Tested in 98 and 99
  | Mozilla Firefox (latest versions) | Tested in 97 and 96
iOS | Safari for iOS 12.1 and later | Tested in 12.0, 13.0, 14.1 
  | Google Chrome | Tested in latest version (only one version available via LambdaTest)
Android | Google Chrome (latest versions) | Tested latest version (only one version available via LambdaTest)
  | Samsung Internet (latest versions) | Not tested - not available via LambdaTest


ds_judgements_public_ui/sass/includes/_search.scss
```diff
@@ -38,7 +38,7 @@
   &__search-term-input {
     @include text_field;
     width: 80%;
-    min-height: calc($spacer__unit * 3);
+    min-height: calc($spacer__unit * 2);
```
lozette, line 41: I have a question about this: in Chrome, I was finding that the $spacer__unit was not being interpolated correctly, so that all the instances of `calc()` containing $spacer__unit were failing to be calculated. I don't know loads about SASS so I had a google, and apparently it needs to be something like `calc(#{spacer__unit} * 2)` - is this correct in your experience?

gtvj: Thanks @lozette - I think this is the right syntax for `calc()` but when I saw your question I thought I might be out of date. I've just checked the Sass documentation and it does seem it's better not to avoid interpolation in calls to `calc()`. [This page](https://sass-lang.com/documentation/values/calculations) says:

> _You can also use interpolation in a calculation. However, if you do, nothing in the parentheses that surround that interpolation will be simplified or type-checked, so it’s easy to end up with extra verbose or even invalid CSS. Rather than writing `calc(10px + #{$var})`, just write `calc(10px + $var)`!_

In this case, the compiled CSS should have `min-height: 2rem` which seems to be what's generated by both `@debug` statements and the output. Is the field looking strange?

lozette: I'm sorry my local is completely broken and I don't know how soon I can fix it, but IIRC Chrome isn't able to interpret anything where $spacer__unit is inside a calc(), and as a result things which are meant to have any spacing/padding using a calculated $spacer__unit is just not having any padding/spacing. There's an error in Chrome saying invalid declaration, I think.

lozette: Screenshot 2022-03-16 at 15 30 42

Finally got my local working again! Here's an example, if this makes sense.

gtvj: That's really strange. My Chrome console is fine (screenshot below). To be honest, I feel like it's fine - the `.scss` file is only available via a source map and isn't used by the browser to render. As long as the compiled CSS is fine and has the correct calculated properties it shouldn't pose a problem. I'll created a [Trello card](https://trello.com/c/hZX5RpzI) to take a closer look at this at a later date - we're using variables within `calc()` throughout the Sass. 

Screenshot 2022-03-17 at 10 32 15

lozette (approved):

PR #91, opened 2022-03-17
Add a changelog

lozette:
Once we release the app to staging, we should start using a changelog to record releases and their changes.

TODO: Add release process and start using this changelog once we begin releasing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog 1.0.0].
+
+## [Unreleased]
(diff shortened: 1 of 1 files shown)


LewisDaleUK (approved):

CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog 1.0.0].
```
LewisDaleUK, line 5: I think this is missing the URL

lozette: It's linked down below, so this link should work (or it does for me!)

PR #99, opened 2022-03-22
Feature/download as pdf

LewisDaleUK:


judgments/views.py
```diff
@@ -111,6 +115,27 @@ def detail_xml(_request, judgment_uri):
...
+        return super(PdfDetailView, self).dispatch(request, *args, **kwargs)
+
+    def get_context_data(self, judgment_uri, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        results = api_client.eval_xslt(judgment_uri)
+        multipart_data = decoder.MultipartDecoder.from_response(results)
+        context["judgment"] = multipart_data.parts[0].text
+
+        print(context["judgment"])
```
lozette, line 135: Remove me!

ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
```diff
@@ -8,6 +8,7 @@
         </div>
         <div class="judgment-toolbar__download">
             <a class="judgment-download__option--xml" href="{% url 'detail_xml' context.judgment_uri %}">{% translate "judgment.downloadasxml" %}</a>
+            <a class="judgment-download__option--xml" href="{% url 'detail_pdf' context.judgment_uri %}">{% translate "judgment.downloadaspdf" %}</a>
```
lozette, line 11: I think this class should end with `--pdf` and not `--xml`

package.json
```diff
@@ -7,8 +7,8 @@
     "doc": "docs"
   },
   "scripts": {
-    "start-sass": "sass --watch ds_judgements_public_ui/sass/main.scss:ds_judgements_public_ui/static/css/main.css",
-    "start-scripts": "npx webpack --watch"
+    "start-sass": "sass --watch ds_judgements_public_ui/sass:ds_judgements_public_ui/static/css",
+    "start-scripts": "npx webpack --watch",
```
gtvj, line 11: The trailing comma here is causing the `package.json` to throw a parsing error.

gtvj:
Hi @LewisDaleUK - the front end stuff here seems fine to me. I wouldn't usually use absolute units in CSS but since this is for the PDF output only I've found in other projects where we've done this that it's fine.

lozette (approved):

PR #100, opened 2022-03-22
Install curl to our Docker image

olivia-work:
Dalmatian needs curl to be installed so that it can instal aws-env on
start up so that environment variables can be set.


rjw1 (approved):
We are pairing on this.

LewisDaleUK (approved):

PR #101, opened 2022-03-23
Apply remove search filters functionality

LornaHarwood:
We have added a custom template filter to remove a key from the query
parameters and produce a new query string.


ds_judgements_public_ui/templates/includes/results_applied_filters.html
```diff
@@ -1,38 +1,17 @@
...
-  <ul class="results-search-component__removable-options">
-    {# {% if form.search_term.data %} #}
+{% load query_filters %}
+<h2 class="results-search-component__sub-header">Filters include</h2>
+<ul class="results-search-component__removable-options">
+  {% for key, value in context.query_params.items %}
+    {% if value %}
       <li>
-        <span class="results-search-component__removable-options-key">Search term:</span>
+        <span class="results-search-component__removable-options-key">{{ key }}:</span>
```
gtvj, line 7: HI @LornaHarwood - I wonder if the key should be capitalised (possibly using [capfirst](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#capfirst)

gtvj (changes requested):
Hi @LornaHarwood - this looks good. I've made a small suggestion about using a Django filter to capitalise a value. Let me know if this all seems OK from your perspective 👍🏽

gtvj (approved):

PR #102, opened 2022-03-23
Add new middleware that, when rendering a template using the Template…

LewisDaleUK:
…Response object, will set a flag on the context data called `showGTM`.

This flag will be true if both of these conditions are met:
  1) The `cookies_policy` cookie has been set
  2) The decoded JSON value of `cookies_policy` includes an attribute named `usage` that has been set to `True`

If the flag is set, and is True, then the Google Tag Manager analytics will be enabled. Otherwise, they are disabled.


gtvj:
Hi @LewisDaleUK - I'm getting a `KeyError: 'cookies_policy'` error when running this branch locally?

ds_judgements_public_ui/templates/includes/gtm/gtm_body.html
```diff
@@ -1,6 +1,8 @@
-<!-- Google Tag Manager (noscript) -->
-<noscript>
-    <iframe src="https://www.googletagmanager.com/ns.html?id=GTM-MMVM4CJ"
-            height="0" width="0" style="display:none;visibility:hidden"></iframe>
-</noscript>
-<!-- End Google Tag Manager (noscript) -->
+{% if showGTM %}
```
gtvj, line 1: Hi @LewisDaleUK - would it be OK to add a similar `if` condition to the `gtm_head.html` file?

gtvj (changes requested):

PR #107, opened 2022-03-24
Bugfix: Correct XQuery for publishing judgments

lozette:
This should be `set-property` not `set-properties` - the latter will replace
all properties, the former only adds one property.


LewisDaleUK:
It looks like this code is also duplicated in #108 - do we want to close this and just have the one PR?

PR #122, opened 2022-03-29
Implement /check endpoint to return "OK"

LornaHarwood:
WSGI uses WSGI_AUTH_EXCLUDE_PATHS environment variable to exclude any specific paths from basic auth.

Add WSGI_AUTH_EXCLUDE_PATHS =/check to `.env` file


LewisDaleUK (approved):

Floppy (approved):
Nice!

PR #123, opened 2022-03-30
Do not show non-web images on Judgments

lozette:
Some of the Judgmetns/decisions contain images which are not web formats, e.g.
.emf or .wmf

These formats are not viewable in a browser and cause an error when the
judgment is downloaded as a PDF.

Amend the XSL transform to only show web formats (png, gif, jpg) and ignore
any others.

Ideally, we will need Clerks to only submit images in judgments which are
web-compatible, or we will need to convert non-web images somewhow as part
of the ingestion service.

Note that the XSL transform will need to be re-uploaded to Marklogic as per the
README.


judgments/bootstrap/judgment2.xsl
```diff
@@ -257,10 +257,12 @@
 </xsl:template>
 
 <xsl:template match="img">
-	<img>
-		<xsl:apply-templates select="@*" />
-		<xsl:apply-templates />
-	</img>
+	<xsl:if test="matches(@src, '\.gif|\.png|\.jpg|\.jpeg')">
```
Floppy, line 260: Should we include `webp` and `svg` for completeness while we're here? SVG might actually be the best thing to convert them to in the long run, as WMF can be vector or bitmap.

Floppy:
```suggestion
	<xsl:if test="matches(@src, '\.gif|\.png|\.jpg|\.jpeg|\.webp|\.svg')">
```

Floppy (changes requested):

judgments/bootstrap/judgment2.xsl
lozette: Good call

Floppy (approved):

PR #125, opened 2022-03-31
Change banner to Alpha

gtvj:
This PR just changes the phase banner from 'Beta' to 'Alpha'. I've also removed the uppercasing from the HTML since this is being applied in the CSS anyway.


LornaHarwood (approved):
👍

PR #131, opened 2022-04-06
Add Atom feeds to judgment browse URLs

LewisDaleUK:
Adds an Atom XML feed for each of the browse urls. You can access the atom feeds by appending "/feed.xml" to the url, e.g. /ewca/civ/2017/feed.xml

### DIFF
--- a/judgments/feeds.py
+++ b/judgments/feeds.py
@@ -1,0 +1,55 @@
+import datetime
+
+from django.contrib.syndication.views import Feed
+from django.urls import reverse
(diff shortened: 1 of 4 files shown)


Floppy (approved):

Floppy:
Looks like it will do the job; we might want to tidy up to merge the HTML and Atom views into one, and use content negotiation, but I'm happy to let this through for now in the interests of speed (and me not knowing how to suggest how that refactor would work).

PR #139, opened 2022-04-07
Further front end changes

gtvj:
- [x] Introduce static implementation of browse by court on homepage


ds_judgements_public_ui/templates/includes/browse_by_court.html
```diff
@@ -5,10 +5,9 @@ <h2 id="browse-by-court" class="judgment-browse-by-court__header">Find judgments
 
       <ul class="judgment-browse-by-court__list">
         {% for court in courts %}
-        {# {% for court in courts | sort(attribute='name') %} #}
           <li class="judgment-browse-by-court__court">
             <span class="judgment-browse-by-court__court-name">
-                <a class="judgment-browse-by-court__court-link" href="/results?courts={{ court.code }}">
+                <a class="judgment-browse-by-court__court-link" href="/judgments/advanced_search?court={{ court.code }}">
```
LewisDaleUK, line 10: If you use the `url` template tag, this link won't break if (for any reason) our url scheme changes on the backend.

`href="{% url 'advanced_search' %}?court={{ court.code }}"`

judgments/courts.py
```diff
@@ -0,0 +1,104 @@
+courts = [
```
LewisDaleUK, line 1: Could we move this file into a subdirectory, called something like `fixtures`? Just to distinguish it from business-logic code that sits at the top-level. (You might need to make the subdirectory into a python module by add an empty `__init__.py` file)

ds_judgements_public_ui/templates/includes/browse_by_court.html
```diff
@@ -2,12 +2,12 @@
   <div class="judgment-browse-by-court">
     <div class="judgment-browse-by-court__container">
       <h2 id="browse-by-court" class="judgment-browse-by-court__header">Find judgments by court</h2>
-
+x§
```
lozette, line 5: Remove me!

lozette (changes requested):

ds_judgements_public_ui/templates/includes/browse_by_court.html
gtvj: Thanks - I've removed those now.

lozette (approved):

PR #166, opened 2022-04-13
Add copy for Terms of Use page

gtvj:
Adds HTML for the Terms of Use page (from the Word document provided earlier today).

### DIFF
--- a/ds_judgements_public_ui/templates/pages/terms_of_use.html
+++ b/ds_judgements_public_ui/templates/pages/terms_of_use.html
@@ -2,7 +2,7 @@
 {% load i18n %}
 
 {% block title %}
-{{ page_title }} - {% translate "common.findcaselaw" %}
+  {{ page_title }} - {% translate "common.findcaselaw" %}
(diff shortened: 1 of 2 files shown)


ds_judgements_public_ui/templates/pages/terms_of_use.html
```diff
@@ -14,76 +14,61 @@
...
-      suscipit. Assumenda consectetur dolores nemo nobis pariatur quae recusandae voluptatum? Laborum mollitia
-      quaerat quam ullam?
-    </div>
-    <div>
-      Consequuntur dignissimos ea excepturi necessitatibus non officia similique ut! Alias numquam officia
-      provident tempore vero! A amet consequatur dolorum magni nam. Ab earum eum fuga numquam quod repellat
-      voluptas! Rerum.
-    </div>
+
+    <h2>What is in these terms</h2>
```
LewisDaleUK, line 18: I think this text should be put through the translation service, so that it's added to the `en_GB/django.po` file

PR #167, opened 2022-04-13
Add initial explainer page content

pb4pixels:
Latest version of the explainer page content.

- [ ]  Need to add link to API documentation in the API section. 
- [ ] In page navigation / table of contents

PR #170, opened 2022-04-14
Add content for 'What to expect' page

gtvj:
This PR brings introduces the changes from #167 with some changes to the HTML, including: 

- [x] Introduction of `<section>` elements and some changes to structure and semantics
- [x] Changes to heading hierarchy
- [x] Introduction of additional semantic elements, such as `<q>` and `<abbr>`
- [x] Uses `translate` for common strings (such as email addresses) and `url` for routes

I've initially introduced this as HTML rather than via Django translations for the following reasons: 

1. This is a long page comprising: 111 paragraphs, 58 list items, 29 links, 24 abbreviations - which will generate a huge number (likely hundreds) of necessary translations. I suspect the benefit of these translations may never outweigh the effort necessary for them to be produced and maintained (it seems this content is best suited to a CMS rather than locale files)
2. The page is likely to need ongoing editing and refinement, which will be more straightforward to achieve if it is in HTML 

I'm happy for the page to use translations if others have a strong view on this but, for the time being, it seems most sensible to include the HTML.

### DIFF
--- a/ds_judgements_public_ui/templates/pages/what_to_expect.html
+++ b/ds_judgements_public_ui/templates/pages/what_to_expect.html
@@ -12,8 +12,550 @@
   </header>
 
   <div class="standard-text-template">
-  <h1>What to expect from this new service</h1>
-  <div class="entry-content">
-    <p>...</p>
+    <h1>What to expect from this new service</h1>
+    <p><small>Last updated on 13th April 2022</small></p>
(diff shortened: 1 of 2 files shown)


Floppy:
Yeah, that's huge. I'm OK with keeping the text out of the translation file for now; we can move it later easily, and it's all in one place, not scattered across the app. I can live with it.

PR #179, opened 2022-04-20
Fix atom feed validation errors by adding author and updated timestam…

LewisDaleUK:
…ps to each entry


lozette (approved):

judgments/feeds.py
```diff
@@ -10,6 +10,7 @@
 
 class LatestJudgmentsFeed(Feed):
     feed_type = Atom1Feed
+    author_name = "The National Archives'"
```
Floppy, line 13:
```suggestion
    author_name = "The National Archives"
```

Floppy (changes requested):

PR #186, opened 2022-04-26
Space out download buttons on smaller screens

symroe:
I'm not 100% of this implementation. I think I would use flexbox for the spacing, and use a uniform stack system where `margin-top` is applied to everything. However, using breakpoints like this does work, and uses the existing system.

## Before
Screenshot 2022-04-26 at 17-05-38 KICKABOUT PRODUCTIONS LIMITED v THE COMMISSIONERS FOR HER MAJESTY’S REVENUE AND CUSTOMS - Find case law

## After

Screenshot 2022-04-26 at 17-05-54 KICKABOUT PRODUCTIONS LIMITED v THE COMMISSIONERS FOR HER MAJESTY’S REVENUE AND CUSTOMS - Find case law

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
@@ -12,9 +12,17 @@
     display: inline-block;
     text-decoration: none;
     outline-offset: 3px;
+    @media (min-width: $grid__breakpoint-small) {
+      margin-top: $spacer__unit/2;
+    }
     &:first-child {
       margin-right: $spacer__unit;
+      @media (min-width: $grid__breakpoint-small) {
+        margin-right: 0;
+        margin-top: 0;
+      }
     }
+    
   }
 
   &__return-link {


gtvj:
Thanks for your contribution. I agree, flexbox would be better. Can you provide a bit more information about what you mean by "uniform stack system" - I'm not familiar with that term.

symroe:
Sure thing!

I'm a big fan of this system: https://every-layout.dev/layouts/stack/

In short, it applies a `margin-top` to each "element that comes after an element" (`* + *`). The ends up in a consistent vertical layout and means you can avoid some of the complications that a mix of paddings and margins can produce.

It's a larger refactor though, so not one I'm suggesting you implement now! Especially as I couldn't get the project working locally because of the Marklogic requirement (it would be nice to have some fixtures for mocking the `MarklogicApiClient` locally, and removing it as a hard requirement — I know it's early days in the project though, and that it's been a rush to get here, so not complaining!)

By the way, the "cluster" pattern on that side is what I was thinking about with flexbox: https://every-layout.dev/layouts/cluster/

Again though I didn't want to make too big of a change. I think what I've done will work with minimal disruption to the rest of the CSS!

symroe:
Looks like this was fixed in https://github.com/nationalarchives/ds-caselaw-public-ui/commit/2bab7daa6fee58ac958b15c9b98284b2e2614725

PR #204, opened 2022-05-06
Add sorting to search results, by relevance or date descending

lozette:
This is a temporary-ish fix until we get front-end design input.

Add Sort by date/Sort by relevance links to the search results.

The links re-write the request to alter the `order` query parameter. As such,
we have some code in the view to remove the `order` param and re-add it with
the relevant value.

We also have a tweak in the front end to not show the Order parameter in the
removable filter UI pattern.

Screenshot 2022-05-06 at 11 24 46
Screenshot 2022-05-06 at 11 24 20

### DIFF
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -2,6 +2,7 @@ import datetime
 import math
 import os
 import re
+import urllib
 
 from caselawclient.Client import (
     RESULTS_PER_PAGE,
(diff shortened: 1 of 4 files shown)


judgments/views.py
```diff
@@ -107,7 +107,11 @@ def advanced_search(request):
         context["paginator"] = paginator(int(page), model.total)
 
         context["query_string"] = "&".join(
-            [f'{key}={query_params[key] or ""}' for key in query_params]
+            [
+                f'{key}={query_params[key] or ""}'
+                for key in query_params
+                if key not in "order"
```
LewisDaleUK, line 113: I'm not sure what this line is doing - is it just to eliminate the `order` param from the query_string?

lozette: Yeah my thought was to remove it from the params so it could be overwritten, but I'm not sure if this is the best way!

Floppy: I feel like this is something that should be doable with a bit of standard code, like `URI.encode_www_form(hash)` in Ruby. Then you could just overwrite the order in the hash and convert. But I don't know the python equivalent.

Floppy: Ah, this is it: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode

Floppy (changes requested):

PR #223, opened 2022-05-23
Remove postgres

dragon-dxw:
Whilst this appears to be working successfully locally, I'm worried about the bigger implications and think it's worth talking about -- there's some challenges with the database migrations when there isn't a database...

Also, very untested in production mode, where there's a very different untested entrypoint...


dragon-dxw:
[Remove postgres, redis and email config from public UI app](https://trello.com/c/V2yauuR4/240-remove-postgres-redis-and-email-config-from-public-ui-app)

Floppy (changes requested):
Looks good to me; there might be a couple more things that need to change to take account of the lack of DB (and speficially the site settings table).

* should we remove the admin interface at https://github.com/nationalarchives/ds-caselaw-public-ui/blob/a00d3a7f07fac4cc42460185ee59662ae8439ed7/config/urls.py#L51 - if there's no database, presumably this will explode?
* I *think* that the site name gets used in https://github.com/nationalarchives/ds-caselaw-public-ui/blob/ae573d750e3a594ded4f72e41e7289ae5c99a37c/judgments/feeds.py#L14 (I think this is why I had to go set it. We should check that doesn't explode too, and provide a static name if it does.

PR #224, opened 2022-05-24
Remove emailers

dragon-dxw:
We don't need the email libraries (anymail, mailgun) which were installed by the Django quick-start pack we used.


dragon-dxw:
[Remove postgres, redis and email config from public UI app](https://trello.com/c/V2yauuR4/240-remove-postgres-redis-and-email-config-from-public-ui-app)

Floppy (approved):

PR #270, opened 2022-06-22
Order simple search results (from homepage) by date if there is a query

lozette:
On the simple search results (from the homepage), the search results should be
ordered by:

- Relevance if there is a query
- Date if there is not a query

Caveat: if an `order` param is passed in, we should respect it.

Trello: https://trello.com/c/81UiEmFs


dragon-dxw (approved):

dragon-dxw:
Interesting to note that order=-relevance and order=relevance give the same results...!

PR #281, opened 2022-06-29
Show a 404 for unpublished judgments

lozette:
## Changes in this PR:

A user hacking the URLs managed to stumble across a judgment which exists, but is not published. They complained about getting a 500 error.

This PR returns a 404 if a judgment is in Marklogic (exists) but is unpublished (unavailable). The first commit shows a 404 and "Judgment not found" whether the judgment doesn't exist or is unavailable. The second commit shows a slightly different message if the judgment exists and is unavailable.

We can drop the second commit if we just want to show "Judgment not found" no matter what. 

https://rollbar.com/dxw/tna-caselaw-public-ui/items/24/?item_page=0&#instances

https://trello.com/c/budavgUm

### After

### DIFF
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -74,7 +74,7 @@ def detail(request, judgment_uri):
         context["page_title"] = model.metadata_name
         context["judgment_uri"] = judgment_uri
         context["pdf_size"] = get_pdf_size(judgment_uri)
-    except MarklogicResourceNotFoundError:
+    except (MarklogicResourceNotFoundError, AttributeError):
         raise Http404("Judgment was not found")
     template = loader.get_template("judgment/detail.html")
     return TemplateResponse(request, template, context={"context": context})
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -65,19 +65,27 @@ def browse(request, court=None, subdivision=None, year=None):
 def detail(request, judgment_uri):
     context = {}
     try:
-        results = api_client.eval_xslt(judgment_uri)
-        xml_results = api_client.get_judgment_xml(judgment_uri)
-        multipart_data = decoder.MultipartDecoder.from_response(results)
-        judgment = multipart_data.parts[0].text
-        model = Judgment.create_from_string(xml_results)
-        context["judgment"] = judgment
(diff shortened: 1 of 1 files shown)


judgments/views.py
```diff
@@ -65,19 +65,27 @@ def browse(request, court=None, subdivision=None, year=None):
...
+            context["judgment"] = judgment
+            context["page_title"] = model.metadata_name
+            context["judgment_uri"] = judgment_uri
+            context["pdf_size"] = get_pdf_size(judgment_uri)
+        except MarklogicResourceNotFoundError:
+            raise Http404("Judgment was not found")
+        template = loader.get_template("judgment/detail.html")
+        return TemplateResponse(request, template, context={"context": context})
+    else:
+        raise Http404("This Judgment is not available")
```
dragon-dxw, line 88: There's a part of me going "shouldn't this be a 403?" but I think it's conceptually simpler for consumers iterating over all the numbers to get a 404.

judgments/views.py
```diff
@@ -65,19 +65,27 @@ def browse(request, court=None, subdivision=None, year=None):
...
-        xml_results = api_client.get_judgment_xml(judgment_uri)
-        multipart_data = decoder.MultipartDecoder.from_response(results)
-        judgment = multipart_data.parts[0].text
-        model = Judgment.create_from_string(xml_results)
-        context["judgment"] = judgment
-        context["page_title"] = model.metadata_name
-        context["judgment_uri"] = judgment_uri
-        context["pdf_size"] = get_pdf_size(judgment_uri)
-    except MarklogicResourceNotFoundError:
+        is_published = api_client.get_published(judgment_uri)
```
dragon-dxw, line 68: I think we're kinda fortunate that the page 500'd, because otherwise we'd be publishing unpublished pages... I don't know if there's any improvements we can make to make that less likely?

dragon-dxw (approved):

judgments/views.py
lozette: Yeah I was wondering that!! Definitely worth looking into why we missed that

PR #284, opened 2022-07-04
Add last page to paginator

dragon-dxw:
If there were 25 judgments, only two pages would be shown on the paginator
due to an off-by-one error: list(range(1,5)) returns [1, 2, 3, 4]. We
now add one to the maximum value.

This has the side effect of making page 20 visible from page 10 when there
are a large number of pages. This feels more like the correct behaviour
to me.

The 2000-page pagination test has been increased to 2500 to test behaviour
when there are more judgements than the paginator wants to show; additional
tests now exist for 5 and 25 judgments to test smaller edge cases.

### Before

image
but also
image

### After

image

image
(note page 11... maybe not ideal...)

### DIFF
--- a/judgments/tests.py
+++ b/judgments/tests.py
@@ -116,17 +116,42 @@ class TestJudgmentModel(TestCase):
 
 
 class TestPaginator(TestCase):
-    def test_paginator(self):
+    def test_paginator_2500(self):
         expected_result = {
             "current_page": 10,
             "has_next_page": True,
             "has_prev_page": True,
             "next_page": 11,
             "prev_page": 9,
-            "next_pages": [11, 12, 13, 14, 15, 16, 17, 18, 19],
-            "number_of_pages": 200,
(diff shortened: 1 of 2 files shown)


judgments/tests.py
```diff
@@ -116,17 +116,42 @@ def test_can_parse_judgment(self):
 
 
 class TestPaginator(TestCase):
-    def test_paginator(self):
+    def test_paginator_2500(self):
```
lozette, line 119: This is the smallest of typos but it should be `250` not `2500`

lozette (approved):

judgments/tests.py
dragon-dxw: It's the number of judgments, not the number of pages, in each case.

dragon-dxw:
[Pagination on Editor does not show all page numbers, the last is always missing](https://trello.com/c/HiBwnQvH/525-pagination-on-editor-does-not-show-all-page-numbers-the-last-is-always-missing)

PR #303, opened 2022-07-22
Use context processor instead of middleware for cookie consent

dragon-dxw:
https://trello.com/c/qqdJ4fYZ/476-add-gtm-snippet-to-error-pages

With many, many thanks to Abdul Aziz Barkat for their clear
explanation of why this is happening under the hood and the
right approach to take:

https://stackoverflow.com/questions/73070115/why-is-the-middleware-not-processed-on-404-page-in-django/73070884#73070884

(summarised)

> Django calls the render function on the response from the exception handler making it not a TemplateResponse object anymore. 

> It seems you are using the middleware to add values to your context, which is not the best place to do this. You should instead write a [custom context processor](https://docs.djangoproject.com/en/4.0/ref/templates/api/#writing-your-own-context-processors) which will be called consistently.

This appears to work wonderfully with minimal changes.


dragon-dxw:
@LewisDaleUK You might be interested in this utterly mystifying bug and how to solve it, for your own knowledge: https://stackoverflow.com/questions/73070115/why-is-the-middleware-not-processed-on-404-page-in-django/73070884#73070884

gtvj:
Thanks so much @dragon-dxw for solving this one. Definitely an utterly mystifying one. 

cc. @mozaki1 @TNA-Allan @ACTNA - just to make you aware of the fix here because it's almost certainly something you'll encounter in Etna when trying to pass values to the context of 403, 404 and 500 pages (such as those we use to set the `showGtm` variable which, in turn, is used to include or exclude the Google Tag Manager snippets).

lozette (approved):

PR #308, opened 2022-08-03
Add breadcrumbs and back links on 404/403/500 pages

lozette:
## Changes in this PR:

The error pages in the application offered no way for a user to return
to the Find case law homepage. Add a breadcrumb & a link back to the main
site on these pages.

Trello: https://trello.com/c/vcqbPLxF

### After

Screenshot 2022-08-03 at 12 03 53
Screenshot 2022-08-03 at 12 03 44
Screenshot 2022-08-03 at 12 03 34

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_header.scss
+++ b/ds_judgements_public_ui/sass/includes/_header.scss
@@ -24,6 +24,7 @@
       display: inline-block;
       position: relative;
       margin-right: calc($spacer__unit / 2);
+      color: $color__white;
 
       &:not(:nth-child(1)) {
         padding-left: $spacer__unit;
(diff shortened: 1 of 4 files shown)


ds_judgements_public_ui/sass/includes/_header.scss
```diff
@@ -24,6 +24,7 @@
       display: inline-block;
       position: relative;
       margin-right: calc($spacer__unit / 2);
+      color: $color__white;
```
dragon-dxw, line 27: My assumption is this is necessary because the last item isn't a link, unlike every other breadcrumb (and it doesn't make sense to be a link, since 404 pages occur anywhere) -- is that right?

dragon-dxw (approved):

ds_judgements_public_ui/sass/includes/_header.scss
lozette: Before I added this, the text was black by default, which made it invisible on the header. I might have added it in the wrong place, though (I'm not great at CSS!)

PR #316, opened 2022-08-05
Change where we pick a judgement's date from in the model

lozette:
## Changes in this PR:

Back when we started developing this application, we weren't really aware of the
differences between `FRBRWork`, `FRBRManifestation` and `FRBRExpression`. When
picking the date to show when a judgment was handed down, we decided to choose
any `FRBRdate` element with an attribute of `judgment`.

As we have received more & more judgments, some of them have been displayed
in the public ui without a date. This is because not all the judgments have
an `FRBRdate/@name=judgment` in their XML.

Jim jurisdatum suggested that we look instead for ANY date in the `FRBRWork`
block, and use that as the judgment date. It might be a judgment date, a hearing
date etc. but it is the "correct" date for the work.

Co-authored-by: jurisdatum <jim@jurisdatum.com>

https://trello.com/c/naUx8R1j/846-date-public-ui

### After

### DIFF
--- a/judgments/models.py
+++ b/judgments/models.py
@@ -20,7 +20,9 @@ class Judgment(xmlmodels.XmlModel):
         "//akn:proprietary/uk:cite", ignore_extra_nodes=True
     )
     date = xmlmodels.XPathTextField(
-        "//akn:FRBRdate[@name='judgment']/@date", ignore_extra_nodes=True
+        "/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/"
+        + "akn:FRBRdate/@date",
+        ignore_extra_nodes=True,
     )
     court = xmlmodels.XPathTextField("//akn:proprietary/uk:court")
     content_hash = xmlmodels.XPathTextField("//akn:proprietary/uk:hash")
--- a/judgments/models.py
+++ b/judgments/models.py
@@ -14,7 +14,9 @@ class Judgment(xmlmodels.XmlModel):
         }
 
     metadata_name = xmlmodels.XPathTextField(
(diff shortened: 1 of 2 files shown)


judgments/tests.py
```diff
@@ -117,11 +120,41 @@ def test_can_parse_judgment(self):
         model = Judgment.create_from_string(xml)
         self.assertEqual("My Judgment Name", model.metadata_name)
         self.assertEqual("[2017] EWHC 3289 (QB)", model.neutral_citation)
-        self.assertEqual("2004-06-10", model.date)
+        self.assertEqual("2004-06-10T10:30:00", model.date)
```
erbridge, line 123: Why this date and not `2020-01-01`?

erbridge: Nevermind. I have now reread the PR description and understand.

erbridge (approved):

judgments/models.py
```diff
@@ -14,13 +14,17 @@ class Meta:
...
+        + "akn:FRBRname/@value",
+        ignore_extra_nodes=True,
     )
     neutral_citation = xmlmodels.XPathTextField(
         "//akn:proprietary/uk:cite", ignore_extra_nodes=True
     )
     date = xmlmodels.XPathTextField(
-        "//akn:FRBRdate[@name='judgment']/@date", ignore_extra_nodes=True
+        "/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/"
+        + "akn:FRBRdate/@date",
```
dragon-dxw, line 26: This will impact the published dates in the ATOM feeds too, I strongly suspect, which could do with documenting. Feel free to make a card for me to do that on Wednesday.

dragon-dxw (approved):

PR #337, opened 2022-08-31
Remove remnants of the bootstrapping script

lozette:
## Changes in this PR:

We originally bootstrapped the application locally by adding the XSLT and
XQueries from the application into Marklogic.

Now we add these assets via Gradle from another repository (`ds-caselaw-public-
access service`) so they're no longer needed here.

https://trello.com/c/dylwn8lt/830-rename-judgment2-and-judgment0-xslts-in-public-access-service

### After


dragon-dxw (approved):
Sounds sensible, although I feel out of my depth.

PR #339, opened 2022-09-01
fix/h1 headings line height was too small for headings of more than o…

Terry-Price:
## Changes in this PR:
Added more line-height to H1 so that it is more readable when these headings are on more than one line.
Compensated the new extra space above the H1s by reducing the top margin of the Standard text template by 1rem.
the starting position is now the same.

### Before

Screenshot 2022-09-01 at 13 52 57

### After
Screenshot 2022-09-01 at 13 52 31

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_standard_text_template.scss
+++ b/ds_judgements_public_ui/sass/includes/_standard_text_template.scss
@@ -8,7 +8,7 @@
     max-width: 30rem;
   }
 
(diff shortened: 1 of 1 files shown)


Terry-Price:
Fixed H1s so they are more spaced when they run beyond one line. Extra space has been compensated by reducing standard text template top margin by 1 rem.

lozette (approved):

PR #342, opened 2022-09-07
Allow search engines to index static pages

dragon-dxw:
I've taken the simple route of just adding the pages to robots.txt, rather than changing our URL scheme.


dragon-dxw:
[Index static pages (in addition to home page)](https://trello.com/c/ggWzD6hq/922-index-static-pages-in-addition-to-home-page)

lozette (approved):

PR #348, opened 2022-09-12
Remove robots directives from static pages

timcowlishaw:
This commit selectively removes the noindex,nofollow 'robots' meta tag from all the static pages, as requested in [Trello ticket #922](https://trello.com/c/ggWzD6hq/922-%F0%9F%91%A9%E2%9A%96%EF%B8%8Findex-static-pages-in-addition-to-home-page).

I've taken a hidden-by-default approach - the robots tag appears in the `base.html` template, and we use a jinja2 block to allow it to be overridden in the templates that extend it. In the static page templates we therefore override this with an empty block to allow them to be indexed.

I've included a test of the homepage (as an example of a page where we do not want the tag to be included) and the results page (as an example of a page where we want the tag to be included) in order to validate the effectiveness of this approach, however it seemed overkill to include a test for every single template. If you disagree however, i'd be happy to add them!

### DIFF
--- a/judgments/tests.py
+++ b/judgments/tests.py
@@ -252,3 +252,18 @@ class TestConverters(TestCase):
     def test_subdivision_converter_fails_to_parse(self):
         converter = converters.SubdivisionConverter()
         self.assertIsNone(re.match(converter.regex, "notasubdivision"))
+
+
+class TestRobotsDirectives(TestCase):
+    def test_homepage(self):
+        # The homepage should not have a robots meta tag with nofollow,noindex
+        response = self.client.get("/")
+        self.assertNotContains(
+            response, '<meta name="robots" content="noindex,nofollow">'
+        )
+
+    def test_judgment_results(self):
+        # The judgment search results page should not have a robots meta tag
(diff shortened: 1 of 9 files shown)


judgments/tests.py
```diff
@@ -252,3 +252,18 @@ def test_subdivision_converter_parses_court(self):
...
+class TestRobotsDirectives(TestCase):
+    def test_homepage(self):
+        # The homepage should not have a robots meta tag with nofollow,noindex
+        response = self.client.get("/")
+        self.assertNotContains(
+            response, '<meta name="robots" content="noindex,nofollow">'
+        )
+
+    def test_judgment_results(self):
+        # The judgment search results page should not have a robots meta tag
```
lozette, line 266: I think this should be `should have a robots meta tag` (no "not")

lozette (changes requested):
One tiny change please!

judgments/tests.py
timcowlishaw: hah, yes, of course!

lozette (approved):

PR #352, opened 2022-09-14
Fix django-admin errors

timcowlishaw:
When  DJANGO_SETTINGS_MODULE is set in the environment for `fab sh`, the `django-admin` command fails as the given settings module is not on the python import path. resetting this so that it is empty fixes the problem (as seen also in `compose/production/django/start`, line 11). This commit ensures that this env var is reset on starting `fab-sh` so that django-admin commands can be run from within it without error!

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -106,7 +106,9 @@ def sh(c):
     """
     Run bash in a local container (with access to dependencies)
     """
-    subprocess.run(["docker-compose", "exec", "django", "bash"])
+    subprocess.run(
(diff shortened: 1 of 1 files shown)


dragon-dxw:
Probably also needs fixing in the editor-ui (these are currently slowly-diverging codebases)

dragon-dxw (approved):

dragon-dxw:
We've realised this doesn't quite do what we want it to do -- `django-admin` gives a different error:

"Note that only Django core commands are listed as settings are not properly configured (error: Requested setting INSTALLED_APPS, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.)."

PR #356, opened 2022-09-15
add CSS for forcing paragraph alignments in judgments

timcowlishaw:
## Changes in this PR:

Three little CSS escape hatches that Jim can use in the judgment html to ensure we have the correct layout!

https://trello.com/c/MOeZYajS/954-%F0%9F%91%A9%E2%9A%96%EF%B8%8Fcentre-headings-parser-and-fe

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text.scss
@@ -171,6 +171,10 @@
     text-align: center;
   }
 
+  // Classes with `pr-` prefix are purely presentational
+  // escape-hatches for dealing with specific formatting
+  // details in judgments that must be preserved.
+
   &__pr-center {
     text-align: center !important;
   }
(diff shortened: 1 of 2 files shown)


ds_judgements_public_ui/sass/includes/_judgment_text.scss
```diff
@@ -170,6 +170,18 @@
   &__approved-judgment {
     text-align: center;
   }
+
+  &__pr-center {
```
dragon-dxw, line 174: What does pr stand for?

dragon-dxw (approved):

ds_judgements_public_ui/sass/includes/_judgment_text.scss
timcowlishaw: aha, my thinking here was that it'd be useful to somehow give all purely presentational classes a common prefix so that we can identify them later in the event that we find a way of getting rid of them, hence `pr-` for `presentation`. I've added a comment to explain this (as it's definitely not clear on its own), but to be honest, I'm not actually convinced it's necessary, thinking again, so could be convinced to drop them!

PR #361, opened 2022-09-16
UI text changes for `s/Queen/King/gi`

timcowlishaw:
This removes all appropriate mentions of "Queen" / "Her Majesty" and replaces with a sensible alternative in the user facing text.

No identifiers are changed for the moment.

In the case of ui elements for searches / filters for judgments at the king's bench, this provides the single option "King's / Queen's bench division of the high court"

should work as a starter :-)


dragon-dxw (approved):

timcowlishaw:
Leaving this open so it can be merged next Tuesday if necessary!

PR #363, opened 2022-09-21
Feature/958 fix back to search results

timcowlishaw:
## Changes in this PR:

This reinstates the 'back to search results' link on individual judgments, which was present in the prototype but broken in such a way that it never displayed in the production django app.

The back link should only show when the user navigates to a judgment from a search results page - I have added tests for this logic.

@matt-blair  I'm not 100% happy with the typography / layout currently at the small breakpoint - I feel that perhaps the back link should be less prominent, but it looks messy in smaller type. Any advice greatly appreciated!

https://trello.com/c/d2DvIFNO/958-%F0%9F%8F%86back-to-search-results-button-on-judgements-not-displaying

### Before
Screenshot 2022-09-21 at 12 07 29
Screenshot 2022-09-21 at 12 07 43

### After
Screenshot 2022-09-21 at 12 00 51
Screenshot 2022-09-21 at 12 00 35

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
@@ -5,35 +5,18 @@
   padding: $spacer__unit;
   border-bottom: 0.125rem solid $color__grey;
 
-  a {
-    background-color: $color__aqua-blue;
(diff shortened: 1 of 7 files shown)


dragon-dxw (approved):
Looks great! Thank you for the screenshots!

PR #364, opened 2022-09-21
Add the option to choose the number of search results

timcowlishaw:
## Changes in this PR:

Users can choose either 10, 25, or 50 results to be shown per page in the results page and in the advanced search.

This also fixes a bug whereby if the user refined their search filters, the search order selected would be forgotten.

https://trello.com/c/TyBKKRXo/986-%F0%9F%8F%86see-more-search-results-option-for-10-20-50-per-page

### Before

Screenshot 2022-09-21 at 16 59 15
Screenshot 2022-09-21 at 16 59 03

###
Screenshot 2022-09-21 at 16 59 19
Screenshot 2022-09-21 at 16 58 38
 After

### DIFF
--- a/judgments/views.py
+++ b/judgments/views.py
@@ -33,6 +33,8 @@ from .utils import perform_advanced_search
 
 env = environ.Env()
 
+MAX_RESULTS_PER_PAGE = 50
+
 
 def browse(request, court=None, subdivision=None, year=None):
     court_query = "/".join(filter(lambda x: x is not None, [court, subdivision]))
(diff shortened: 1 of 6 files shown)


judgments/views.py
```diff
@@ -100,9 +100,13 @@ def advanced_search(request):
         "order": params.get("order", ""),
         "from": params.get("from"),
         "to": params.get("to"),
+        "per_page": params.get("per_page"),
     }
     page = params.get("page", 1)
     page = page if page else 1
+    per_page = (
+        params.get("per_page") if params.get("per_page") else str(RESULTS_PER_PAGE)
```
dragon-dxw, line 108: Might be worth putting a sensible maximum on the page size -- it feels like a good way of asking for literally everything, which might make denial of service a possibility. (I'm not convinced this is actually a problem, but best to be proactive.)

dragon-dxw (approved):

judgments/views.py
timcowlishaw: aha yep good shout - i will take a look!

PR #367, opened 2022-09-22
Fix/button visited colour

Terry-Price:
Fixed  visited colour change on the text for search tag buttons and also 'Show more filters text so that they remain black. 
### Before
Screenshot 2022-09-22 at 10 09 16
Screenshot 2022-09-22 at 10 55 14

### After
Screenshot 2022-09-22 at 10 09 51
Screenshot 2022-09-22 at 10 01 35


timcowlishaw (approved):
Looks good to me!

PR #371, opened 2022-09-23
Rethink judgment paragraph numbering layout on small breakpoint

timcowlishaw:
## Changes in this PR:
A slightly 'creative' solution in code terms, but I think design-wise it makes sense (credit for which goes to Jim).

At the smallest breakpoint, hanging section numbers take up too much screen real estate making the judgement hard to read. Instead, we indent the first line of the para after the number, making the para numbers easily recognisable while not losing  whole length of the line for text.

https://trello.com/c/PtAFMfAb/965-mobile-view-paragraph-numbering-of-judgments-is-appearing-above-the-text

### Before
Screenshot 2022-09-23 at 13 08 05

### After
Screenshot 2022-09-23 at 16 36 28

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text.scss
@@ -201,6 +201,10 @@
   }
 
   &__number {
+    @media (max-width: $grid__breakpoint-medium) {
+      padding-right: $spacer__unit;
+    }
+
     &:target {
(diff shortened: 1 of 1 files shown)


timcowlishaw:
(Ignore that shadow on the left margin in the  'after' screenshot, that's not actually part of the design)

dragon-dxw (approved):
👍

PR #372, opened 2022-09-26
Content Security Policy headers

dragon-dxw:
We use the `django-csp` library to create CSP headers for us.

We permit a couple of Google URLs to provide script, style and fonts
to us: these can be found on the front page (and probably every page)

The inscrutible sha-256 hash `IWjjek`... refers to the inline script
in `templates/base.py` and appears to be the only inline script in
the public-ui.

There's two bigger problems:

* HTML versions of the judgments have a great many inline styles. These
are taken directly from the word document. We don't have a good way
of not using inline styles here right now.

* `data:` uris are used in the CSS, but isn't secure. It's fine, because the
CSS isn't dynamically generated, but I can't see an obvious way to
allow the CSS to use `data:`

We've set it to report only for now; it should report CSP issues
to rollbar as warnings. I recommend we run this in production for
a while, and if there's no errors, turn on the actual protection a
little later.

Or possibly this is only useful as an understanding of the challenges we'd face if we wanted to tighten up Content Security.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -123,6 +123,7 @@ MIDDLEWARE = [
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
+    "csp.middleware.CSPMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.common.BrokenLinkEmailsMiddleware",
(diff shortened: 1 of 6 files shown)


dragon-dxw:
[Add Content-Security-Policy headers](https://trello.com/c/dMmdisu3/463-add-content-security-policy-headers)

dragon-dxw:
Isn't quite right - there are issues with inline styles.

anthonyhashemi:
@dragon-dxw  I see in the ticket for this you said that it might not be worth doing this and just throw it away. Should we close this PR?

PR #376, opened 2022-09-26
Skip to end of judgement

timcowlishaw:
## Changes in this PR:

This adds a 'skip to bottom' link on individual judgments, and makes a few stylistic changes to the 'back to top' / 'skip to bottom' to make them a little more obviously 'separate' to the judgment text itself.

https://trello.com/c/x6r4sen4/994-move-search-specific-fields-to-the-end-of-the-filter-options

### After
Screenshot 2022-09-26 at 18 12 44
Screenshot 2022-09-26 at 18 12 35

Screenshot 2022-09-26 at 18 13 07
Screenshot 2022-09-26 at 18 13 01
Screenshot 2022-09-26 at 18 14 20
Screenshot 2022-09-26 at 18 14 14


dragon-dxw (approved):
It'd be really nice if we could rebase these down to one commit: the way I'd do that is:

* make sure on the right branch
* `git fetch`
* `git reset --hard` (will destroy any uncommitted changes)
* `git rebase -i origin/main`

That'll bring up an editor window -- if you change all but the first `pick` to `fixup` they'll be absorbed into the first commit. Or rearrange them until they make sense -- happy to talk this though!

(If you want to change the first commit's name, you can do that with `reword` at the same time).

* `git push --force`

Yell if this is scary/unclear/I've lead you astray, and I can talk you through it (I'm a recent convert)

-------------------------

Bikeshedding: I don't like the floating scroll up/down pane partially overlapping the text at medium screen resolutions. It isn't my decision, though, and I have no useful suggestions.

timcowlishaw:
Hey, @dragon-dxw  i've rebased off master but just seen your instructions about doing an interactive rebase - i didn't do that this time so you'll get my entire history, but it sounds like a plan for the next big-ish PR. When i get to that point would you be up for a quick screenshare to talk me through it?

PR #377, opened 2022-09-28
Court URIs and per-page bug

timcowlishaw:
## Changes in this PR:

This was meant to be a bunch of very small uncontroversial tweaks (hence rolling them into one PR): Making sure we use the `kc` and `scco` courts identifiers everywhere instead of their `qc` and `costs` aliases, and ensuring that the "number or search results" option never clears the advanced filters.

However i think there's one aspect that might deserve a little more attention on review as it was a little more complex than I anticipated: In particular the changes to the logic in results_controls.html. Essentially, we want to ensure that the hidden form fields are always rendered when there are *any* query params (so that filters are preserved), but that the "order results" dropdown is only displayed when there is a freetext query string (as 'order by' only makes sense when there's a relevance score). I think this now works as expected, and have tested every permutation I can think of, but another pair of eyes would be great!

Thanks!

Tim

1015/1016/1017

### DIFF
--- a/ds_judgements_public_ui/static/js/src/modules/document_navigation_links.js
+++ b/ds_judgements_public_ui/static/js/src/modules/document_navigation_links.js
@@ -50,7 +50,12 @@ $(() => {
                 });
         }
     }
-
-    createObserver(document.querySelector(".judgments-footer"));
-    createObserver(document.querySelector(".judgment-toolbar__container"));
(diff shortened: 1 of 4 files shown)


dragon-dxw (commented):
The first and third commits don't seem directly relevant to this PR -- unless I'm missing something? (They look good though)

dragon-dxw (approved):

PR #378, opened 2022-09-29
Use fields from search results in SearchResult model

lozette:
## Changes in this PR:

~This should not be pushed to production before nationalarchives/ds-caselaw-public-access-service#88 is deployed to Marklogic production.~

Previously, we were getting a judgement's neutral citation, name, court, date and hash by making an API call to Marklogic, getting the judgment XML, and parsing it. We can actually get these elements as part of the search results, and it means we lose an extra call to the API. As we show at least 10 results per page, this means the app will load faster, as we lose at least 10 calls per page.

https://trello.com/c/wBZUsMuz

### After

### DIFF
--- a/judgments/models.py
+++ b/judgments/models.py
@@ -65,9 +65,53 @@ class SearchResult:
         matches = SearchMatch.create_from_string(
             etree.tostring(node, encoding="UTF-8").decode("UTF-8")
         )
-        judgment_xml = api_client.get_judgment_xml(uri)
-        judgment = Judgment.create_from_string(judgment_xml)
-
+        namespaces = {
+            "search": "http://marklogic.com/appservices/search",
+            "uk": "https://caselaw.nationalarchives.gov.uk/akn",
+            "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
+        }
+        neutral_citation = (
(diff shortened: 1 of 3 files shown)


judgments/models.py
```diff
@@ -65,9 +37,53 @@ def create_from_node(node):
...
+            else ""
+        )
+        metadata_name = (
+            node.xpath("search:extracted/akn:FRBRname/@value", namespaces=namespaces)[0]
+            if node.xpath("search:extracted/akn:FRBRname/@value", namespaces=namespaces)
+            else ""
+        )
+        date = (
+            node.xpath(
+                "search:extracted/akn:FRBRdate[not(@name='transform')]/@date",
```
dragon-dxw, line 62: Feels odd to use "not transform date"; there feels like there's potentially an issue if more dates come along (e.g. enrichment). Is this "judgment" vs "decision", and would `(@name='judgment' or @name='decision')` be better?

lozette: Yes this felt weird to me too. Unfortunately the dates both come out as `search:extracted/akn:FRBRdate`. Adding (@name='judgment' or @name='decision') would work though!

lozette: Amended

dragon-dxw (approved):

PR #383, opened 2022-10-05
Add `per_page` to query string passed with pagination links

timcowlishaw:
This was an oversight when adding the page size selector (my fault) - the parameter needs to be passed with the pagination links in order to preserve the number of results per page as you pass through pages!


dragon-dxw (approved):
I'm looking at this and I'm concerned about the possibility of injection attacks on the other contexts -- will take a deeper look but is a bigger question than this PR.

PR #386, opened 2022-10-06
Boost Strict Transport Security (force HTTPS) to six days

dragon-dxw:
WARNING: There is, as far as I can tell, no sensible way of testing this.
Please check my maths.

We currently force HTTPS if the user has used HTTPS in the last sixty seconds.
This gives us confidence that we're not trying to access some resource that isn't, for some reason, available
via HTTPS.

We should improve the protection this gives us, but we also don't want to accidentally lock ourselves out forever should there be an undiscovered problem. Moving to six days (so we reduce the policy to 0, then work on it the week after) gives an intermediary step between "we can't use caselaw.nationalarchives.gov.uk for a year" and "users aren't protected from http downgrades".


dragon-dxw:
[STS header sets a short time](https://trello.com/c/jjaX7XrN/371-sts-header-sets-a-short-time)

Floppy (approved):
I'm happy with this approach, yes.

PR #391, opened 2022-10-10
Redirect uppercase URLs to lowercase

dragon-dxw:
This is a sledgehammer approach -- ALL urls with capital ASCII letters [A-Z] will be redirected to urls without capital letters.

### DIFF
--- a/config/tests.py
+++ b/config/tests.py
@@ -2,8 +2,20 @@ from django.test import TestCase
 
 
 class TestRedirect(TestCase):
-    def test_slash_redirection_works(self):
+    def test_slash_redirection(self):
         url = "/arbitrary/url/with/slash/at/end/"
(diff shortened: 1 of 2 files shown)


dragon-dxw:
[judgment url is case sensitive - this is correct but unexpected for users](https://trello.com/c/m8QvzhwP/375-judgment-url-is-case-sensitive-this-is-correct-but-unexpected-for-users)

dragon-dxw:
Other approaches tried:
Converters should not have `(?i)` added to the start of them; they get inserted mid-regex, which is deprecated at best and broken at worst.

Alternative:
Consider a custom 404 handler so we know we're not rewriting anything that's got another match. This might be a good place for the `.../` redirection too.

dragon-dxw:
Generally seen as too broad

PR #397, opened 2022-10-12
Don't highlight stop words in search results

lozette:
## Changes in this PR:

https://trello.com/c/Ry47hPz5

This PR does two things:

1. Does not highlight any individual stop words (`the, of, and`) in search result snippets. If the stop words are in a quoted string, they are highlighted (e.g. "weight of evidence"). But if they are individual stop words in an unquoted string, they are not highlighted.

2. If a search query is unquoted and contains stop words (the, and, of), do not pass
those stop words to the search query. Unless the stop word is the only word, in which case pass it through. If the query is quoted (e.g. "weight of evidence") then do not remove the stop
words.

## Screenshots of UI changes:

Both searches are for `evidence of wrongdoing` (without quotes)

### Before

Screenshot 2022-10-17 at 11 19 15

### After

Screenshot 2022-10-17 at 11 18 48

### DIFF
--- a/judgments/tests.py
+++ b/judgments/tests.py
@@ -7,7 +7,7 @@ from lxml import etree
 
 import judgments.models
 import judgments.utils  # noqa: F401 -- used to mock
-from judgments import converters, views
+from judgments import converters, utils, views
 from judgments.models import SearchResult, SearchResults
 from judgments.views import display_back_link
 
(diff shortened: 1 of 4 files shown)


dragon-dxw:
me: "Is there a better way of writing that"
stack overflow: "not in XSLT 1"
https://stackoverflow.com/questions/1007018/xslt-expression-to-check-if-variable-belongs-to-set-of-elements

Do you have any idea what the sticking points on XSLT 2 are (which would also let us use a list of strings for comparison)

judgments/tests.py
```diff
@@ -318,3 +318,15 @@ def test_min_max():
...
+
+def test_prep_query():
+    assert utils.remove_unquoted_stop_words("weight of evidence") == "weight evidence"
+    assert (
+        utils.remove_unquoted_stop_words("'weight of evidence'")
+        == "'weight of evidence'"
+    )
+    assert utils.remove_unquoted_stop_words("the") == ""
+    assert utils.remove_unquoted_stop_words("'the'") == "'the'"
+    assert utils.remove_unquoted_stop_words("judge and jury") == "judge jury"
```
dragon-dxw, line 331: Is `cats and dogs "kittens and puppies"` a meaningful search term we should check?

lozette: Good question!!

dragon-dxw (approved):

PR #409, opened 2022-10-28
Fix empty results bug when browsing by court

timcowlishaw:
## Changes in this PR:
When browsing by court from the homepage, and changing the number of results per page, no courts were returned. This was because the form to change the number of results per page was including Null parameters as hidden fields in error - have added a check to make sure these are not added to the query.

https://trello.com/c/Pe6F0tx5/44-browsing-into-a-court-and-changing-search-results-adds-all-advanced-filters-to-search-with-none-value-leading-to-0-results


dragon-dxw (approved):
Thanks!

PR #410, opened 2022-10-28
Bugfixes arising from first tier tribunals

timcowlishaw:
## Changes in this PR:

Introducing the first tier tribunals revealed a couple of bugs in theUI that needed addressing:

1) A slightly weird styling wrinkle in the search filters that improperly offset the court checkbox when the previous court's name ranged onto two lines

2) The court name list on the homepage was still hardcoded in the app, rather than using the canonical list imported from ds-caselaw-utils

This commit fixes both these things!


ds_judgements_public_ui/sass/includes/_structured_search.scss
```diff
@@ -221,6 +221,10 @@
     width: 50%;
     float: left;
 
+    &:nth-of-type(odd) {
```
dragon-dxw, line 224: I'm not quite getting what this does -- happy to assume it's sensible.

dragon-dxw (approved):

PR #430, opened 2022-11-16
changed text on homepage banner to give year start (2003) and simplif…

Terry-Price:
…ied the link to what to expect page so that it more CTA and taken out New

- Added the start year (from 2003) to help users understand that this is not an archive of judgments going way back.
- Also altered the link to the 'What to expect' page so that it was simpler and more direct. Also dropped the 'new' word.

This is part of a 'let's do this for now' approach to show the start of the judgment year range.
https://trello.com/c/5FQqzmGM/58-pui-holistic-approach-search-results
### Before
Screenshot 2022-11-16 at 16 24 49

### After
Screenshot 2022-11-16 at 16 24 28

### DIFF
--- a/locale/en_GB/LC_MESSAGES/django.po
+++ b/locale/en_GB/LC_MESSAGES/django.po
@@ -197,7 +197,7 @@ msgstr "Search results"
 
 #: ds_judgements_public_ui/templates/pages/home.html:23
 msgid "home.useservice"
-msgstr "Use this service to find, view and download judgments and tribunal decisions from 2003 gigga"
+msgstr "Use this service to find, view and download judgments and tribunal decisions from 2003"
 
 #: ds_judgements_public_ui/templates/pages/how_to_use_this_service.html:8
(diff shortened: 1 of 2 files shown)


locale/en_GB/LC_MESSAGES/django.po
```diff
@@ -197,8 +197,7 @@ msgstr "Search results"
 
 #: ds_judgements_public_ui/templates/pages/home.html:23
 msgid "home.useservice"
-msgstr ""
-"Use this service to find, view and download judgments and tribunal decisions"
+msgstr "Use this service to find, view and download judgments and tribunal decisions from 2003 gigga"
```
dragon-dxw, line 200: sorry, the random junk I asked you to add is still here.

dragon-dxw (commented):
Sorry, the random junk I asked you to put in to see if we could get recompiling the translations to work is still here.

Terry-Price:
I will resend a new pull request with updated amends.

PR #433, opened 2022-11-17
fix to neutral citation number to keep it on one

Terry-Price:
## Changes in this PR:
Fix to the Neutral citation where the number can break in to two lines on mobile viewports.
I have added a `white-space: no-wrap` to make sure the whole number stays on one line and if it does break, it will break after the label only.

### Before
BEFORE

### After
AFTER

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text.scss
@@ -114,7 +114,7 @@
     text-align: right;
 
     span {
-      font-weight: bolder;
+      font-weight: normal;
       display: inline-block;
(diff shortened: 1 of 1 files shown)


ds_judgements_public_ui/sass/includes/_judgment_text.scss
```diff
@@ -114,7 +114,7 @@
     text-align: right;
 
     span {
-      font-weight: bolder;
+      font-weight: normal;
```
dragon-dxw, line 117: I don't see a `white-space: nowrap`, have you commited and pushed those changes?

dragon-dxw (changes requested):

PR #434, opened 2022-11-17
updated fix changes to NCN so that it no longer wraps on to two lines…

Terry-Price:
… on mobile

## Changes in this PR:
Fixed an issue where the NCN was wrapping on two lines on smaller mobile devices.
It now uses a span around the actual NCN number of `white-space:nowrap` to stop it breaking.

### Before
BEFORE

### After
AFTER

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text.scss
@@ -114,9 +114,10 @@
     text-align: right;
 
     span {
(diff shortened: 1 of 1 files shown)


dragon-dxw:
(Assuming you're happy with the changes to the font-weight; I don't see it reflected in the image)

Terry-Price:
yes, I think that will be okay. Not sure why it was set to `bolder` anyway.

PR #435, opened 2022-11-18
fixed ncn from wrapping with a new span class

Terry-Price:
## Changes in this PR:

- Apologies Dragon, I did have to target the NCN no-wrap issue with a new extra span class.
- I have also taken off the `font-weight-bolder` on the generic `span` as it was overwriting the NCN label style which takes president.

### Before
BEFORE

### After
AFTER


Terry-Price:
need to fix lint check

jacksonj04 (approved):
These changes look good to me; if we could squash these commits together in to one to keep it cleaner that would be great. Hit me up if you're not sure how to go about that.

PR #437, opened 2022-11-22
Add a check for secrets

jacksonj04:
## Changes in this PR:

This action will check the repository for committed secrets, and fail (allowing the user to take action) if they're present.

https://trello.com/c/9VhTJmtg/138-add-check-for-secrets-to-ds-caselaw-public-ui

### DIFF
--- a/.github/workflows/secrets.yml
+++ b/.github/workflows/secrets.yml
@@ -1,0 +1,11 @@
+name: Source safety
+on:
+  pull_request:
+  push:
+jobs:
+  secrets:
+    name: Check for secrets
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


dragon-dxw:
Hmmm, it runs twice:
image

(Perhaps it doesn't need to run on PR? Seems fine as is though)
@jacksonj04

dragon-dxw (approved):

dragon-dxw (approved):

dragon-dxw:
> They overlap when you create PRs from the same repo, but you need pull_request if you want to run an action when you receive a PR from a fork for example. 

https://stackoverflow.com/questions/70767090/github-actions-on-push-and-on-pull-request-difference

This implies this is absolutely correct. Shame it can't just run once...

PR #473, opened 2022-12-14
Add note on translations and django-admin

timcowlishaw:
Add a note to the readme elaborating on how to change existing translations, and how to get around the module not defined error that arises when using the django-admin command


dragon-dxw (approved):
A couple of typos I'd like fixed

PR #481, opened 2022-12-20
fixed mobile view of judgment party table where the last column words…

Terry-Price:
Fixed mobile view of judgment party table where breaking into two lines, also moved the centered text to the left a little as a layout compromise to enable the fix

https://trello.com/c/qq5DbMaK/116-reduce-party-wrapping-on-mobile
### Before
Screenshot 2022-12-20 at 12 48 51

### After
Screenshot 2022-12-20 at 12 47 23


jacksonj04 (approved):
This is definitely better! I suspect we'll run into the same problem again if people view it with larger than usual default text size, but ultimately there's only so much space on screen.

🚀

timcowlishaw (approved):
Looks great to me, nicely done!

PR #502, opened 2023-01-06
SPIKE: Add ability to cache court dates in the django postgres database

timcowlishaw:
## Changes in this PR:

Nick propose a cron job that populates the database with the start and end years for each court's current ingested range  on a daily or weekly basis - this PR is intended as a conversation starter on how that might work on the python side.

We add some thin decorators around the existing Court and CourtGroup classes from the utils library which check for the existence of the dates in the db, falling back to those defined in the yaml file if they don't exist.

One nice thing is that this will transparently fallback to the existing data in utils if the database records don't exist, therefore we could deploy something-like-this invisibly even if the cron job / marklogic query isn't yet ready.

This is a spike, so will need (a) some test coverage, and (b) some thinking about optimisation before it's ready for release. In particular, it currently does a single database query for each court in the metadata file, whenever the court metadata is needed, which is probably not ideal.

Comments, feedback, criticisms, counterproposals all welcome!


judgments/models.py
```diff
@@ -119,3 +124,87 @@ class Meta:
...
+        return self.__decorate(self.repo.get_listable_groups())
+
+    def get_listable_courts(self):
+        return self.__decorate(self.repo.get_listable_courts())
+
+    def get_listable_tribunals(self):
+        return self.__decorate(self.repo.get_listable_tribunals())
+
+
+courts = CourtsRepositoryDecoratedWithDates(utils_courts)
```
timcowlishaw, line 210: (I'm aware that these classnames are all getting a bit enterprisey and `AbstractSingletonProxyFactoryBean`-like. Not sure if this is because the patterns themselves are un-pythonic?)

judgments/models.py
```diff
@@ -119,3 +124,87 @@ class Meta:
...
     transform_to_html = xmlmodels.XsltField(join(dirname(__file__), "search_match.xsl"))
+
+
+class CourtDates(models.Model):
+    param = models.CharField(max_length=64, primary_key=True)
+    start_year = models.IntegerField(blank=False)
+    end_year = models.IntegerField(blank=False)
+
+
+class CourtDecoratedWithDates:
```
dragon-dxw, line 135: Are there good reasons for this not to be a subclass of utils.Court?

dragon-dxw (approved):
Radical rethink: how would this look if we only had the Django model and registered a template function that returned the best dates given a court param? We keep the existing functionality in one place (rather than duplicating it with additional complexity) at the cost of slightly more mysterious templates.

judgments/models.py
timcowlishaw: none at all - and that will simplify things nicely. Will do that now! 🙌

PR #513, opened 2023-01-17
Feature/override court dates in db

timcowlishaw:
## Changes in this PR:

This is a 'actually ready to be merged if we want to' version of [my previous spike](https://github.com/nationalarchives/ds-caselaw-public-ui/pull/502) incorporating @dragon-dxw and @jacksonj04's feedback.

Nick propose a cron job that populates the database with the start and end years for each court's current ingested range on a daily or weekly basis - this adds logic on the python side in the form of a django model to store that data, and a template helper which looks up the date range for a given court and transparently falls back to the details in the utils yaml if none exists. Therefore, it could be deployed even in the absence of the cronjob for now and nothing would change in user-facing terms.

I've also added some test coverage!

## Trello card:

https://trello.com/c/NElb5lZV/324-updating-judgment-dates-from-marklogic


dragon-dxw (approved):

timcowlishaw:
Thanks @dragon-dxw - I'll wait to discuss in standup before merging just to check everyone's happy with the general approach!

Donna-H (approved):

PR #516, opened 2023-01-19
Add ability to cache court dates in the django postgres database (again)

timcowlishaw:
## Changes in this PR:

Re-applying this commit after reverting it -before merging we will need to make sure that migrations are run on deploy.

Nick propose a cron job that populates the database with the start and end years for each court's current ingested range  on a daily or weekly basis - this handles the persistence layer in the app and the logic to choose the correct date range when displaying a court.

https://trello.com/c/NElb5lZV/324-bau-updating-court-dates-from-marklogic


judgments/templatetags/court_utils.py
```diff
@@ -10,3 +13,18 @@ def get_court_name(court):
     if court_object is None:
         return ""
     return court_object.name
+
+
+@register.filter
+def get_court_date_range(court):
+    try:
+        db_dates = CourtDates.objects.get(pk=court.canonical_param)
+        start_year = db_dates.start_year
```
anthonyhashemi, line 22:
```suggestion
        start_year = court_dates.start_year
```

judgments/templatetags/court_utils.py
```diff
@@ -10,3 +13,18 @@ def get_court_name(court):
...
+        end_year = db_dates.end_year
```
anthonyhashemi, line 23:
```suggestion
        end_year = court_dates.end_year
```

judgments/models.py
```diff
@@ -119,3 +119,9 @@ class Meta:
         namespaces = {"search": "http://marklogic.com/appservices/search"}
 
     transform_to_html = xmlmodels.XsltField(join(dirname(__file__), "search_match.xsl"))
+
+
+class CourtDates(models.Model):
```
anthonyhashemi, line 124: not a concern for now - but do we envisage adding other court info to the db?

if so, we may consider generalising this class to something like `Court`.

timcowlishaw: Aha, so, i started a spike for moving all the metadata into the DB and it complicated things considerably, so decided to keep this PR more specific. Agree that it's probably a good medium-term aim though!

timcowlishaw:
Folks, I'm gonna mark this ready for review before leaving but please *DO NOT MERGE* until we're happy the migrations will run on deploy!

dragon-dxw (approved):

jacksonj04 (approved):

PR #518, opened 2023-01-23
Simplify readme to reference manage.py not django-admin

anthonyhashemi:
## Changes in this PR:

- Replaced all references to `django-admin` commands with the equivalent `python manage.py` command.
- Removed the notes on how we need to set the DJANGO_SETTINGS env var when running `django-admin` commands as this is done automatically with `manage.py`

The reason I think this is fine is because manage.py is what django suggests to use as opposed to django-admin once the project has been created, 

Excerpt from

https://docs.djangoproject.com/en/4.1/ref/django-admin/#:~:text=Generally%2C%20when%20working%20on%20a,%2D%2Dsettings%20command%20line%20option.

```
Generally, when working on a single Django project, it’s easier to use manage.py than django-admin. If you need to switch between multiple Django settings files, use django-admin with [DJANGO_SETTINGS_MODULE](https://docs.djangoproject.com/en/4.1/topics/settings/#envvar-DJANGO_SETTINGS_MODULE) or the [--settings](https://docs.djangoproject.com/en/4.1/ref/django-admin/#cmdoption-settings) command line option.

The command-line examples throughout this document use django-admin to be consistent, but any example can use manage.py or python -m django just as well.
```

## Trello card

https://trello.com/c/TmK81uEu/330-bau-update-public-ui-readmemd-so-dev-startup-is-simpler

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,13 +179,13 @@ $ fab sh
 Run the following inside the `django` container
 
 ```console
-# python manage.py migrate
+python manage.py migrate
 ```
 
 #### Run a 'development' web server
 
 ```console
-$ python manage.py runserver_plus 0.0.0.0:3000
+python manage.py runserver_plus 0.0.0.0:3000
 ```
 
 ### Running the test suite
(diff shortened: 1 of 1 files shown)


dragon-dxw (approved):

dragon-dxw:
I hadn't quite understood the difference between the two until now. Thanks!

anthonyhashemi:
> 

Yeah I wasn't 100% sure either but since I always use manage.py and I saw the django-admin references here I went to look at the docs

PR #535, opened 2023-01-30
Tweak mobile layout to ensure placeholder text is visible

timcowlishaw:
## Changes in this PR:

Thanks to Terry for the inspiration that led to this solution, and help with some annoying typography tweaks - at the smallest breakpoint we swap the label and placeholder text, to ensure the full instructions are visible!

https://trello.com/c/nLaavzHS/346-da-%E2%9C%8F%EF%B8%8F-pui-placeholder-text-in-search-bar-gets-cut-off-in-mobile-view

### Before
Screenshot 2023-01-30 at 13 47 37

### After

Screenshot 2023-01-30 at 13 31 18

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_search.scss
+++ b/ds_judgements_public_ui/sass/includes/_search.scss
@@ -54,8 +54,18 @@
     padding-right: calc($spacer__unit * 3);
   }
 
-  &__search-term-label {
+  &__search-term-label.desktop {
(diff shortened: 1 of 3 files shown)


dragon-dxw (approved):
TIL that `display: none` also successfully hides from screenreaders, so this should be fine.

jacksonj04 (approved):

PR #549, opened 2023-02-01
Remove unnecessary autogenerating docs

jacksonj04:
We don't use Sphinx to autogenerate any documentation, nor do we use readthedocs. Remove unnecessary files and dependencies.


anthonyhashemi (approved):
Approved.

So I imagine there was an initial desire to use readthedocs but we don't want to anymore / it just never ended up happening?

dragon-dxw (approved):

PR #554, opened 2023-02-02
Set views and static content to have 1hr cache headers

dragon-dxw:
We configure Whitenoise for the assets and set up a middleware for the cache headers.

1 hour value based on legislation.gov.uk -- do we want s-maxage and no-transform?

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -123,6 +123,7 @@ MIDDLEWARE = [
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
+    "config.middleware.CacheHeaderMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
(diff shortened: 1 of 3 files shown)


config/settings/base.py
```diff
@@ -247,3 +248,4 @@
 # Your stuff...
 # ------------------------------------------------------------------------------
 APPEND_SLASH = False
+WHITENOISE_MAX_AGE = 3600  # set Cache-Control header on static-served images
```
jacksonj04, line 251: Is 1hr all we want here, or can we turn it up?

dragon-dxw: Default is 60 seconds; this might even be *too* long as it is.

The default is chosen to be short enough not to cause problems with stale versions but long enough that, if you’re running WhiteNoise behind a CDN, the CDN will still take the majority of the strain during times of heavy load.

http://whitenoise.evans.io/en/latest/django.html#WHITENOISE_MAX_AGE

jacksonj04: 👍🏻

jacksonj04 (approved):

dragon-dxw:
[Cache headers in PUI](https://trello.com/c/NcdoC4p0/414-cache-headers-in-pui)

PR #571, opened 2023-02-21
adding modified package json files to avoid stashing them in future

Terry-Price:
## Changes in this PR:

Added the two automatically modified Package.json files when I install Sass via NPM.
This happens every time I install sass, so would like to avoid stashing or restoring these files each time.

[](https://trello.com/c/mILukxmQ/473-make-editor-public-ui-automigrate-in-development)

### Before
Screenshot 2023-02-21 at 11 25 13

### After
Screenshot 2023-02-21 at 11 26 02

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,8 +5,12 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "ds-judgments-public-ui",
       "version": "1.0.0",
       "license": "ISC",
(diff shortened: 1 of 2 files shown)


dragon-dxw:
Since we use sass it's not a terrible thing if it's in the node package.json

dragon-dxw (approved):

PR #583, opened 2023-03-01
Mobile view buttons panels fix

Donna-H:
## Changes in this PR:
Fix mobile view, The search button was no longer as wide as the text box, fix styling issues on the ‘Update filters’ button, and on the label text above the search buttons

https://trello.com/c/AQU5XFb7/536-mobile-view-buttons-label-text
### Before

before-mobile-update-button

### After

[update-mob-view-fix.webm](https://user-images.githubusercontent.com/75584408/222178674-f1cee1a5-1565-4a15-a7e8-dde4a2c9af9c.webm)

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_structured_search.scss
+++ b/ds_judgements_public_ui/sass/includes/_structured_search.scss
@@ -16,7 +16,7 @@
     color: $color__almost-black;
     margin: calc($spacer__unit * 3) 0 0 0;
     @media (max-width: $grid__breakpoint-medium) {
-      margin: 1rem 0 1rem 1rem;
+      margin: 1px 0 1px 1px;
       line-height: 2.6rem;
     }
   }
@@ -78,7 +78,7 @@
 
   &__limit-to-label {
     display: block;
-    margin-bottom: 0;
+    margin-bottom: 1rem;
     font-size: 1rem;
     font-weight: bold;
   }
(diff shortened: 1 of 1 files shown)


Terry-Price (changes requested):
Hi Donna,
Thanks for this.

I did notice that the form labels have differing spaces between them and the text fields. Can those be fixed to be consistent.

Also, Looking at the new styles, was unsure about why the margin is 1px. Seems a bit small, should it be 0?

The blue button at the bottom is still full width on mobile, could you look to match the Search button at the top?

I can help maybe if you want to pair up later?

Donna-H:
> Hi Donna, Thanks for this.
> 
> I did notice that the form labels have differing spaces between them and the text fields. Can those be fixed to be consistent.
> 
> Also, Looking at the new styles, was unsure about why the margin is 1px. Seems a bit small, should it be 0?
> 
> The blue button at the bottom is still full width on mobile, could you look to match the Search button at the top?
> 
> I can help maybe if you want to pair up later?

@Terry-Price the trello card stated " search button was no longer as wide as the text box" so submit button at bottom is  as top search button now changed to as wide as text box

jacksonj04 (changes requested):
I think this could be rebased and all these commits squashed into one.

Donna-H:
@jacksonj04 rebased please check all is okay

jacksonj04:
@Donna-H if we can squash them into one commit I think this is good to go!

jacksonj04 (approved):

Terry-Price (approved):
we'll pick up any issues with this when Tim is around.

PR #590, opened 2023-03-09
Use assets domain link if S3 PDF exists

dragon-dxw:
Use [staging.]assets.caselaw.nationalarchives.gov.uk urls in preference to data.pdf -- added note that data.pdf is still relevant!

I'm not 100% sure this is the right answer -- it means we have two different PDF links.


dragon-dxw:
[PUI: Change asset/PDF behaviour](https://trello.com/c/QFdtPscQ/567-pui-change-asset-pdf-behaviour)

jacksonj04 (approved):
:shipit:

PR #591, opened 2023-03-09
Change view cache to 15 minutes

dragon-dxw:
Doesn't affect static assets.

### DIFF
--- a/config/middleware.py
+++ b/config/middleware.py
@@ -13,7 +13,7 @@ class CacheHeaderMiddleware:
         # the view (and later middleware) are called.
 
         response = self.get_response(request)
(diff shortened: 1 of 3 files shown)


dragon-dxw:
[PUI: Change caching time](https://trello.com/c/0W9iXCyb/568-pui-change-caching-time)

jacksonj04 (approved):
These changes look sound, but there's an unchanged test which is failing (possibly a duplicate).

dragon-dxw:
Hmmm. Tests pass locally. I think this might be a "local vs live" problem. Going to set the other caches to 15 minutes too.

PR #593, opened 2023-03-14
Fix advanced search bug

timcowlishaw:
## Changes in this PR:

This fixes the bug identified by Nicki where the advanced search page failed to filter by query term. The issue was an oversight when making earlier changes to the query form apperance on mobile - in order to show the placeholder text only on desktop, we need to provide two different inputs (one for mobile, one for desktop), and selectively show each one with CSS. This in turn means they need different IDs, or the one that is second in the markup will 'overwrite' the one that is first at submit time. This is what happened here - both fields were named 'query' and therefore the (blank) mobile field was used in preference to the (filled) desktop one.

Have fixed by renaming the mobile field, and added a test to ensure the correct behaviour.

It's been mentioned that perhaps there's no need to use placeholder text at any breakpoint, which would remove a lot of incidental technical complexity here, and stop these kinds of things from happening again. Will leave that for another day, but it'd be a very worthwhile simplification i think!

https://trello.com/c/moe7NdzX/594-pui-odd-search-behaviour


dragon-dxw (approved):
:shipit:

jacksonj04 (approved):
Hooray for improved tests!

PR #596, opened 2023-03-15
Fix format of date ranges supplied to date inputs

timcowlishaw:
## Changes in this PR:
This fixes [#630 - Calendar year missing on firefox browers](https://trello.com/c/hG5D5ivY/630-calendar-year-missing-on-firefox-browsers) - the issue was that the `min` and `max` dates specified used the wrong format ("DD-MM-YYYY" instead of "YYYY-MM-DD"). It appears that most browsers were clever enough to parse the other format, but FF had troubles. Have amended to the YYYY-MM-DD format which is specified in the standard.

https://trello.com/c/hG5D5ivY/630-calendar-year-missing-on-firefox-browsers

## Screenshots of UI changes:
image

### DIFF
--- a/ds_judgements_public_ui/templates/includes/results_filters_inputs.html
+++ b/ds_judgements_public_ui/templates/includes/results_filters_inputs.html
@@ -13,12 +13,12 @@
   <div class="structured-search__multi-fields-panel">
     <div class="structured-search__limit-to-container">
       <label for="from_date" class="structured-search__limit-to-label">From date</label>
-      <input class="structured-search__date-input" id="from_date" max="3-3-2022" min="02-01-2003" name="from"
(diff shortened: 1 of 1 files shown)


timcowlishaw:
Oh, i've also removed the `max` constraint from both fields as it was already out of date by a year (it only allowed dates up to march 2022 to be specified, and we've got plenty of judgments since then). Unless there's value that i've missed in adding this back and having it updated dynamically from the judgment collection, i suggest it's unnecessary?

anthonyhashemi (approved):

anthonyhashemi:
> 

yeah I was going to ask about this. Surprised that nobody complained about the max day being in the past for so long?

dragon-dxw (approved):

PR #612, opened 2023-03-28
Add more tracking parameters to feedback survey link

jacksonj04:
These extra parameters give us more information on what the user was actually doing at the time they hit the feedback link, rather than relying on them telling us.

### DIFF
--- a/config/middleware.py
+++ b/config/middleware.py
@@ -52,6 +52,9 @@ class FeedbackLinkMiddleware:
                 "feedback_survey_judgment_uri"
             ]
 
+        if "feedback_survey_court" in response.context_data:
+            params["court"] = response.context_data["feedback_survey_court"]
+
         response.context_data["feedback_survey_link"] = (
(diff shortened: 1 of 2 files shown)


config/middleware.py
```diff
@@ -19,3 +21,41 @@ def __call__(self, request):
         # the view is called.
 
         return response
+
+
+class FeedbackLinkMiddleware:
+
+    BASE_FEEDBACK_URL: str = (
+        "https://corexmsnp4n42lf2kht3.qualtrics.com/jfe/form/SV_0lyyYAzfv9bGcyW"
```
dragon-dxw, line 29: A part of me is saying "this should be an environment variable", another says it's totally not worth it.

dragon-dxw (approved):

config/middleware.py
jacksonj04: I had the exact same thought, and concluded "eh". If we change it in future we can revisit it, at least it's still in exactly one place.

PR #615, opened 2023-04-03
Remove no-results

dragon-dxw:
The no-results view and template were redundant.

## Screenshot
[not that this is different from before]

image


dragon-dxw:
[PUI: Remove 'no results' view](https://trello.com/c/ijoJ84TY/705-pui-remove-no-results-view)

PR #622, opened 2023-04-06
Add a 'back to top' link to static pages

timcowlishaw:
## Changes in this PR:

For ease of navigation.

This is added to  the accessibility statement, how to use this service, open justice licence, terms of use, transactional licence, and what to expect pages.

## Screenshots of UI changes:
Screenshot 2023-04-06 at 12 08 04

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_help_end_document_marker.scss
+++ b/ds_judgements_public_ui/sass/includes/_help_end_document_marker.scss
@@ -1,11 +1,10 @@
-.judgment-end-document-marker {
+.help-end-document-marker {
     width: 100%;
-    border: 1px dashed $color__black;
-    border-top: none;
+    border-top: 4px solid $color__yellow;
     height: $spacer__unit;
     max-width: 90%;
     margin: auto;
-    margin-bottom: calc($spacer__unit * 7);
+    margin-bottom: calc($spacer__unit * 8);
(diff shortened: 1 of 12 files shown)


ds_judgements_public_ui/sass/includes/_help_end_document_marker.scss
```diff
@@ -0,0 +1,55 @@
+.help-end-document-marker {
```
anthonyhashemi, line 1: This class only slightly differs from `.judgment-end-document-marker ` with the differences shown in the screenshot:

Screenshot 2023-04-06 at 11 31 30

Are these differences intentional?

If so can we refactor so that we reduce duplication, otherwise can we drop one of them?

jacksonj04 (commented):
Would it be easier to have a new `static-document` template which we pass a header into, and that does all the lifting for making sure the anchor and link are in place?

ds_judgements_public_ui/sass/includes/_help_end_document_marker.scss
timcowlishaw: aha yep, the stylistic difference is on purpose, but yep, agreed on cleaning up the definitions. I'll take a look!

Terry-Price (approved):
looks great Tim

PR #629, opened 2023-04-06
Provoke AWS into sending Content-Length header

dragon-dxw:
head used to get content lengths, but does not any more, unless Accept-Encoding is unset, presumably because compressed files might have different lengths (unsure if a requests or AWS change)


dragon-dxw:
LGTM from Nick

PR #635, opened 2023-04-11
took out search label to let users find thier own way

Terry-Price:
## Changes in this PR:

- Having reviewed the original requirements, we decided to try and take the label and the help text away from the main search box for both the home screen and the structured search screen.
- I have also reduced some excess margin to help reduce scrolling and display more page content at the top.
- **Please note, to keep this PR small, I will be revisiting the "_Filter by court, date or person_" text design on a new card after this one.**

https://trello.com/c/65C8ohkh/761-%F0%9F%94%8D-pui-browse-make-search-instructions-label-more-prominent-add-instructions-for-phrasal-search
### Before
Screenshot 2023-04-11 at 16 17 50
Screenshot 2023-04-11 at 16 17 36

### After
Screenshot 2023-04-11 at 16 10 34
Screenshot 2023-04-11 at 16 10 07

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_buttons.scss
+++ b/ds_judgements_public_ui/sass/includes/_buttons.scss
@@ -1,0 +1,43 @@
+.button-cta {
+    @include call-to-action-button;
+    border: 2px solid $color__cta-background;
+
+    &:disabled,  &[aria-disabled="true"] {
+      background-color: $color__dark-grey;
+      &:focus, &:hover {
+        outline: none;
+        background-color: $color__dark-grey;
+        color: $color__white;
+        text-decoration: none;
+        cursor: not-allowed;
+      }
+    }
+  }
+
+  .button-secondary {
+    @include call-to-action-button;
+    background-color: $color__white;
+    color: $color__cta-background;
+    border: 2px solid $color__cta-background;
+
+    &:visited {
+      color: $color__cta-background;
+    }
+
+    &:focus, &:hover {
+      color: white;
+    }
+
+    &:disabled,  &[aria-disabled="true"] {
+      color: $color__dark-grey;
+      background-color: transparent;
+      border-color: $color__grey;
+      &:focus, &:hover {
+        outline: none;
+        background-color: transparent;
+        color: $color__dark-grey;
+        text-decoration: none;
+        cursor: not-allowed;
+      }
+    }
+  }
--- a/locale/en_GB/LC_MESSAGES/django.po
+++ b/locale/en_GB/LC_MESSAGES/django.po
@@ -197,9 +197,11 @@ msgstr "Search"
 msgid "basicsearchform.placeholder"
 msgstr "Search..."
 
+
+
 #: ds_judgements_public_ui/templates/includes/search_term_component.html
-msgid "basicsearchform.helptext"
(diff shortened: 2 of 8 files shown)


locale/en_GB/LC_MESSAGES/django.po
```diff
@@ -197,9 +197,11 @@ msgstr "Search"
 msgid "basicsearchform.placeholder"
 msgstr "Search..."
 
+
+
 #: ds_judgements_public_ui/templates/includes/search_term_component.html
 msgid "basicsearchform.helptext"
-msgstr "Enter keyword, party name or neutral&nbsp;citation"
+msgstr "Enter keyword&#44; party name neutral&nbsp;citation or phrases surrounded by quote marks e.g. &#34;blade crime&#34;"
```
jacksonj04, line 204: I think this has lost some punctuation in transit.
```suggestion
msgstr "Enter keyword, party name, neutral&nbsp;citation or phrases surrounded by quote marks e.g. &#34;blade crime&#34;"
```

jacksonj04 (changes requested):

locale/en_GB/LC_MESSAGES/django.po
dragon-dxw, line 204: the comma should be absolutely fine as-is.
I'd prefer `&quot;` to `&#34;` if `'..."blade crime"'` doesn't work.
Blade crime is a bit of a meme amongst ourselves, but maybe something more generic would be useful: `"statutory instrument"`, perhaps?

ds_judgements_public_ui/sass/includes/_buttons.scss
jacksonj04, line 1: We really need to have a think about unifying styles between the two services!

locale/en_GB/LC_MESSAGES/django.po
```diff
@@ -197,9 +197,11 @@ msgstr "Search"
 msgid "basicsearchform.placeholder"
 msgstr "Search..."
 
+
+
 #: ds_judgements_public_ui/templates/includes/search_term_component.html
-msgid "basicsearchform.helptext"
-msgstr "Enter keyword, party name or neutral&nbsp;citation"
+msgid "basic.search.helptext"
```
jacksonj04, line 203:
```suggestion
msgid "search.basic.helptext"
```

This should probably go from least-specific to most-specific.

jacksonj04 (approved):
One tiny ordering fix in a key, otherwise 🚢

ds_judgements_public_ui/sass/includes/_buttons.scss
Terry-Price: Yes @jacksonj04 I only brought those styles in because at one point I thought I would need them, but the card direction changed.

We only have one common scss file and that is for judgments content. it would be much better if there was more common single sources of truth which we could change in one place . I don't know how to do that though.

locale/en_GB/LC_MESSAGES/django.po
timcowlishaw: aha, agree, but maybe this is a thing we should look at across the whole site? our translation keys are somewhat improvised everywhere at the moment, and it'd be good to give them a more consistent heirarchy! I'll add a card to look at this as a refactoring if that's ok!

PR #637, opened 2023-04-12
add time class to html

Donna-H:
## Changes in this PR:
Add time class to html
https://trello.com/c/NQ8zVIAi/765-%F0%9F%94%8D-pui-add-time-tags-where-we-display-datetimes
### Before
before-time-add-pui
### After
after-time-added


dragon-dxw (approved):

timcowlishaw (changes requested):
Hey Donna, this is great, but those `<time>` tags also need their `datetime` attribute set - it should be the date of the judgment in YYYY-MM-DD format!

timcowlishaw (approved):

PR #645, opened 2023-04-17
Install django-flags

timcowlishaw:
## Changes in this PR:
Following instructions from https://cfpb.github.io/django-flags/

This requires access to the admin UI, enabling which exposed a couple of issues:

1) The admin urls in urls.py need to come *before* the redirect to remove the trailing slash, as the trailing slash is required in the admin urls (otherwise it tries to redirect to  eg /adminlogin instead of /admin/login for some reason)

2) The ADMIN_URL needs to *not* have a preceding slash

3) The django default templates need to be loaded for the admin UI to display

This PR fixes all the above, new flags can be added to FLAGS in `setttings/base.py` as required, then enabled in the admin.

https://trello.com/b/OYgnaYZc/1-caselaw-beta

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -3,6 +3,7 @@ Base settings to build other settings files upon.
 """
 from pathlib import Path
 
+import django
 import environ
 
 ROOT_DIR = Path(__file__).resolve(strict=True).parent.parent.parent
@@ -48,6 +49,8 @@ DATABASES["default"]["ATOMIC_REQUESTS"] = True
 # https://docs.djangoproject.com/en/stable/ref/settings/#std:setting-DEFAULT_AUTO_FIELD
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
+FLAGS = {}
+
 # URLS
 # ------------------------------------------------------------------------------
 # https://docs.djangoproject.com/en/dev/ref/settings/#root-urlconf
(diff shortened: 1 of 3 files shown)


jacksonj04:
Do we want django-flags or django-waffle? Flags doesn't support progressive roll-out, _but_ it does have a cleaner implementation for templating and forces us to keep flags defined in one place.

timcowlishaw:
Hmm, i'm pro django-flags for the reasons you give - is progressive
roll-out something we're likely to want?

jacksonj04:
I think it's something we're likely to want _eventually_, but it's pretty low-effort to migrate between flag libraries so I'm happy with this.

config/settings/base.py
```diff
@@ -161,7 +165,7 @@
         # https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-TEMPLATES-BACKEND
         "BACKEND": "django.template.backends.django.DjangoTemplates",
         # https://docs.djangoproject.com/en/dev/ref/settings/#dirs
-        "DIRS": [str(APPS_DIR / "templates")],
```
jacksonj04, line 164: I'm not sure I follow the reason for this change.

timcowlishaw: Aha, this is to include the templates for django-admin, as we need them to set flags!

jacksonj04 (approved):
Ship it!

PR #649, opened 2023-04-20
Remove privacy link from footer

dragon-dxw:
Whilst we need one, it is better to not have one than have one with wrong information.

Before:
image

After:
image


timcowlishaw (approved):

jacksonj04:
Do we have a followup card for actually solving this?

PR #650, opened 2023-04-20
Simple command for translation file editing

dragon-dxw:
`fab translate` will update the translation file.
Sadly, the date change precludes putting this in precommit at this time.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -110,6 +110,24 @@ def sh(c):
 
 
 @task
+def translate(c):
+    """
+    Make messages for translations
+    """
+    subprocess.run(
+        [
+            "docker-compose",
+            "exec",
+            "django",
+            "bash",
+            "-c",
+            "python manage.py makemessages --no-obsolete --add-location file -l en_GB "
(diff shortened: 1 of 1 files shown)


anthonyhashemi (approved):

jacksonj04:
I did look at solving the date change problem by having a script which would do a diff on contents and only update the file if there was an actual change, but it involves a whole horrible dance of stashing stuff and making copies to compare if there's actually been a state change, so I decided against it.

It's probably only an hour's work or so if we can combine brains and talk the problem through with a whiteboard though.

fabfile.py
```diff
@@ -109,6 +109,23 @@ def sh(c):
...
+    Make messages for translations
+    """
+    subprocess.run(
+        [
+            "docker-compose",
+            "exec",
+            "django",
+            "bash",
+            "-c",
+            "python manage.py makemessages --no-obsolete --add-location file -l en_GB",
```
jacksonj04, line 124: Should this also `compilemessages`?

PR #651, opened 2023-04-20
Bugfix/796 exact phrase search bug

timcowlishaw:
## Changes in this PR:
Normalise quotes in search strings

This fixes two related bugs that were raised by Trello card 796.

     - Firstly, when a search query is pasted in from somewhere (eg word, google docs) that applies 'smart quotes' formatting - the curly quote characters aren't picked up as 'quotes' by marklogic, and it appears that the phrasal search fails. This normalises them to regular ascii quotes beforehand, to ensure that they're respected by marklogic as phrases
     - Secondly, we've been removing stopwords (outside of quoted phrases) when searching from the basic search form, but not from the advanced form. This leads to an incinsistency in the results returned between the two.
     - Finally, our tests for removing stopwords assumed single quotes which *aren't* treated as quotes by marklogic - if we wanted them to be, then this would be a more difficult problem - possibly intractible given we can't identify single quotes meant as quotes from single quotes meant as apostrophes (eg in the phrase ('that's the problem'). This PR at least makes the behaviour consistent for both forms of english double quotes (I've assumed eg spanish-style angled quotes are out of scope, as they're not typically automatically changed in any software).

The result of this is that the search results returned will always be consistent across advanced and basic search, and double quotes will always be respected as phrases.

https://trello.com/c/64eURjn6/796-pui-exact-phrase-search-results-arent-ordered-by-judgments-containing-the-exact-phrases

### Before

image
image

### After
Screenshot 2023-04-20 at 19 09 37
Screenshot 2023-04-20 at 19 07 06

### DIFF
--- a/judgments/tests.py
+++ b/judgments/tests.py
@@ -37,12 +37,37 @@ class TestSearchResults(TestCase):
     @patch("judgments.views.results.perform_advanced_search")
     @patch("judgments.models.SearchResult.create_from_node")
     @patch("judgments.utils.perform_advanced_search")
-    def test_judgment_results_desktop(self, f3, fake_result, fake_advanced_search):
+    def test_judgment_results(self, f3, fake_result, fake_advanced_search):
         fake_advanced_search.return_value = fake_search_results()
         f3.r = fake_search_results()
         fake_result.return_value = fake_search_result()
-        # The judgment search results view takes the query term from the (desktop) query input
         response = self.client.get("/judgments/results?query=waltham+forest")
+        self.assertContains(
+            response,
+            '<span class="results-search-component__removable-options-value-text">waltham forest</span>',
+        )
+
+    @patch("judgments.views.results.perform_advanced_search")
+    @patch("judgments.models.SearchResult.create_from_node")
+    @patch("judgments.utils.perform_advanced_search")
+    @patch("judgments.views.results.preprocess_query")
+    def test_jugdment_results_query_preproccesed(
+        self, fake_preprocess_query, f3, fake_result, fake_advanced_search
+    ):
+        fake_advanced_search.return_value = fake_search_results()
+        f3.r = fake_search_results()
+        fake_result.return_value = fake_search_result()
+        fake_preprocess_query.return_value = "normalised query"
+        self.client.get("/judgments/results?query=waltham+forest")
+
+        fake_preprocess_query.assert_called()
+
+    @patch("judgments.views.advanced_search.perform_advanced_search")
+    @patch("judgments.models.SearchResult.create_from_node")
+    def test_judgment_advanced_search(self, fake_result, fake_advanced_search):
+        fake_advanced_search.return_value = fake_search_results()
+        fake_result.return_value = fake_search_result()
+        response = self.client.get("/judgments/advanced_search?query=waltham+forest")
         self.assertContains(
             response,
             '<span class="results-search-component__removable-options-value-text">waltham forest</span>',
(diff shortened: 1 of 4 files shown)


judgments/tests.py
```diff
@@ -396,13 +396,14 @@ def test_min_max():
 
 
 def test_prep_query():
-    assert utils.remove_unquoted_stop_words("weight of evidence") == "weight evidence"
-    assert (
-        utils.remove_unquoted_stop_words("'weight of evidence'")
-        == "'weight of evidence'"
-    )
+    # Stopwords outside quoted strings are removed.
+    assert utils.preprocess_query("weight of evidence") == "weight evidence"
```
anthonyhashemi, line 400: Could be worth splitting each of these asserts into their own test functions since they are testing different aspects of the functions and if they are so.

e.g. 
 # Stopwords outside quoted strings are removed.
    assert utils.preprocess_query("weight of evidence") == "weight evidence"
    
    can become `test_preprocess_query_removes_stopwords_quoted_strings`


At minimum I think we should definitely split out the pre_process_query asserts from the remove_unquoted_stop_words asserts.

judgments/views/results.py
```diff
@@ -14,6 +14,7 @@
     has_filters,
     paginator,
     perform_advanced_search,
+    preprocess_query,
```
anthonyhashemi, line 17: can we add a test for the results view that captures the fact that the query now passes through `preprocess_query`?

judgments/tests.py
jacksonj04: Agree that `preprocess_query` and `remove_unquoted_stop_words` are two different things. I'm not sure each assertion needs an individual test at this point unless we see regressions in future though, but I have no strong opinions either way so if we want the robustness then go for it!

judgments/tests.py
```diff
@@ -396,13 +396,14 @@ def test_min_max():
...
+    # Stopwords inside quoted strings are removed
+    assert utils.preprocess_query('"weight of evidence"') == '"weight of evidence"'
+    # Curly quotes are normalised to straight ones and treated as such:
+    assert utils.preprocess_query("“weight of evidence”") == '"weight of evidence"'
     assert utils.remove_unquoted_stop_words("the") == "the"
-    assert utils.remove_unquoted_stop_words("'the'") == "'the'"
+    assert utils.remove_unquoted_stop_words('"the"') == '"the"'
```
jacksonj04, line 406: This line looks like Black has chewed it up a bit, we might need an ignore statement.

judgments/utils.py
```diff
@@ -51,6 +51,14 @@ def perform_advanced_search(
     return SearchResults.create_from_string(multipart_data.parts[0].text)
 
 
+def preprocess_query(query):
+    return remove_unquoted_stop_words(normalise_quotes(query))
```
jacksonj04, line 55: I'd be tempted to break this down so it's more obvious what it's doing (and the order), even though it's more lines:

```suggestion
   query = normalise_quotes(query)
   query = remove_unquoted_stop_words(query)
   return query
```

judgments/utils.py
```diff
@@ -51,6 +51,14 @@ def perform_advanced_search(
...
+
+
+def normalise_quotes(query):
+    return re.sub(r"[“”]", '"', query)
```
jacksonj04, line 59: Is it also worth normalising single quotes?

jacksonj04 (commented):
Couple of smol bits, and the WIP commit needs squashing. Nice bit of work though, hopefully will deal with a whole bunch of related search unhappiness at once.

judgments/utils.py
timcowlishaw: aha, been pondering  that - marklogic doesn't currently treat them as phrase delimiters, so it's probably less important. That is a further, thornier question - if we were to treat them as phrase delimiters, we need to be very careful about single quotes that are apostrophes (which i think means we're out of regex land and into chomsky grammars and such as we need to find matched pairs of apostrophes, not just blindly replace them), so i'd be inclined to leave it for now, maybe add a ticket for further investigation?

judgments/tests.py
timcowlishaw: aaaah yep well spotted - i'll add.

judgments/utils.py
timcowlishaw: hah lol, I nearly went the other way and started using functools to compose functions together, (my Haskell past coming back to bite me) but yes, agreed :-)

anthonyhashemi (approved):
Looks good!

PR #652, opened 2023-04-21
If a date from the database is unparseable, use the raw string

jacksonj04:
Some judgments don't have a valid date attribute as understood by the `SearchResult` class. This means their date is set to `""`, which is unparseable. Since we try parse this string, it throws an error.

This PR catches the error and falls back to the raw date string. This will usually be an empty string, but could (in theory) be all kinds of unparseable stuff which nonetheless can be understood by a human.

### DIFF
--- a/judgments/models.py
+++ b/judgments/models.py
@@ -2,6 +2,7 @@ from os.path import dirname, join
 
 from caselawclient.Client import api_client
 from dateutil import parser as dateparser
+from dateutil.parser import ParserError
 from django.db import models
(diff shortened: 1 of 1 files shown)


anthonyhashemi (approved):
can we have a followup pr adding a test?

jacksonj04:
I'm in the process of refactoring how we handle the entire `SearchResults` class so we can unify it in the API library, which will include writing tests for the whole thing.

PR #653, opened 2023-04-24
Fix server error in RSS feeds with unexpected judgment dates

jacksonj04:
New behaviour in `SearchResult` introduced in https://github.com/nationalarchives/ds-caselaw-public-ui/pull/652 would result in cases where `date` was a string, rather than a `datetime`. This behaviour is changed so that it returns `None`.

A string would cause the RSS library to fail; a `None` is correctly handled.


judgments/models.py
```diff
@@ -30,7 +30,7 @@ def __init__(
         try:
             self.date = dateparser.parse(date)
         except ParserError:
-            self.date = date
+            self.date = None
```
dragon-dxw, line 33: It might be worth flagging a rollbar error if the date is not empty and fails to parse; but it shouldn't raise.

dragon-dxw (approved):
Whilst there's improvements that can be made, this is a production problem, and this feels like a good fix.

judgments/tests/test_atom_feed.py
```diff
@@ -0,0 +1,43 @@
...
+    @patch("judgments.utils.perform_advanced_search")
+    def test_bad_page_404(self, fake_advanced_search):
+        # "?page=" 404s, not 500
+        fake_advanced_search.return_value = fake_search_results()
+        response = self.client.get("/atom.xml?page=")
+        self.assertEqual(response.status_code, 404)
+
+    @patch("judgments.feeds.perform_advanced_search")
+    @patch("judgments.models.SearchResult.create_from_node")
+    def test_feed_with_empty_date(self, fake_result, fake_advanced_search):
```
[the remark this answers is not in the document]
jacksonj04: Backported some testing for this from the refactor work, just for belt and braces.

judgments/models.py
jacksonj04: Added with tests

PR #655, opened 2023-04-24
Change date selectors to year/month/date text inputs

Donna-H:
## Changes in this PR:
With js and server side validation. Currently validation errors are not displayed, the filter is just ignored - this will be addressed in future work.

This lives behind the TEXT_DATE_INPUT feature flag, which will have to be enabled to see it :-)

https://trello.com/c/72rPrOKi/768-%F0%9F%94%8D-pui-try-out-using-seperate-month-year-inputs-instead-of-a-calendar-picker-on-advanced-search
### Before
before-date

### After
after-date

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -50,7 +50,8 @@ DATABASES["default"]["ATOMIC_REQUESTS"] = True
 # https://docs.djangoproject.com/en/stable/ref/settings/#std:setting-DEFAULT_AUTO_FIELD
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
-FLAGS: Dict[str, Any] = {}
+# Feature flags for django-flags
+FLAGS: Dict[str, Any] = {"TEXT_DATE_INPUT": []}
 
 # URLS
 # ------------------------------------------------------------------------------
--- a/ds_judgements_public_ui/static/js/src/modules/date_input.js
+++ b/ds_judgements_public_ui/static/js/src/modules/date_input.js
@@ -1,0 +1,26 @@
+import $ from "jquery";
+
+function constrainInputToMinMax() {
+    const input = $(this);
+    var lastValidValue = input.value;
+    input.on("input", (event) => {
+        const validity = input[0].validity;
+        if (validity.badInput) {
+            this.value = lastValidValue;
+        } else if (validity.rangeOverflow) {
+            this.value = this.max;
+        } else {
+            lastValidValue = this.value;
+        }
+    });
+}
+
+function setupDateInput() {
+    const wrapper = $(this);
+    const inputs = wrapper.find("input");
+    inputs.each(constrainInputToMinMax);
+}
+
+$(() => {
+    $(".js-date-input").each(setupDateInput);
+});
--- a/judgments/models.py
+++ b/judgments/models.py
@@ -7,6 +7,7 @@ from caselawclient.Client import api_client
 from dateutil import parser as dateparser
 from dateutil.parser import ParserError
 from django.db import models
+from django.db.models import Max, Min
 from djxml import xmlmodels
 from ds_caselaw_utils.courts import CourtNotFoundException, courts
 from lxml import etree
@@ -144,3 +145,13 @@ class CourtDates(models.Model):
     param = models.CharField(max_length=64, primary_key=True)
     start_year = models.IntegerField(blank=False)
     end_year = models.IntegerField(blank=False)
+
+    @staticmethod
+    def min_year():
+        result = CourtDates.objects.aggregate(Min("start_year"))
+        return result["start_year__min"]
+
+    @staticmethod
+    def max_year():
+        result = CourtDates.objects.aggregate(Max("end_year"))
+        return result["end_year__max"]
--- a/judgments/templatetags/court_utils.py
+++ b/judgments/templatetags/court_utils.py
@@ -1,3 +1,6 @@
+import logging
+from datetime import date
+
 from django import template
 from django.utils.safestring import mark_safe
 from ds_caselaw_utils.courts import CourtNotFoundException
@@ -15,6 +18,26 @@ def get_court_name(court):
     except CourtNotFoundException:
         return ""
     return court_object.name
+
+
+@register.simple_tag
+def get_first_judgment_year():
+    if min_year := CourtDates.min_year():
+        return min_year
+    else:
+        logging.warning("CourtDates table is empty! using fallback min_year.")
+        return min(court.start_year for court in all_courts.get_selectable())
+
+
+@register.simple_tag
+def get_last_judgment_year():
+    if max_year := CourtDates.max_year():
+        return max_year
+    else:
+        logging.warning("CourtDates table is empty! using fallback max_year.")
+        # The dates in all_courts don't work as a fallback, as they can't
+        # be relied on to be up to date.
+        return date.today().year
 
 
 @register.filter
--- a/judgments/test_date_parsing.py
+++ b/judgments/test_date_parsing.py
@@ -1,0 +1,130 @@
+from typing import Dict
+
+from django.test import TestCase
+
+from judgments.utils import parse_date_parameter
+
+
+class TestDateParsing(TestCase):
+    def test_when_a_date_is_provided(self):
+        """
+        When a parameter is provided with the given name, return that value itself.
+        """
+        params = {"date": "2019-12-02"}
+        parsed = parse_date_parameter(params, "date")
+        self.assertEqual(parsed, "2019-12-02")
+
+    def test_provided_date_takes_precedence(self):
+        """
+        When a parameter is provided with the given name, return that value, even if day month and year are provided.
+        """
+        params = {
+            "date": "2019-12-02",
+            "date_day": "3",
+            "date_month": "1",
+            "date_year": "2020",
+        }
+        parsed = parse_date_parameter(params, "date")
+        self.assertEqual(parsed, "2019-12-02")
+
+    def test_a_blank_date_does_not_count_as_provided(self):
+        """
+        When a blank parameter is provided with the given name, use the day month year parameters for preference.
+        """
+        params = {"date": "", "date_day": "3", "date_month": "1", "date_year": "2020"}
+        parsed = parse_date_parameter(params, "date")
+        self.assertEqual(parsed, "2020-01-03")
+
+    def test_constructs_a_date_from_date_parts(self):
+        """
+        When no date is provided directly, it constructs one from the given day month and year.
+        """
+        params = {"date_day": "3", "date_month": "1", "date_year": "2020"}
+        parsed = parse_date_parameter(params, "date")
+        self.assertEqual(parsed, "2020-01-03")
+
+    def test_returns_none_if_no_date_provided(self):
+        """
+        When no date or date parts are provided, it returns None.
+        """
+        params: Dict[str, str] = {}
+        parsed = parse_date_parameter(params, "date")
+        self.assertIsNone(parsed)
+
+    def test_returns_beginning_of_month_when_no_day(self):
+        """
+        When a year and month are provided, but no day, it defaults to the first day of the month.
+        """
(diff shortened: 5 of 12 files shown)


ds_judgements_public_ui/static/js/src/modules/date_input.js
```diff
@@ -0,0 +1,26 @@
...
+function constrainInputToMinMax() {
+    const input = $(this);
+    var lastValidValue = input.value;
+    input.on("input", (event) => {
+        const validity = input[0].validity;
+        if (validity.badInput) {
+            this.value = lastValidValue;
+        } else if (validity.rangeOverflow) {
+            this.value = this.max;
+        } else if (!validity.badInput) {
```
anthonyhashemi, line 12: Would an `else` be sufficient here since by default `!validity.badInput` is true at this point in the conditional block?

ds_judgements_public_ui/static/js/src/modules/date_input.js
```diff
@@ -0,0 +1,26 @@
+import $ from "jquery";
+
+function constrainInputToMinMax() {
```
anthonyhashemi, line 3: It would be worth testing the event listener. Do we have any tests for other event listeners in the repo?

judgments/views/advanced_search.py
```diff
@@ -12,13 +12,26 @@
...
+    try:
+        from_date = parse_date_parameter(params, "from")
+    except ValueError:
+        from_date = None  # TODO - we need to add a mechanism
+        # for informing user of input validation errors
+
+    try:
+        to_date = parse_date_parameter(params, "to", default_to_last=True)
+    except ValueError:
+        to_date = None  # TODO - as above :-)
```
anthonyhashemi, line 33: same here about linking to trello card if not being addressed in this PR

judgments/test_date_parsing.py
```diff
@@ -0,0 +1,135 @@
+from django.test import TestCase
+
+from judgments.utils import parse_date_parameter
+
+
+class TestDateParsing(TestCase):
+    def test_when_a_date_is_provided(self):
+        """
+        When a parameter is provided with teh given name, return that value itself.
```
anthonyhashemi, line 9:
```suggestion
        When a parameter is provided with the given name, return that value itself.
```

judgments/utils.py
```diff
@@ -167,3 +168,33 @@ def has_filters(query_params, exclude=["order", "per_page"]):
...
+    month_param_name = f"{param_name}_month"
+    day_param_name = f"{param_name}_day"
+    if parameter_provided(params, param_name):
+        return params[param_name]
+    elif parameter_provided(params, year_param_name):
+        year = parse_parameter_as_int(params, year_param_name, default=1)
+        default_month = 12 if default_to_last else 1
+        month = parse_parameter_as_int(params, month_param_name, default=default_month)
+        default_day = monthrange(year, month)[1] if default_to_last else 1
+
```
anthonyhashemi, line 197: for consistency with the year and month fields being next to each other, can we remove this newline between these 2 lines?

judgments/utils.py
```diff
@@ -167,3 +168,33 @@ def has_filters(query_params, exclude=["order", "per_page"]):
...
+        return default
+
+
+def parse_date_parameter(params, param_name, default_to_last=False):
+    year_param_name = f"{param_name}_year"
+    month_param_name = f"{param_name}_month"
+    day_param_name = f"{param_name}_day"
+    if parameter_provided(params, param_name):
+        return params[param_name]
+    elif parameter_provided(params, year_param_name):
```
anthonyhashemi, line 192: Why is `elif parameter_provided(params, year_param_name):` required?

Could we not just have this whole elif block as the main function block, since we return early if `param_name` is provided? 

i.e

```
def parse_date_parameter(params, param_name, default_to_last=False):
    if parameter_provided(params, param_name):
        return params[param_name]

    default_year = 1
    year = parse_parameter_as_int(params, f"{param_name}_year", default=default_year)

    default_month = 12 if default_to_last else 1
    month = parse_parameter_as_int(params, f"{param_name}_month", default=default_month)

    default_day = monthrange(year, month)[1] if default_to_last else 1
    day = parse_parameter_as_int(params, f"{param_name}_day", default=default_day)
    
    dt = datetime(year, month, day)
    return dt.strftime("%Y-%m-%d")
```

judgments/models.py
```diff
@@ -135,3 +136,13 @@ class CourtDates(models.Model):
     param = models.CharField(max_length=64, primary_key=True)
     start_year = models.IntegerField(blank=False)
     end_year = models.IntegerField(blank=False)
+
+    @staticmethod
+    def min_year():
```
anthonyhashemi, line 141: can we add tests for min_year and max_year

anthonyhashemi (commented):
Generally looks good! I havent reviewed the css in detail, but the screenshots looks good to me. I have left quite a few comments, so please feel free to reach out. More than happy to jump on a huddle to go through them all :D

config/settings/base.py
```diff
@@ -49,7 +49,7 @@
 # https://docs.djangoproject.com/en/stable/ref/settings/#std:setting-DEFAULT_AUTO_FIELD
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
-FLAGS = {}
+FLAGS = {"TEXT_DATE_INPUT": []}
```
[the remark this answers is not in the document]
timcowlishaw: aha, this is what django-flags expects, annoyingly - I don't *think* it's changable. I wil check though! Agree something more explicit would be better

judgments/views/advanced_search.py
```diff
@@ -12,13 +12,26 @@
...
 )
 
 
 def advanced_search(request):
     params = request.GET
+
+    try:
+        from_date = parse_date_parameter(params, "from")
+    except ValueError:
+        from_date = None  # TODO - we need to add a mechanism
```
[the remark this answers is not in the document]
timcowlishaw: Let's make a trello card - i think it's an 'important enhancement' rather than an essential - since this sits behind  a feature flag it's in a state that's user-test-able (the validation never outright fails with eg a 500, even if it doesn't surface errors), even though if we wanted to *release* it publicly we'd need to think harder about error recovery. Will add a card!

judgments/test_date_parsing.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        When a year is provided but no month or day, and the default_to_last
+        option is passed, it defaults to the last day of december of that year.
+        """
+        params = {"date_year": "2020"}
+        parsed = parse_date_parameter(params, "date", default_to_last=True)
+        self.assertEqual(parsed, "2020-12-31")
+
+    # TODO Donna! Copy the test above twice and add two new tests for when you
+    # pass a date with a year and a month, and have default_to_last=True. One
```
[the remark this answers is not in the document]
timcowlishaw: hah, yes, good point!

judgments/templatetags/court_utils.py
```diff
@@ -17,6 +19,24 @@ def get_court_name(court):
...
+    else:
+        return min(court.start_year for court in all_courts.get_selectable())
+
+
+@register.simple_tag
+def get_last_judgment_year():
+    if max_year := CourtDates.max_year():
+        return max_year
+    else:
+        # The dates in all_courts don't work as a fallback, as they can't
```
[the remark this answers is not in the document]
timcowlishaw: aha, i *thought* this is because we never actually set up a cron job to populate the court dates table, but now i realise that maybe it was just that it was empty in my dev database - will check, and flag with nick if we need that job to be set up on production. I'm tempted to say that it's worth handling that case anyway though, just so we can fail gracefully if anything goes awry (otherwise we risk a 500 on a public facing page because an offline job has gone wrong). Re point (3), that is true, but that's what the CourtDates work was meant to solve, so if that's properly deployed it's not something we'll hit apart from if there's a failure somewhere along the line.

ds_judgements_public_ui/static/js/src/modules/date_input.js
timcowlishaw: aha, we have no js tests at all currently (and very little js). I agree that would be useful though - do you favour any particular JS testing framework / tool?

judgments/utils.py
timcowlishaw: Just need to check that doesn't choke if no year is provided for some reason, but yep, i think that'd be clearer

ds_judgements_public_ui/static/js/src/modules/date_input.js
timcowlishaw: aha i think that needs to be "if there are no other errors at all", which isn't ".badInput", it's ... another property who's name escapes me. I'll go back and check if the logic is correct here, and amend if not. If it is correct, will amend as you suggest. Good spot!

judgments/utils.py
timcowlishaw: aha yep, we need to explicitly handle that case or we get back "0000-1-1"

ds_judgements_public_ui/static/js/src/modules/date_input.js
timcowlishaw: aha no, you're entirely right. Fixed!

timcowlishaw:
hey @anthonyhashemi  - just pushed amends for most of your comments, and have commented on a few others - let me know if you're happy with the current state of it!

config/settings/base.py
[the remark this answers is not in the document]
dragon-dxw: A comment would probably be the best solution here.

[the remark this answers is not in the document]
anthonyhashemi: ah fair enough! ignore me then - comment could be useful as Dragon said

ds_judgements_public_ui/static/js/src/modules/date_input.js
anthonyhashemi: no preference, but Im familiar with jest. I've most recently been doing react where we used react-testing-library, but in this scenario I think we can use DOM Testing Library.

But again, no preference

judgments/utils.py
anthonyhashemi: ah fair enough! I imagine that was caught in one of the unit tests

judgments/templatetags/court_utils.py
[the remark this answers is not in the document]
anthonyhashemi: I do see your point, that since the CourtDates table is populated by an external cronjob, there is the possibility that the cronjob has never been run.

Another possibility is that that db has been nuked for some reason but I dont think we should account for that scenario in source code.

I do agree that court dates being wrong due to empty CourtDate object is not a big enough issue to give a 500 for though.

Can we at least add a log warning message so that we can look to see that the CourtDates db is empty for some reason - as thats obviously an issue.

TLDR: I now agree that this makes sense to keep. But it highlights the issue of having to have 2 related dbs that need to be kept in sync manually, and I think we should log a warning so we are aware of the bad db state.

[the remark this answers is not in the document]
timcowlishaw: aha, good plan!

ds_judgements_public_ui/static/js/src/modules/date_input.js
timcowlishaw: ok, good stuff - I'll take a look into Jest! do you mind if i add a ticket to improve JS test coverage and punt this conversation over there?

jacksonj04: If we're starting to push functionality into the JS then I'm a big fan of adding something like jest to prove that it's doing what we think.

If we could add this to both EUI and PUI at the same time (even if EUI literally just does a 'hello world' type test) that would be great, so they're using the same framework and config.

anthonyhashemi: Yeah sure thing @timcowlishaw, do let me know if you'd like some help with that too

anthonyhashemi (approved):

PR #656, opened 2023-04-25
Require query default value to be empty string, not None

dragon-dxw:
This was causing the following error
https://app.rollbar.com/a/dxw/fix/item/tna-caselaw-public-ui/166


anthonyhashemi (approved):
Happy for tests in a followup pr so we fix prod asap

jacksonj04 (approved):

PR #660, opened 2023-04-26
Ensure more thorough testing of meta robots tag on HTML pages

dragon-dxw:
Minor additions to testing to ensure correct behaviour.

Note that this does *not* extend to:
data.pdf
data.xml
and potentially other pages.

### DIFF
--- a/judgments/tests/tests.py
+++ b/judgments/tests/tests.py
@@ -33,6 +33,9 @@ class TestJudgment(TestCase):
         self.assertIn("(1.1\xa0GB)", decoded_response)
         # We don't use the Download as PDF text because there's an issue with localisated strings on CI
         self.assertEqual(response.status_code, 200)
+        self.assertIn(
+            '<meta name="robots" content="noindex,nofollow" />', decoded_response
+        )
 
     @patch("judgments.views.detail.requests.head")
     @patch("judgments.views.detail.decoder.MultipartDecoder")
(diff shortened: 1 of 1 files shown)


dragon-dxw:
[XML still showing up in bing search results](https://trello.com/c/3ZLbDsnl/848-xml-still-showing-up-in-bing-search-results)

judgments/tests/tests.py
```diff
@@ -206,6 +209,18 @@ def test_judgment_results(self, fake_result, fake_advanced_search):
             response, '<meta name="robots" content="noindex,nofollow" />'
         )
 
+    @patch("judgments.views.results.perform_advanced_search")
+    @patch("judgments.models.SearchResult.create_from_node")
+    def test_static_page(self, fake_result, fake_advanced_search):
```
jacksonj04, line 214: Should a static page have `noindex,nofollow`?

dragon-dxw: No, it is safe to index; this tests that it should not.

judgments/tests/tests.py
```diff
@@ -206,6 +209,21 @@ def test_judgment_results(self, fake_result, fake_advanced_search):
...
+        fake_advanced_search.return_value = fake_search_results()
+        fake_result.return_value = fake_search_result()
+        # The judgment search results page should have a robots meta tag
+        # with nofollow,noindex
+        response = self.client.get("/what-to-expect")
+        self.assertNotContains(
+            response, '<meta name="robots" content="noindex,nofollow" />'
```
jacksonj04, line 221: This might be better a bit fuzzier, just checking it doesn't contain `noindex`, so that if we make eg whitespace changes it doesn't break the test.

jacksonj04 (approved):
One quick non-critical fuzziness comment, otherwise yay ship it!

judgments/tests/tests.py
dragon-dxw: good plan!

PR #661, opened 2023-04-26
Set X-Robots-Tag response header to noindex,nofollow

dragon-dxw:
We add a middleware layer to add a `X-Robots-Tag` header to the outgoing response.

There's a bunch of tests to make sure that we're getting correct headers for a pdfs and XML as well.

### DIFF
--- a/ds_judgements_public_ui/templates/robots.txt
+++ b/ds_judgements_public_ui/templates/robots.txt
@@ -1,6 +1,16 @@
+User-Agent: Googlebot
+User-Agent: bingbot
+Allow: /
+Allow: /transactional-licence-form
+Allow: /what-to-expect
+Allow: /how-to-use-this-service
+Allow: /accessibility-statement
+Allow: /open-justice-licence
+Allow: /terms-of-use
+Disallow: /$
+
 User-Agent: *
 Disallow: /
-Allow: /$
 Allow: /transactional-licence-form
 Allow: /what-to-expect
 Allow: /how-to-use-this-service
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
         run: docker-compose run --rm django python manage.py migrate --settings=config.settings.test
 
       - name: Run Django Tests with coverage
-        run: docker-compose run django coverage run -m pytest
+        run: docker-compose run django coverage run -m pytest -m "not local"
 
       - name: Generate coverage XML
         run: docker-compose run django coverage xml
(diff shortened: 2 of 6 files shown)


dragon-dxw:
[XML still showing up in bing search results](https://trello.com/c/3ZLbDsnl/848-xml-still-showing-up-in-bing-search-results)

ds_judgements_public_ui/templates/robots.txt
```diff
@@ -1,6 +1,16 @@
+User-Agent: Googlebot
```
dragon-dxw, line 1: We need to let them in so we can tell them "no" in the headers. :(

ds_judgements_public_ui/templates/robots.txt
```diff
@@ -1,6 +1,16 @@
...
+User-Agent: bingbot
+Allow: /
+Allow: /transactional-licence-form
+Allow: /what-to-expect
+Allow: /how-to-use-this-service
+Allow: /accessibility-statement
+Allow: /open-justice-licence
+Allow: /terms-of-use
+Disallow: /$
+
```
anthonyhashemi, line 11: The reason we have 2 blocks, one for all other user-agents other than googlebot and bingbot, is that we trust that google and bing will respect no-index Robots metadata headers on our routes, and so they need to be able to access these:

"
A robots.txt file tells search engine crawlers which URLs the crawler can access on your site. This is used mainly to avoid overloading your site with requests; it is not a mechanism for keeping a web page out of Google. To keep a web page out of Google, block indexing with noindex or password-protect the page.
"

google ref: https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag

bing ref: https://www.bing.com/webmasters/help/block-urls-from-bing-264e560a#:~:text=The%20best%20way%20to%20block,not%20appear%20in%20search%20results.


Whereas we are not certain that over crawlers would respect the no-index metadata tags and so we want to block them from crawling everything except for documentation and the homepage, since the homepage has links to other judgments, which even though those other judgments now have the tag they might be respected and could index them? << @dragon-dxw I'm not quite sure about this sentence. Can you clarify for others please

dragon-dxw: That seems correct to me -- certainly the policy ScraperWiki had a decade ago was "just check robots.txt"; whether that's changing webstandards or better awareness of the complexities is uncertain.

If we discover there's another search engine which is a problem, we can think about taking the same approach, but this should solve the big three (Google, Bing, DuckDuckGo -- doesn't do its own scraping.)

anthonyhashemi (approved):

PR #665, opened 2023-04-27
Add dragon's magic Google link so we can configure how search engines work

dragon-dxw:
Allow dragon to configure the website (caselaw.nationalarchives.gov.uk) via https://search.google.com/search-console


anthonyhashemi (approved):
Add this for now but before we mark these cards as done we should set up a tna caselaw account - in case you're not around!

PR #666, opened 2023-04-28
Move PUI judgment detail view to use Judgment model

jacksonj04:
This moves retrieval of a single judgment (as either HTML or PDF) to use the new unified model from the API library, instead of hand-rolling the queries.

### DIFF
--- a/judgments/tests/factories.py
+++ b/judgments/tests/factories.py
@@ -1,0 +1,44 @@
+from typing import Any
+from unittest.mock import Mock
+
+from caselawclient.models.judgments import Judgment
+
+
+class JudgmentFactory:
+    # "name_of_attribute": ("name of incoming param", "default value")
+    PARAMS_MAP: dict[str, Any] = {
+        "uri": "test/2023/123",
+        "name": "Judgment v Judgement",
+        "neutral_citation": "[2023] Test 123",
+        "court": "Court of Testing",
+        "judgment_date_as_string": "2023-02-03",
+        "is_published": False,
+        "is_sensitive": False,
+        "is_anonymised": False,
+        "has_supplementary_materials": False,
+        "is_failure": False,
+        "source_name": "Example Uploader",
+        "source_email": "uploader@example.com",
+        "consignment_reference": "TDR-12345",
+        "assigned_to": "",
+        "versions": [],
+    }
+
+    @classmethod
+    def build(cls, **kwargs) -> Judgment:
+        judgment_mock = Mock(spec=Judgment, autospec=True)
+
+        if "html" in kwargs:
+            judgment_mock.return_value.content_as_html.return_value = kwargs.pop("html")
+        else:
+            judgment_mock.return_value.content_as_html.return_value = (
+                "<p>This is a judgment.</p>"
+            )
+
+        for param, value in cls.PARAMS_MAP.items():
+            if param in kwargs:
+                setattr(judgment_mock.return_value, param, kwargs[param])
+            else:
+                setattr(judgment_mock.return_value, param, value)
+
+        return judgment_mock()
--- a/judgments/tests/factories.py
+++ b/judgments/tests/factories.py
@@ -28,12 +28,18 @@ class JudgmentFactory:
     def build(cls, **kwargs) -> Judgment:
         judgment_mock = Mock(spec=Judgment, autospec=True)
 
-        if "html" in kwargs:
-            judgment_mock.return_value.content_as_html.return_value = kwargs.pop("html")
-        else:
-            judgment_mock.return_value.content_as_html.return_value = (
-                "<p>This is a judgment.</p>"
-            )
+        judgment_mock.return_value.content_as_html.return_value = kwargs.pop(
+            "html",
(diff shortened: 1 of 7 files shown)


judgments/views/detail.py
```diff
@@ -60,50 +52,54 @@ def get_best_pdf(request, judgment_uri):
...
         },
     )
 
 
 def detail_xml(_request, judgment_uri):
     try:
-        judgment_xml = api_client.get_judgment_xml(judgment_uri)
+        judgment = get_judgment_by_uri(judgment_uri)
+
+        if not judgment.is_published:
```
anthonyhashemi, line 93: similarly here, could we move

```
       if not judgment.is_published:
            raise Http404("This Judgment is not available")
```

out of the try except block?

anthonyhashemi (commented):
I'm halfway through reviewing - have reviewed the logic but not gone through tests yet

judgments/tests/factories.py
```diff
@@ -0,0 +1,52 @@
...
+        "assigned_to": ("assigned_to", ""),
+        "versions": ("versions", []),
+    }
+
+    @classmethod
+    def build(cls, **kwargs) -> Judgment:
+        judgment_mock = Mock(spec=Judgment, autospec=True)
+
+        if "html" in kwargs:
+            judgment_mock.return_value.content_as_html.return_value = kwargs.pop("html")
```
anthonyhashemi, line 32: Instead of the if else block, can we pass a second, `default`, argument to the pop method?

like so:

`judgment_mock.return_value.content_as_html.return_value = kwargs.pop("html", "<p>This is a judgment in HTML.</p>")`

judgments/tests/factories.py
```diff
@@ -0,0 +1,52 @@
...
+        else:
+            judgment_mock.return_value.content_as_html.return_value = (
+                "<p>This is a judgment in HTML.</p>"
+            )
+
+        if "xml" in kwargs:
```
anthonyhashemi, line 38: similarly here:

`judgment_mock.return_value.content_as_xml.return_value = kwargs.pop("xml", """<?xml version="1.0" encoding="UTF-8"?>
<judgment>
<p>This is a judgment in XML.</p><
/judgment>""")

judgments/tests/factories.py
```diff
@@ -0,0 +1,52 @@
...
+            judgment_mock.return_value.content_as_xml.return_value = kwargs.pop("xml")
+        else:
+            judgment_mock.return_value.content_as_xml.return_value = """<?xml version="1.0" encoding="UTF-8"?>
+<judgment>
+<p>This is a judgment in XML.</p><
+/judgment>"""
+
+        for map_to, map_from in cls.PARAMS_MAP.items():
+            if map_from[0] in kwargs:
+                setattr(judgment_mock.return_value, map_to, kwargs[map_from[0]])
```
anthonyhashemi, line 48: and here:

`setattr(judgment_mock.return_value, map_to, kwargs.get(map_from[0], map_from[1])`

judgments/tests/tests.py
```diff
@@ -4,90 +4,91 @@
...
-        decoded_response = response.content.decode("utf-8")
-        self.assertIn("data.pdf", decoded_response)
-        self.assertNotIn("2004_632.pdf", decoded_response)
-        # We don't use the Download as PDF text because there's an issue with localisated strings on CI
-        self.assertEqual(response.status_code, 200)
+    @patch("judgments.views.detail.get_pdf_size")
+    @patch("judgments.views.detail.get_judgment_by_uri")
+    @patch.dict(environ, {"ASSETS_CDN_BASE_URL": "https://example.com"})
+    def test_pdf_link_with_no_size(self, mock_judgment, mock_pdf_size):
+        if "ASSETS_CDN_BASE_URL" not in environ:
```
anthonyhashemi, line 81: if we have patched the environ dict in the decorator, won't `"ASSETS_CDN_BASE_URL" not in environ` always be false so this error will never be raised?

judgments/tests/tests.py
```diff
@@ -270,6 +271,44 @@ def test_no_referrer(self):
         self.assertIs(display_back_link(None), False)
 
 
+class TestGetPdfSize(TestCase):
```
anthonyhashemi, line 274: nice 👍

judgments/tests/tests.py
jacksonj04: Good catch

judgments/tests/test_factories.py
```diff
@@ -0,0 +1,46 @@
...
+class TestJudgmentFactory:
+    def test_default_uri(self):
+        # The default URI gets a test where others don't because without a URI judgments fall apart
+        judgment = JudgmentFactory.build()
+
+        assert type(judgment.uri) == str
+        assert judgment.uri != ""
+
+    @pytest.mark.parametrize(
+        "parameter, value",
```
anthonyhashemi, line 14: nice :

anthonyhashemi (approved):
Cheers for addressing the comments, looks good 👍

PR #681, opened 2023-05-04
FCL 29b: Ensure tests support new exceptions from custom-api-client

dragon-dxw:
Requires 7.0 of the custom-api-client

### DIFF
--- a/judgments/tests/tests.py
+++ b/judgments/tests/tests.py
@@ -1,5 +1,4 @@
 import re
-from os import environ
 from unittest import skip
 from unittest.mock import patch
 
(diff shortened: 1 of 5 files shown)


judgments/utils.py
```diff
@@ -212,4 +214,7 @@ def get_judgment_by_uri(judgment_uri: str) -> Judgment:
         use_https=settings.MARKLOGIC_USE_HTTPS,
     )
 
-    return Judgment(judgment_uri, api_client)
+    try:
+        return Judgment(judgment_uri, api_client)
+    except JudgmentNotFoundError:
+        raise Http404("Judgment not found")
```
jacksonj04, line 220: This doesn't feel like the right place to raise a 404 error; we might be using this utility for other things where we want JudgmentNotFound to bubble up.

judgments/tests/tests.py
```diff
@@ -44,10 +43,11 @@ def test_judgment_not_published_404_response(self, mock_judgment):
         self.assertIn("Page not found", decoded_response)
         self.assertEqual(response.status_code, 404)
 
-    @patch("judgments.views.detail.get_judgment_by_uri")
-    def test_judgment_not_found_404_response(self, mock_judgment):
-        mock_judgment.side_effect = MarklogicResourceNotFoundError()
-
+    @patch(
+        "caselawclient.models.judgments.Judgment.judgment_exists",
```
jacksonj04, line 47: This feels like we're patching the wrong thing; we're not testing that the model doesn't the right thing if the judgment doesn't exist, we're testing that the view does the right thing if the model tells us the judgment doesn't exist.

jacksonj04 (changes requested):
I'm not sure these changes quite nail the intent; they're moving the scope of where interactions happen in a way which works but which I think is blurring concerns.

jacksonj04 (approved):

PR #687, opened 2023-05-09
Tighten URL regex

dragon-dxw:
re_path checks if a URL matches a regular expression, but partial matches are still matches; examples typically are bracketted in `^`...`$` .

https://docs.djangoproject.com/en/4.2/ref/urls/#django.urls.re_path

We bracket our URLs this way so that `/eat/2022/1/press-summary/1` does not link to the judgment in future.

More controversially, we rewrite the URIs in some search results which were coming back as `/eat/2022/1.xml` to omit the trailing .xml which now fails to have a reverse (since that's not a valid website URI.)

~~This could do with a test.~~ Tested.

### DIFF
--- a/judgments/tests/test_detail.py
+++ b/judgments/tests/test_detail.py
@@ -30,6 +30,11 @@ class TestGetPublishedJudgment:
         with pytest.raises(Http404):
             get_published_judgment_by_uri("not-a-judgment")
 
+    def test_judgment_url_with_trailing_stuff_fails_to_resolve(self, client):
+        response = client.get("/eat/2022/1/kitten")
+        with pytest.raises(Resolver404):
+            response.resolver_match.func
+
 
 class TestJudgment(TestCase):
     @patch("judgments.views.detail.get_pdf_size")
--- a/judgments/tests/test_detail.py
+++ b/judgments/tests/test_detail.py
@@ -5,6 +5,7 @@ import pytest
 from caselawclient.errors import JudgmentNotFoundError
 from django.http import Http404
 from django.test import TestCase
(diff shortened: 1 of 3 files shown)


judgments/tests/test_detail.py
```diff
@@ -5,28 +5,29 @@
...
+from django.urls.exceptions import Resolver404
 from factories import JudgmentFactory
 
 from judgments.views.detail import get_pdf_size, get_published_judgment_by_uri
 
 
 class TestGetPublishedJudgment:
     @patch("judgments.views.detail.get_judgment_by_uri")
-    def judgment_is_published(self, mock_judgment):
+    def test_judgment_is_published(self, mock_judgment):
```
jacksonj04, line 16: \o/

judgments/urls.py
```diff
@@ -39,18 +39,20 @@
         name="feed",
     ),
     re_path(
-        r"(?P<judgment_uri>.*/\d{4}/\d+)/data.pdf",
+        r"^(?P<judgment_uri>.*/\d{4}/\d+)/data.pdf$",
```
jacksonj04, line 42: Is it worth pulling the `(?P<judgment_uri>.*/\d{4}/\d+)` into a "this is what a judgment URI segment looks like" constant?

jacksonj04 (approved):
Happy with this as it stands, nice spot on the unexpected behaviour and those tests! One DRY comment which might be worth solving but I don't feel strongly about.

judgments/urls.py
dragon-dxw: This is probably another commit/PR.

PR #695, opened 2023-05-12
Add space between snippet chunks

dragon-dxw:
This has been bugging me forever -- note the space between CAT 35 and LORD, or MACUR and MR

### before

image

### after

image

### before

image

### after

image

(Possibly this should be an ellipsis or an obvious non-text symbol, but I think this is a strong enough improvement to include as-is.)

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgments_listing.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgments_listing.scss
@@ -30,6 +30,7 @@
   &__matches {
     p {
       display: inline;
+      padding-right: 0.5rem;
     }
   }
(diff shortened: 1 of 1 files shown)


dragon-dxw:
[Spaces between different snippets](https://trello.com/c/jjTYBYBH/914-spaces-between-different-snippets)

jacksonj04 (approved):
Happy to include as is because it's an obvious improvement, but I'd like @Terry-Price to weigh in with ideas for how to make it clearer that these are actually different things (even if it's literally "stick a &middot; between them") rather than just a bit of double-spacing in the document.

PR #702, opened 2023-05-16
added a seprator to repeating search results text

Terry-Price:
## Changes in this PR:
Added a visual indicator to separate search results when there is more than one being returned within a judgment.
https://trello.com/c/dVboWSNL/918-pui-design-a-visual-separator-for-search-queries-that-appear-more-than-once-on-a-single-judgment
### Before
Screenshot 2023-05-16 at 14 55 41

### After
Screenshot 2023-05-16 at 14 55 05


dragon-dxw (changes requested):
You might want to remove line 33 which I added in the previous PR.

dragon-dxw (approved):

PR #705, opened 2023-05-18
Suppress WeasyPrint log spam

dragon-dxw:
Weasyprint tells us a lot of information about the internal workings of the PDF / CSS problems. We broadly don't care, and when people are crawling the back-catalog of PDFs they trigger many lines of logs per PDF hit, and we're hitting our Papertrail limit.

Sets an environment variable `SHOW_WEASYPRINT_LOGS` which should be set to True to permit WeasyPrint logs; this environment variable is optional.


dragon-dxw:
I don't believe there's a scenario where we actually care about these messages. Will add env variable anyway.

jacksonj04 (approved):
👍🏻

PR #712, opened 2023-05-22
Show errors when invalid dates are passed

timcowlishaw:


judgments/views/advanced_search.py
```diff
@@ -19,18 +19,28 @@
...
+    except ValueError as error:
+        to_date = None
+        errors.add_error(
+            "Errors in 'to' date - see below for details", "to_date", str(error)
+        )
+
+    if to_date is not None and from_date is not None and to_date < from_date:
+        errors.add_error(
+            "The 'to' date entered is before the 'from' date!",
+            "to_date",
```
dragon-dxw, line 41: does this need an f-string or similar?

PR #713, opened 2023-05-22
PUI - toolbar refine design

Donna-H:
## Changes in this PR:
Add judgment name and NCN to toolbar based on design, reposition the 'More download options' link
https://trello.com/c/0JalmFTE/925-pui-toolbar-refine-design-1st
### Before
old-mobile-toolbar
old-toolbar-pui-desktop

### After
new-mobile-toolbar-pui
new-toolbar-pui-desktop

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
@@ -9,13 +9,13 @@
     margin: auto;
     display: flex;
     flex-direction: column;
-    justify-content: space-between;
-    align-items: stretch;
+    justify-content: center;
+    align-items: center;
+    flex-wrap: wrap;
+    align-content: center;
 
-
-    @media (min-width: $grid__breakpoint-small) {
-      flex-direction: row;
-      align-items: flex-start;
+    > h1 {
+      color: black;
     }
 
     > div {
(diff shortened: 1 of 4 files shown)


ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
```diff
@@ -87,27 +86,24 @@
...
   margin: 0 auto;
 
-  @media (min-width: $grid__breakpoint-small) {
-    margin-left: auto;
-    margin-right: 0;
-  }
 
   &__option--pdf {
-    margin: 0 !important;
+    //margin: 0 !important;
```
timcowlishaw, line 91: If this isn't needed any more, we should remove it entirely rather than leaving it commented out

ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
```diff
@@ -87,27 +86,24 @@
...
     text-align: center;
   }
 
   &__download-options {
     color: $color__dark-grey;
     font-size: 0.75rem;
-    margin: calc($spacer__unit * 0.25) 0 0 0;
+    //margin: calc($spacer__unit * 0.25) 0 0 0;
```
timcowlishaw, line 98: Ditto - don't leave this commented out if we don't need it!

ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
```diff
@@ -87,27 +86,24 @@
...
 
     a {
       background-color: transparent;
       color: $color__aqua-blue;
-      padding: 0;
+      padding: 1px;
       display: inline;
       text-decoration: underline;
+      font-size: 16px;
```
timcowlishaw, line 106: Can we set this in `rem` rather than pixels?

ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
```diff
@@ -1,11 +1,11 @@
-{% load i18n %}
+{% load static i18n %}
```
timcowlishaw, line 1: What's the `static` for here? I can't see it being used in the changes below, but maybe i'm missing something?

judgments/views/detail.py
```diff
@@ -70,6 +70,8 @@ def detail(request, judgment_uri):
     context["judgment"] = judgment.content_as_html("")  # "" is most recent version
     context["page_title"] = judgment.name
     context["judgment_uri"] = judgment.uri
+    context["judgment_title"] = judgment.name
+    context["judgment_ncn"] = judgment.neutral_citation
```
timcowlishaw, line 74: 👏 👏 👏  👍

judgments/views/detail.py
```diff
@@ -70,6 +70,8 @@ def detail(request, judgment_uri):
     context["judgment"] = judgment.content_as_html("")  # "" is most recent version
     context["page_title"] = judgment.name
```
timcowlishaw, line 71: Not necessarily for now, but i'm wondering if we could remove this `page_title` and just use `judgment_title` everywhere it's used

timcowlishaw: (we might not actually be able to if eg it's used in the shared page layout)

PR #715, opened 2023-05-23
Assorted typography tightening on PUI

jacksonj04:
This addresses a few small annoyances in the typography, positioning etc on PUI.

No behavioural changes, just nicer to look at and a few bits which reduce the cognitive load for users.

### DIFF
--- a/judgments/tests/test_courts.py
+++ b/judgments/tests/test_courts.py
@@ -33,7 +33,7 @@ class TestCourtDatesHelper(TestCase):
     ):
         get.side_effect = CourtDates.DoesNotExist
         court = self.mock_court_dates(2011, 2012)
-        self.assertEqual(get_court_date_range(court), "2011 &ndash; 2012")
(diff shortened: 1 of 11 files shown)


judgments/tests/test_courts.py
```diff
@@ -33,7 +33,7 @@ def test_when_court_with_param_exists_and_no_dates_in_db_and_start_end_different
     ):
         get.side_effect = CourtDates.DoesNotExist
         court = self.mock_court_dates(2011, 2012)
-        self.assertEqual(get_court_date_range(court), "2011 &ndash; 2012")
+        self.assertEqual(get_court_date_range(court), "2011&nbsp;&mdash;&nbsp;2012")
```
timcowlishaw, line 36: I'm a little confused by these, as the implementation above seems to return 'XXXX to YYYY' instead of 'XXXX – YYYY' - do they still pass? if so it might be worth checking the right thing is being tested!

timcowlishaw (changes requested):
Ome small doubt otherwise LGTM! 🔥

PR #716, opened 2023-05-23
Fix for "back to search" always appearing

jacksonj04:


ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
```diff
@@ -1,8 +1,10 @@
 {% load i18n %}
 <div class="judgment-toolbar">
-  <div class="judgment-toolbar__return-link">
-    <a href="{{ context.back_link }}">{% translate "judgment.back" %}</a>
-  </div>
+  {% if context.back_link %}
```
anthonyhashemi, line 3: can we test this on the view where it is used?

```
        self.assertIn(back_link, rendered)
        self.assertIn("Back to search results", rendered)
```

jacksonj04: We can!

timcowlishaw (approved):
💥

PR #721, opened 2023-05-23
New/publication policy page

Terry-Price:
https://trello.com/c/OYVqgxKh/748-fcl-policies-all-ducks-in-a-row-%F0%9F%A6%86
### Before
no page existed
### After
new page

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -49,6 +49,11 @@ urlpatterns = [
         name="terms_of_use",
     ),
     path(
+        "publishing-policy",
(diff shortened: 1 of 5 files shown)


jacksonj04 (changes requested):
I think this should hold off until we've got the unified static content pages in; it'll be easier solving the merge of this into that than vice-versa.

timcowlishaw (approved):
LGTM, and have fixed the merge conflicts! @Terry-Price it'd be worth a look on staging to check that nothing went awry in the merge.

PR #722, opened 2023-05-23
Harmonise the PUI containers and headers

jacksonj04:
**Watch out:** Merge #719 first!

The PUI has built up a range of different implementations of containers and headers across a range of templates. This means some pages behave differently, and containers can misbehave (sometimes on the same pages as ones behaving as expected), especially around breakpoints.

This PR simplifies the SCSS (and generated CSS) used to define container widths and padding, and makes all containers behave the same way.

It also tidies up and simplifies layout templates to improve reuse and reduce complexity, bringing all static pages into the same template for consistency, always generating titles and breadcrumb context in the same place (ie in the view).

Together these changes make the site more consistent and easier to maintain.


timcowlishaw (approved):
Looks good to me (lots of changes that are i think due to prettier being added to the linter? but good to have them all in one place in any case)

PR #725, opened 2023-05-24
Fix/about caselaw page update

Terry-Price:
- Made changes and additions to the footer 
- Created a 'Privacy Notice' page – not linked from footer yet as it needs Director level sign off, but when it is, it will sit in leagl column under the Cookies link.
- Made wholesale changes to the 'What to expect page' and renamed it 'About this service'.
- Took link out of page header and moved it to the footer under 'About' column.

https://trello.com/c/OYVqgxKh/748-fcl-policies-all-ducks-in-a-row-%F0%9F%A6%86
### After


timcowlishaw (commented):
I've just had to make a fair few changes to this myself as a result of rebasing on master, so will defer to @jacksonj04 @anthonyhashemi or @dragon-dxw for explicit approval!

jacksonj04 (approved):
I'm happy with this as far as the code goes. I can't speak to the substance of the text, but nothing looks obviously wrong.

PR #726, opened 2023-05-25
Fix for gutters and vertical pacing on mobiles

jacksonj04:
A bunch of assorted CSS/layout fixes and tweaks which improve page gutters and vertical pacing, particularly in mobile views.

### Home

Home Old

Home New

### Static content

HTU Old

HTU New

### Footer

HTU Foot Old

HTU Foot New


Terry-Price (approved):
You might get a conflict as I added a footer change earlier today to add more content Nick.

PR #728, opened 2023-05-25
add court date ranges alongside court filters

Donna-H:
## Changes in this PR:
Displaying court date ranges in context alongside the existing and new (hierarchical) court filters on about page, browse and structured search.
https://trello.com/c/tLHaCVSW/757-%F0%9F%94%8D-pui-browse-displaying-court-date-ranges-in-context-alongside-search-filters-but-see-857
### Before
about-before
browse-before
structure-before

### After
about-after
browse-after
structure-after
results-filter

### DIFF
--- a/ds_judgements_public_ui/templates/includes/courts_options.html
+++ b/ds_judgements_public_ui/templates/includes/courts_options.html
@@ -9,7 +9,7 @@
              {% for alias in context.court %}{% if alias in court.param_aliases %}checked="checked"{% endif %}
              {% endfor %} />
       <span class="structured-search__court-option-label">{{ court.name }}
-        <span style="color:#555; font-size: 0.9rem">&mdash; {{ court|get_court_date_range }}</span>
+        <span class="structured-search__court-date-range">&mdash; {{ court|get_court_date_range }}</span>
       </span>
     </label>
   </div>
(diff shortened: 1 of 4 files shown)


ds_judgements_public_ui/templates/includes/courts_options.html
```diff
@@ -7,7 +8,9 @@
              id="court_{{ court.canonical_param }}"
              {% for alias in context.court %}{% if alias in court.param_aliases %}checked="checked"{% endif %}
              {% endfor %} />
-      <span class="structured-search__court-option-label">{{ court.name }}</span>
+      <span class="structured-search__court-option-label">{{ court.name }}
+        <span style="color:#555; font-size: 0.9rem">&mdash; {{ court|get_court_date_range }}</span>
```
timcowlishaw, line 12: could we add this style in the appropriate SCSS file (`_structured_search.scss`) and add a class here (following BEM conventions, something like `structured-search__court-date-range` would be appropriate) , rather than using inline styles?

 It'd also be good to use a named colour variable (that one is $color__dark-grey) rather than hardcoding the colour

timcowlishaw (changes requested):
Looks good to me Donna, and well done figuring out where the dates come from in  django! One small code standards request in a comment but apart from that, all good.

Can I check, did you test that the court dates show correctly in the results filters as well as the structured search page? it should just be reusing the same template so no extra work there, would just like to be 100% sure.

ds_judgements_public_ui/templates/includes/courts_options.html
Donna-H: @timcowlishaw  all done and added the image for results filter

PR #737, opened 2023-05-30
Tidy up heading typography

jacksonj04:
This reunifies H1-3 styles across the site using new calculated proportions – H1 is a multiple of a base size, H2 and H3 are multiples of H1 size, line height and margins are related to the font size, scaling on medium views and down is a simple multiplier as well.


timcowlishaw (approved):
Lovely stuff 😍

PR #738, opened 2023-05-30
Use new search methods from api client

anthonyhashemi:
## Changes in this PR:
- Update api client to 9.0.0
- Update all search and parsing functionality due to the changes in api client
- Remove classes and functions that were extracted into api client
- Specifically use `search_judgments_and_parse_response` method from client which takes the old searh functionality one step further, by filtering by `judgments` collection

       - Note: because we now filter by the `judgments` collection, before merging this in and deploying to production we need to update all judgments (which should be all current documents) on staging and production respectively to be in the `judgments` collection.
       
To do this for one document the xquery is:

```
xquery version "1.0-ml";
let $collections := ("judgments")

return xdmp:document-add-collections(
   "URI_OF_DOC.xml",
   $collections
)
```

But we will want to do it for the whole caselae-content `module`.

Since that needs to be done for 92618 documents on staging and 113063 documents on production, we will need to make sure that we don't impact performance of the service in a non managed way.

The options I found for the bulk processing are the following:
1. xquery for loop in marklogic console
2. Data Movement SDK
3. MLCP (MarkLogic Content Pump)
4. [CoRB](https://github.com/marklogic-community/corb2)

From these options CoRB seems like the best combination of ease of use and performance since it abstracts away the low level boiler plate of batching and parallelization but is more efficient and failproof than just using the for loop in xquery. MLCP is probably a decent alternative.

https://trello.com/c/T6x6GXmc/803-ensure-judgments-dont-display-press-summaries-erroneously


judgments/feeds.py
```diff
@@ -65,16 +70,18 @@ class LatestJudgmentsFeed(Feed):
...
-        court_query = "/".join(filter(lambda x: x is not None, [court, subdivision]))
-        slugs = filter(lambda x: x is not None, [court, subdivision, year])
-        slug = "/".join([str(s) for s in slugs])
         try:
             page = int(request.GET.get("page", 1))
         except ValueError:
             # e.g. the user provided ?page= or ?page=jam
             raise Http404
+
+        court_query = "/".join(filter(lambda x: x is not None, [court, subdivision]))
```
anthonyhashemi, line 79: just moving these below the try except so we error out quicker

judgments/views/advanced_search.py
```diff
@@ -66,25 +68,27 @@ def advanced_search(request):
...
-        model = perform_advanced_search(
+        search_parameters = SearchParameters(
             query=query_without_stop_words,
             court=query_params["court"],
             judge=query_params["judge"],
             party=query_params["party"],
             neutral_citation=query_params["neutral_citation"],
             specific_keyword=query_params["specific_keyword"],
-            page=page,
+            page=int(page),
```
anthonyhashemi, line 78: [mypy] the new SearchParameters dataclass in apiclient expecting a int

judgments/views/advanced_search.py
```diff
@@ -66,25 +68,27 @@ def advanced_search(request):
...
             order=order,
             date_from=query_params["from"],
             date_to=query_params["to"],
-            per_page=per_page,
+            page_size=int(per_page),
```
anthonyhashemi, line 82: [mypy] the new SearchParameters dataclass in apiclient expecting a int

jacksonj04 (approved):
Nice, so much less code to look after! The drop in test coverage isn't a real drop so not worried there.

I suspect (as always) that there's more we can clean up (eg some of the testing is still a bit boilerplate), but that only really showed up because you touched the code so it's not relevant to this refactor.

🚢

anthonyhashemi:
> Nice, so much less code to look after! The drop in test coverage isn't a real drop so not worried there.
> 
> I suspect (as always) that there's more we can clean up (eg some of the testing is still a bit boilerplate), but that only really showed up because you touched the code so it's not relevant to this refactor.
> 
> 🚢

@jacksonj04 yeah agreed - there was quite a bit i saw that I wanted to fix up in followup prs but this piece of work has taken a long time as is so will leave for now

PR #740, opened 2023-05-31
Add management command to recalculate court dates

jacksonj04:
This adds a new command to recalculate the date ranges covered by each court. It relies on the existing search functionality, grabbing results for each court sorted once by date ascending, once by date descending.

This does _not_ add automation, as running this command may expose data errors which we ideally want to fix first. Once errors have been resolved, we can run the command using `--write` to commit the data, and then we should open a new PR which adds a scheduled job (probably daily) to recalculate ranges.

### DIFF
--- a/judgments/admin.py
+++ b/judgments/admin.py
@@ -1,3 +1,8 @@
-# from django.contrib import admin
+from django.contrib import admin
 
-# Register your models here.
+from judgments.models import CourtDates
+
+
+@admin.register(CourtDates)
+class CourtDatesAdmin(admin.ModelAdmin):
+    pass
--- a/judgments/management/commands/recalculate_court_dates.py
+++ b/judgments/management/commands/recalculate_court_dates.py
@@ -19,7 +19,6 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args, **options):
-        CourtDates.objects.all().delete()
         for court in courts.get_all():
             self.stdout.write(self.style.NOTICE(f"{court.name}"))
 
--- a/judgments/management/commands/recalculate_court_dates.py
+++ b/judgments/management/commands/recalculate_court_dates.py
@@ -1,0 +1,106 @@
+import datetime
+
+from django.core.management.base import BaseCommand
+from ds_caselaw_utils import courts
+
+from judgments.models import CourtDates, SearchResult
+from judgments.utils import perform_advanced_search
+
+
+class Command(BaseCommand):
+    help = "Recalculates the date ranges for known courts"
+
+    def add_arguments(self, parser):
+        # Named (optional) arguments
+        parser.add_argument(
+            "--write",
+            action="store_true",
+            help="Write changes to the database",
+        )
+
+    def handle(self, *args, **options):
+        CourtDates.objects.all().delete()
+        for court in courts.get_all():
+            self.stdout.write(self.style.NOTICE(f"{court.name}"))
+            if court.canonical_param:
+                search_results = perform_advanced_search(
+                    court=court.canonical_param,
+                    order="date",
+                    per_page=1,
+                )
+
+                oldest_document = SearchResult.create_from_node(
+                    search_results.results[0]
+                )
+
+                if oldest_document.date:
+                    start_year = oldest_document.date.year
+                    self.stdout.write(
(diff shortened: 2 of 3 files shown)


judgments/admin.py
```diff
@@ -1,3 +1,8 @@
-# from django.contrib import admin
+from django.contrib import admin
 
-# Register your models here.
+from judgments.models import CourtDates
+
+
+@admin.register(CourtDates)
+class CourtDatesAdmin(admin.ModelAdmin):
```
anthonyhashemi, line 7: Currently, since the class isnt doing anything different to thee base model admin, this isn't doing anything that the default admin setup would do anyway? If so, I'd suggest removing it, but maybe I am missing something

judgments/management/commands/recalculate_court_dates.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def add_arguments(self, parser):
+        # Named (optional) arguments
+        parser.add_argument(
+            "--write",
+            action="store_true",
+            help="Write changes to the database",
+        )
+
+    def handle(self, *args, **options):
```
dragon-dxw, line 21: Inclined to agree with the linter here, this is a pretty gnarly function. I think a separate function which attempts to repair and report on problems might clear this up a lot.

dragon-dxw (approved):

judgments/management/commands/recalculate_court_dates.py
anthonyhashemi: suggested refactor:

```
import datetime

from django.core.management.base import BaseCommand
from ds_caselaw_utils import courts

from judgments.models import CourtDates, SearchResult
from judgments.utils import perform_advanced_search


class Command(BaseCommand):
    help = "Recalculates the date ranges for known courts"

    def add_arguments(self, parser):
        # Named (optional) arguments
        parser.add_argument(
            "--write",
            action="store_true",
            help="Write changes to the database",
        )

    def handle(self, *args, **options):
        CourtDates.objects.all().delete()
        
        for court in courts.get_all():
            self.stdout.write(self.style.NOTICE(f"{court.name}"))
            
            if not court.canonical_param:
                self.stdout.write(
                    self.style.ERROR(f"{court.name} has no canonical_param! Skipping.")
                )
                continue

            start_year = self.get_start_year(court)
            end_year = self.get_end_year(court)

            if not options["write"]:
                self.stdout.write(self.style.NOTICE("Skipping write…"))
                continue
            
            CourtDates.objects.update_or_create(
                param=court.canonical_param,
                defaults={"start_year": start_year, "end_year": end_year},
            )


    def get_start_year(self, court):
        start_year = self._get_year_of_first_document_in_order(self, court.canonical_param, "date", "newest", court.start_year)
        if start_year < 2000:
            self.stdout.write(
                self.style.WARNING(
                    f"Calculated start year of {start_year} seems improbable,\
                        falling back to config value of {court.start_year}"
                )
            )
            start_year = court.start_year
            
        return start_year

    def get_end_year(self, court):
        end_year = self._get_year_of_first_document_in_order(self, court.canonical_param, "-date", "newest", court.end_year)

        if end_year > datetime.date.today().year:
            self.stdout.write(
                self.style.WARNING(
                    f"Calculated end year of {end_year} is impossible,\
                        falling back to config value of {court.end_year}"
                )
            )
            end_year = court.end_year
            
        return end_year


    def _get_year_of_first_document_in_order(self, canonical_param, order, first, fallback):
        search_results = perform_advanced_search(
            court=canonical_param,
            order=order,
            per_page=1,
        )

        first_document = SearchResult.create_from_node(
            search_results.results[0]
        )

        if first_document.date:
            year = first_document.date.year
            self.stdout.write(
                self.style.NOTICE(
                    f"{first.capitalize()} document: {first_document.uri} @ {first_document.date.year}"
                )
            )
        else:
            year = fallback
            self.stdout.write(
                self.style.WARNING(
                    f"Couldn't find date of {first} document {first_document.uri},\
                        falling back to config value of {fallback}"
                )
            )
        return year
```

judgments/admin.py
jacksonj04: You need to do the `@admin.register` bit for the model to appear in the admin.

PR #742, opened 2023-06-01
Add visual marker date errors

Donna-H:
## Changes in this PR:
Added to validation visual marker for error states -using the red GDS side border marker
https://trello.com/c/MRQqUFei/935-pui-form-validation-visual-marker-for-error-states
### Before
before-update-error

### After
update-error-after

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_structured_search.scss
+++ b/ds_judgements_public_ui/sass/includes/_structured_search.scss
@@ -173,7 +173,6 @@
     color: #d4351c;
     font-weight: 700;
     font-size: 0.8rem;
-    #display: none;
   }
 
   &__to-date-error-message {
@@ -182,7 +181,6 @@
     color: #d4351c;
     font-weight: 700;
     font-size: 0.8rem;
-    #display: none;
   }
 
   input:invalid {
@@ -307,6 +305,18 @@
     border: none;
     padding: 0;
     margin: 0;
+  }
+
+  &__multi-fields-panel {
+    fieldset.set-1 {
+      padding-left: 20px;
+      border-left: 4px solid #d4351c;
+    }
+
+    fieldset.set-2 {
+      padding-left: 20px;
+      border-left: 4px solid #d4351c;
+    }
   }
 
   &__update-filters-button {
--- a/ds_judgements_public_ui/sass/includes/_structured_search.scss
+++ b/ds_judgements_public_ui/sass/includes/_structured_search.scss
@@ -308,13 +308,13 @@
   }
 
   &__multi-fields-panel {
-    fieldset.set-1 {
-      padding-left: 20px;
+    fieldset.with-errors {
+      padding-left: calc($spacer__unit * 1.5);
       border-left: 4px solid #d4351c;
     }
 
-    fieldset.set-2 {
-      padding-left: 20px;
+    fieldset.with-errors {
+      padding-left: calc($spacer__unit * 1.5);
       border-left: 4px solid #d4351c;
     }
   }
--- a/ds_judgements_public_ui/templates/includes/results_filters_inputs.html
+++ b/ds_judgements_public_ui/templates/includes/results_filters_inputs.html
@@ -11,7 +11,7 @@
 <div class="structured-search__multi-fields-panel">
   <div class="structured-search__limit-to-container">
     {% if context.errors|has_errors_for_field:"from_date" %}
-      <fieldset class="set-1">
+      <fieldset class="{% if context.errors %}with-errors{% endif %}">
       {% endif %}
       <legend>
         <span class="structured-search__limit-to-label" id="from_date_label">From date</span>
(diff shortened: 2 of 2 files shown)


timcowlishaw (changes requested):
Aha either the `validation-errors` branch needs to be merged to main first, or this needs to be merged into that branch, as this is bringing with it a whole load of other work that isn't yet merged!

In addition I think we'll need to rebase on the latest  state of the validation-errors work beforehand - @Donna, I'll take you through how to do this next week!

ds_judgements_public_ui/templates/includes/results_filters_inputs.html
```diff
@@ -10,7 +10,7 @@
 </div>
 <div class="structured-search__multi-fields-panel">
   <div class="structured-search__limit-to-container">
-    <fieldset>
+    {% if context.errors %}<fieldset class="set-1">{% endif %}
```
timcowlishaw: A tiny stylistic thing, but i think i'd prefer the condition to be *inside* the class here, to make it clear what's actually varying and cut down on duplication, so for instance:

```
<fieldset class="{% if context.errors %}set-1{% endif %}">
```

(although remember that the classname will be different, see my comment above)

timcowlishaw (changes requested):
A few more stylistic things to address, but as in previous comment, please don't merge yet! we'll rebase and sort out on monday.

ds_judgements_public_ui/templates/includes/results_filters_inputs.html
```diff
@@ -10,104 +10,112 @@
 </div>
 <div class="structured-search__multi-fields-panel">
   <div class="structured-search__limit-to-container">
-    <fieldset>
+    {% if context.errors|has_errors_for_field:"from_date" %}
+      <fieldset class="set-1">
```
anthonyhashemi, line 14: We verified that this worked locally, but when this is false the html will have a closing fieldset tag without an opening one which of course isn't best practice.

Like I said we verified that it generally looked fine in the UI with the code like this, but we have in fact removed fieldset from the non error path, which means that either fieldset was redundant before or that it had some other importance that wasn't necessarily visual.




We thought it might be possible to have the conditional statement inside the `<fieldset>` tag, but that broke the html /not rendering anything after this line when the condition was true. Maybe we did it wrong, or is that just not allowed in django templates?

An alternative to this, that definitely would work, would  be duplicating the whole fieldset here and wrapping the if block around that and then adjusting the else blocks fieldset to have no `class=set-1` (we did try this initially but got messy due to the fieldset being large and we got a bit tangled, but should be fine if we do it carefully)

ds_judgements_public_ui/templates/includes/results_filters_inputs.html
anthonyhashemi: oh so _this_ is how you do it!

we were doing `<fieldset {% if context.errors %}class="set-1"{% endif %}">`

ds_judgements_public_ui/templates/includes/results_filters_inputs.html
anthonyhashemi: just noticed that @timcowlishaw has given the correct form of the conditional within the tag itself in the above comment!

anthonyhashemi:
> Aha either the `validation-errors` branch needs to be merged to main first, or this needs to be merged into that branch, as this is bringing with it a whole load of other work that isn't yet merged!
> 
> In addition I think we'll need to rebase on the latest state of the validation-errors work beforehand - @Donna, I'll take you through how to do this next week!

Worth us adjusting the base branch from main to your branch @timcowlishaw ? Will help focus the diff in this PR also

timcowlishaw:
Aha yes please although I think it *may* need to be hard rebased from origin first, as we may have prematurely squashed some commits. Shall I have a look, as I know what changes to look out for?

ds_judgements_public_ui/templates/includes/results_filters_inputs.html
Donna-H: @timcowlishaw @anthonyhashemi updated to`<fieldset class="{% if context.errors %}with-errors{% endif %}">`

ds_judgements_public_ui/sass/includes/_structured_search.scss
```diff
@@ -309,6 +307,18 @@
     margin: 0;
   }
 
+  &__multi-fields-panel {
+    fieldset.set-1 {
```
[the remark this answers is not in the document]
Donna-H: @timcowlishaw updated

ds_judgements_public_ui/sass/includes/_structured_search.scss
```diff
@@ -309,6 +307,18 @@
...
+      padding-left: 20px;
```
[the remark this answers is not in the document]
Donna-H: updated @timcowlishaw

PR #746, opened 2023-06-05
Label exact matches in search results

timcowlishaw:
## Changes in this PR:

This adds a subtle highlight to results where the query term matches the title or ncn, modulo case, punctuation and stopwords. It also adds 'v', 'vs' and '&' to the query stopwords, to reduce noise in the result preview and allow for a little more querying flexibility.

In addition, this subtley restyles the highlights in the 'matching text snippet' in the results list, so that they're not too prominent - replacing the yellow highlight with a yellow underline maintains the visual link (yellow highlight for matching text), but without making the page too busy).

https://trello.com/c/e8N4LhfQ/902-pui-recovering-from-0-results-simple-changes-first

### Before
Screenshot 2023-06-05 at 19 26 10

### After
Screenshot 2023-06-05 at 19 25 47


Terry-Price (changes requested):
Hi Tim,
Just wondered on the use of italic text?
Is it for decoration, as usually `<em>` is used for emphasis in the HTML which renders it italic. But it doesn't have to be italic to have the emphasis tag (which might be better for screen reader users that need it highlighted to them as the styling will not be enough.
It's only my own personal preference, but find italics on the screen a little harder to read.

judgments/tests/test_search_results_filters.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def test_when_the_name_matches_the_query_but_differs_in_case(self):
+        result = buildResult()
+        query = "alice xenakis v bob young"
+        self.assertTrue(is_exact_match(result, query))
+
+    def test_when_the_name_matches_the_ncn_exactly(self):
+        result = buildResult()
+        query = "[2023] EWHC 1234 (Ch)"
+        self.assertTrue(is_exact_match(result, query))
```
dragon-dxw, line 33: It might be worth also checking that is_exact_ncn_match is also true, since we do care about that behaviour.

judgments/templatetags/search_results_filters.py
```diff
@@ -0,0 +1,37 @@
...
+
+not_alphanumeric = re.compile("[^a-zA-Z0-9]")
+
+
+def preprocess_title(string):
+    return preprocess_query(string).lower()
+
+
+def preprocess_ncn(string):
+    return re.sub(not_alphanumeric, "", preprocess_query(string).lower())
```
dragon-dxw, line 17: Wondering if we want to normalise spaces rather than removing them -- we might highlight matches which split over word boundaries (Manbat v Man == Man v Batman).

Possibly not -- the search filter should remove those, and "BBC" and "B. B. C." matching is good.

dragon-dxw (approved):

PR #749, opened 2023-06-07
Use seperate fields for year/month/day in date selectors

timcowlishaw:
## Changes in this PR:

This is a further development on the new date input work: pass the query string parameter text_date_input=True on the structured search page or search results page to reveal the new date style -  This is done in line with the [GDS date select component](https://design-system.service.gov.uk/components/date-input/) and [GDS style guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style).

This has also required the introduction of some form validation logic (the bulk of this PR)  - after an abortive attempt to move the whole thing to django forms which caused various problems, I've rolled-our-own minimal form validation error class, as well as making some necessary changes to the logic for adding/removing date filters.

I think this is now ready for merging - it's behind a flag so that we can manage release.

Possible future enhancements:

- Finessing the message wording (with some help from a copywriter - this'll be quick because they're all in translations)
- A more general refactor of the advanced search view which i think is a bit of a project in itself.

Tim

https://trello.com/c/vsN1NKLu/833-pui-show-validation-errors-on-the-advanced-search-form


ds_judgements_public_ui/templates/includes/results_search_component.html
```diff
@@ -1,5 +1,7 @@
 <div class="results-search-component">
-  <form action="{% url 'advanced_search' %}" id="analytics-results-filters">
+  <form action="{% url 'advanced_search' %}"
+        id="analytics-results-filters"
+        novalidate="novalidate">
```
anthonyhashemi, line 4: if novalidate wasnt specified, what would happen?

ds_judgements_public_ui/templates/pages/structured_search.html
```diff
@@ -1,22 +1,27 @@
...
   {% translate "structured_search.title" %} - {% translate "common.findcaselaw" %}
 {% endblock title %}
 {% block content %}
   <form action="{% url "advanced_search" %}"
         role="search"
+        novalidate="novalidate"
         id="analytics-structured-search"
-        method="get">
+        method="get"
+        novalidate="novalidate">
```
anthonyhashemi, line 12: there are 2 `novalidate`s here

judgments/templatetags/query_filters.py
```diff
@@ -16,11 +16,46 @@ def make_query_string(params):
     return "&".join(pairs)
 
 
+@register.filter
+def removable_filter_param(key):
```
anthonyhashemi, line 20: can you give some more context around what a `removable_filter_param` is?

ds_judgements_public_ui/templates/pages/structured_search.html
timcowlishaw: Fixed! good spot, thank you!

ds_judgements_public_ui/static/js/src/modules/date_input.js
```diff
@@ -22,5 +22,5 @@ function setupDateInput() {
 }
 
 $(() => {
-    $(".js-date-input").each(setupDateInput);
+    //$(".js-date-input").each(setupDateInput);
```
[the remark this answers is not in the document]
timcowlishaw: absolutely!

judgments/templatetags/query_filters.py
timcowlishaw: aha, so theses are the parameters in the query which are 'removable' using the little yellow 'filter control boxes' on the search page. some elements of the query (court, dates, query term) should be able to be removed to widen the search, others (per_page, order, feature flag state) should be ignored. This is a test to check that a parameter should indeed be displayed as a 'remove' button.

judgments/models.py
```diff
@@ -20,3 +22,20 @@ def min_year():
     def max_year():
         result = CourtDates.objects.aggregate(Max("end_year"))
         return result["end_year__max"]
+
+
+class SearchFormErrors:
```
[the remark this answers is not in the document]
timcowlishaw: aha,three things:

- I could not find any way of replicating the pattern we have here with dates (three seperate fields which produce one value, the combined value is validated as a whole), which was the requirement driving this change. [This](https://crispy-forms-gds.readthedocs.io/en/stable/components/date.html) provided a moment of hope but still didn't appear to actually do what's advertised.
- Django forms seemed to me to be very tightly coupled to CRUD operations on django models - adapting it to validate a simple set of parameters involved a lot of boilerplate (i know realise that encapsulating the query as a model-like object without persistence might actually have been a quite neat refactor that would solve this, but that'd be a bigger undertaking)
- By default django forms makes itself responsible for generating a lot of the form html itself, adapting this to our current design (or rather, styling the django forms html to match our current design) would have been a fairly major rewrite

judgments/templatetags/query_filters.py
```diff
@@ -16,11 +16,46 @@ def make_query_string(params):
...
+        "from_year",
+        "to_day",
+        "to_month",
+        "to_year",
+        "text_date_input",
+    ]
+    return key not in excluded
+
+
+def date_filter_param(key):
```
dragon-dxw, line 36: Could you add something describing what the intention is here? I'm not sure that these are `param`s in the same sense as `to_month` above; one is input, the other a gestalt entity.

I think I get it -- if a 'date filter parameter' is removed, we remove all the subcomponents -- but this is quite subtle and might appreciate a little documentation.

dragon-dxw (approved):

judgments/templatetags/query_filters.py
anthonyhashemi: I see!

judgments/models.py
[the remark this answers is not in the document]
anthonyhashemi: I see, then I agree this choice makes sense 👍

PR #757, opened 2023-06-09
Fix applied court filters bug

anthonyhashemi:
## Changes in this PR:
- Fixed bug where court filters were not displaying properly in the search results page.

Bug was introduced in: https://github.com/nationalarchives/ds-caselaw-public-ui/pull/738

https://trello.com/c/E31VtoeZ

### Before
Screenshot 2023-06-09 at 12 12 17

### After
Screenshot 2023-06-09 at 12 11 39

### DIFF
--- a/judgments/views/advanced_search.py
+++ b/judgments/views/advanced_search.py
@@ -37,7 +37,7 @@ def advanced_search(request):
 
     query_params = {
         "query": params.get("query", ""),
-        "court": ",".join(params.getlist("court")),
+        "court": params.getlist("court"),
         "judge": params.get("judge"),
         "party": params.get("party"),
         "neutral_citation": params.get("neutral_citation"),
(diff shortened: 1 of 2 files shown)


judgments/views/advanced_search.py
```diff
@@ -70,7 +70,7 @@ def advanced_search(request):
         query_without_stop_words = preprocess_query(query_params["query"])
         search_parameters = SearchParameters(
             query=query_without_stop_words,
-            court=query_params["court"],
+            court=",".join(query_params["court"]),
```
anthonyhashemi, line 73: This stems from: 
https://github.com/nationalarchives/ds-caselaw-custom-api-client/blob/main/src/caselawclient/search_parameters.py#L12

where `SearchParameters` expects court to be of type `str` or `None`

This is currently misleading, as there is still the legacy code on https://github.com/nationalarchives/ds-caselaw-custom-api-client/blob/main/src/caselawclient/search_parameters.py#L63

which means that the api client search can handle court as a list or a string.

This view was the only place where it was actually utilising the ability to pass in a list, but in https://github.com/nationalarchives/ds-caselaw-public-ui/pull/738 I changed it to be a string since mypy was complaining about the above.

Hadn't realised that the query_params object was being used in the view through the context!

This change fixes the bug and complies with search_params.

We should probably revist this, and get rid of the apiclients ability to accept list or string since that is confusing, and then come back to PUI and make sure this is still working as expected.

PR #762, opened 2023-06-12
Move 'back to search results' link into breadcrumbs and expose context

jacksonj04:
Where the `HTTP_REFERER` is a search link, extract the search context from it and display it in the breadcrumbs bar for a judgment.

This replaces the existing behaviour where a straightforward "back to search results" link is displayed in the judgment header.

### Before

localhost_3000_ewca_civ_2004_575 (1)

### After

localhost_3000_ewca_civ_2004_575


timcowlishaw (approved):
💥

PR #763, opened 2023-06-13
Fix open redirect in resolving URIs with trailing slashes

dragon-dxw:
An open redirect was present on the site where you could use a URL starting with `nationalarchives.gov.uk` to redirect to arbitrary locations on the internet. This was due to a badly implemented redirect function which wasn't sense-checking the URL.

The fix was to remove this behaviour and instead rely on the presence of an optional trailing slash in the resolver for judgment URIs. This fixes the vulnerability, but has two side effects:

- Other URLs on the site which could previously be accessed with a trailing slash become inaccessible. This was judged to be low impact, as it affects a very small portion of resources on the site which wouldn't usually be linked to with a trailing slash.
- Two distinct resources are now presented for each judgment, one with a trailing slash and one without.

A resolution for the second one now checks the requested judgment URI (which may still resolve due to support for historical cases where URIs were presented with a trailing slash) against the canonical one, and redirects the user if they accessed the judgment with a non-canonical URI.

### DIFF
--- a/judgments/views/detail.py
+++ b/judgments/views/detail.py
@@ -5,7 +5,7 @@ import requests
 from caselawclient.errors import JudgmentNotFoundError
 from caselawclient.models.judgments import Judgment
 from django.conf import settings
-from django.http import Http404, HttpResponse
+from django.http import Http404, HttpResponse, HttpResponseRedirect
 from django.shortcuts import redirect
 from django.template.defaultfilters import filesizeformat
 from django.template.response import TemplateResponse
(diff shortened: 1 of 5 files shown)


judgments/views/detail.py
```diff
@@ -69,6 +69,13 @@ def get_best_pdf(request, judgment_uri):
...
 def detail(request, judgment_uri):
     judgment = get_published_judgment_by_uri(judgment_uri)
+
+    # If the judgment_uri which was requested isn't the canonical URI of the judgment, redirect the user
+    if judgment_uri != judgment.uri:
+        return HttpResponseRedirect(
+            reverse("detail", kwargs={"judgment_uri": judgment.uri})
+        )
+
     context = {}
```
anthonyhashemi, line 79: Looks good. Would we be able to add a test that shows when a user goes to, for example caselaw.....//google.com, they aren't redirected, to verify the bug is fixed?

PR #770, opened 2023-06-19
Feature/enhancements to exact title matches

timcowlishaw:
## Changes in this PR:

Some improvements to the 'exact title / ncn match' logic ahead of the marklogic PR i'm about to send which will push exact matches to the top of the ranking :-)

All of this fine to deploy as-is regardless of marklogic changes.

- **Don't stopword single characters** - This makes sense for plain english "I'm", "You'd", "He's" etc, if you're tokenizing on punctuation first, but (a) we're not, and (b) it causes havoc with anonymised judgments where the party name is a single character.
- **Make 'exact match' logic more robust** - This ensures that the exact match logic matches marklogic's search logic as close as possible, and adds a couple of enhancements - parentetitical statements such as (on the application of xxxx) are ignored for the purposes of exact matching, trailing space is ignored, and more forms of writing  'vs' are accepted.


judgments/fixtures/stop_words.py
```diff
@@ -135,13 +133,9 @@
     "should",
     "should've",
     "now",
-    "d",
     "ll",
```
jacksonj04, line 136: Is it worth reviewing this whole list for things like this, which only exist as part of contractions?

jacksonj04 (approved):

PR #779, opened 2023-06-22
update label example text to include just year

Donna-H:
## Changes in this PR:
Update legend to label consistent and styling with the others 'Party Name , Judge...' on the page also to highlight it is possible to search by year only by adding to help text “For example 01 01 2003” to For example 01 01 2003 or 2003”
### Before
before-add

### After
after-add


Terry-Price (commented):
looking good Donna, let's put it live and we can alway tell  Tim what you decided later about the legends. But I think it's now consistent and correct.

Terry-Price (approved):
looking good Donna.

PR #787, opened 2023-06-28
Use balanced line lengths in judgment titles

jacksonj04:
## Changes in this PR:

Use `text-wrap: balance` in judgment headers [where supported](https://github.com/nationalarchives/ds-caselaw-public-ui/pull/new/feature/balance-judgment-titles) to produce visually balanced headings.

In browsers where `balance` isn't supported, behaviour is the same as previously.

### Before

localhost_3000_ewca_civ_2023_701 (1)

### After

localhost_3000_ewca_civ_2023_701

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_text_toolbar.scss
@@ -42,7 +42,18 @@
   }
 
   &__title {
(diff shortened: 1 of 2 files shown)


Terry-Price (approved):
Fantastic @jacksonj04 – I know it;s only supported since this Spring, but will gain usage over time. Well done Sir.

PR #789, opened 2023-06-28
Refer to press summary when viewing press summary

anthonyhashemi:
## Changes in this PR:
Since we were adding the buttons to go back and forth between judgment and press summary, I thought it was only right that we fix up the press summary html pages to show the correct static text.

Adds conditional logic to the detail view we now share between judgment and press summary to display "press summary" in text when its a press summary!

### Before
Screenshot 2023-06-28 at 17 02 21

### After
Screenshot 2023-06-28 at 17 05 08

### DIFF
--- a/judgments/views/detail.py
+++ b/judgments/views/detail.py
@@ -78,12 +78,16 @@ def detail(request, judgment_uri):
 
     context = {}
 
+    if "press-summary" in judgment_uri:
+        context["document_type"] = "press_summary"
+    else:
+        context["document_type"] = "judgment"
+
     context["judgment"] = judgment.content_as_html("")  # "" is most recent version
     context["page_title"] = judgment.name
     context["judgment_uri"] = judgment.uri
     context["judgment_title"] = judgment.name
     context["judgment_ncn"] = judgment.neutral_citation
-
     context["pdf_size"] = get_pdf_size(judgment.uri)
     context["pdf_uri"] = (
         get_pdf_uri(judgment.uri)
--- a/locale/en_GB/LC_MESSAGES/django.po
+++ b/locale/en_GB/LC_MESSAGES/django.po
@@ -138,9 +138,19 @@ msgid "judgment.downloadoptions.pdf.cta"
 msgstr "Download this judgment as a PDF"
 
 #: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
+msgid "press_summary.downloadoptions.pdf.cta"
+msgstr "Download this press summary as a PDF"
+
+#: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
 msgid "judgment.downloadoptions.pdf.description"
 msgstr ""
 "The original format of the judgment as handed down by the court, for "
+"printing and downloading."
+
+#: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
+msgid "press_summary.downloadoptions.pdf.description"
+msgstr ""
+"The original format of the press summary as handed down by the court, for "
 "printing and downloading."
 
 #: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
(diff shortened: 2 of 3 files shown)


locale/en_GB/LC_MESSAGES/django.po
```diff
@@ -137,22 +137,42 @@ msgstr "Download options"
...
+
 #: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
 msgid "judgment.downloadoptions.pdf.description"
 msgstr ""
 "The original format of the judgment as handed down by the court, for "
 "printing and downloading."
 
+#: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
+msgid "press_summary.downloadoptions.pdf.description"
+msgstr ""
```
anthonyhashemi, line 152: I did look into whether there was a way I could not duplicate all these translations just for the sake of changing "judgment" to "press_summary" but I couldnt yet get it working.

I might try a bit more to do that as it would be much nicer without all this duplication.

but if anyone has any suggestions, please shout 🙏

dragon-dxw: https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#blocktranslate-template-tag 

we might be able to say 
```
{% blocktranslate with doctype=judgment.type %}
{{doctype}}.downloadoptions.pdf.description
{% endblocktranslate %}
```

(Also, the django documentation seems to suggest that we should just be using English strings, rather than our arbitary codes, with hints given via comments https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#comments-for-translators-in-templates)

judgments/views/detail.py
```diff
@@ -68,6 +68,13 @@ def get_best_pdf(request, judgment_uri):
 
 
 def detail(request, judgment_uri):
+    context = {}
+
+    if judgment_uri.endswith("press-summary/1"):
```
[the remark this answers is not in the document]
Terry-Price: probably for another card. but just reading the yellow boxes. 
We could definitely think about reducing the box link text. 
perhaps. ‘ Download PDF of this judgment’ might work better as the key word is PDF and it’s a bit earlier in the read. 
Also in the body copy it says ‘printing and downloading’ I think the order should swooped because you would have to download before printing physically.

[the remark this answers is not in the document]
anthonyhashemi: > This should probably be `.contains("/press-summary/")`, since we're open to the possibility of multiple press summaries (or other documents of a type). There might be a need for a "get doc type from URL" util function.

@dragon-dxw  Yeah this was something I wanted to ask about.... I was about it unsure as to whether we had decided we'd keep the /1. I thought Rose had said we are certain we would only have one per judgment. Will bring up as a discussion in slack.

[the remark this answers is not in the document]
anthonyhashemi: > probably for another card. but just reading the yellow boxes. We could definitely think about reducing the box link text. perhaps. ‘ Download PDF of this judgment’ might work better as the key word is PDF and it’s a bit earlier in the read. And might fit on one line too. same for XML. Also in the body copy it says ‘printing and downloading’ I think the order should be swopped because you would have to download before printing physically.

@Terry-Price  Yeah definitely worth thinking about and making another ticket if the copy is agreed on 👍

locale/en_GB/LC_MESSAGES/django.po
dragon-dxw: boo

dragon-dxw:
> Need to add some tests still

Is this still the case?

anthonyhashemi:
> It might be worth leaving this until after we have a chat about our translation approach, but I don't see a problem with this, other than maybe that you wanted some tests (which might be a real pain.)

I will try the tests but will get this in for now. If we simplify things with translations then we can just simplify this alongside that

locale/en_GB/LC_MESSAGES/django.po
anthonyhashemi, line 152: this looks wrong - fix

anthonyhashemi:
Nw version of this PR at https://github.com/nationalarchives/ds-caselaw-public-ui/pull/795 because github was being weird with this one and not updating my with my changes

anthonyhashemi:
reopening

PR #790, opened 2023-06-28
Add press summary button in judgment view and vice versa

anthonyhashemi:
## Changes in this PR:
Added "View Press Summary" and "View Judgment" Buttons as designed here:
https://14z0gp.axshare.com/?id=qizmwu&p=judgment&g=1

Also generalised some css class names to make sense now that they contain / are utilised for more than before.

### Before
before-add-ps

### After
[add-press-summary-to-judgment-toolbar.webm](https://github.com/nationalarchives/ds-caselaw-public-ui/assets/75584408/3f937f8b-d427-44f8-8031-49e81cd55ec9)

### DIFF
--- a/judgments/tests/test_detail.py
+++ b/judgments/tests/test_detail.py
@@ -292,7 +292,10 @@ class TestViewRelatedDocumentButton(TestCase):
         mock_get_judgment_by_uri.side_effect = get_judgment_by_uri_side_effect
 
         expected_html_button = """
-        <a class="judgment-toolbar-download__option--ps btn-ps" role="button" draggable="false" href="/eat/2023/1">
+        <a class="judgment-toolbar-buttons__option--related-document btn-related-document"
+            role="button" draggable="false"
+            href="/eat/2023/1"
+        >
             View Judgment
             <span style="font-weight:normal;font-size:0.9rem"></span>
         </a>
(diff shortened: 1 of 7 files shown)


judgments/views/detail.py
```diff
@@ -80,8 +80,15 @@ def detail(request, judgment_uri):
 
     if judgment_uri.endswith("press-summary/1"):
         context["document_type"] = "press_summary"
+        context["linked_document_uri"] = judgment_uri.rstrip("press-summary/1")
     else:
         context["document_type"] = "judgment"
+        context["linked_document_uri"] = judgment_uri + "/press-summary/1"
```
anthonyhashemi, line 86: @dragon-dxw @jacksonj04 Eventually we will move this logic into api client, but either way, keeping the /1 with the possibility of /2 will mean we need to decide on what version we link to with the press summary button

judgments/views/detail.py
```diff
@@ -78,10 +78,18 @@ def detail(request, judgment_uri):
 
     context = {}
 
-    if "press-summary" in judgment_uri:
```
anthonyhashemi, line 81: I know we previously said that we should allow for urls other than /1, but after discussing, we realise thats a proper edge case so will just always look for /1 until we are told otherwise.

Having it more general would have complicated things here unncessarily

dragon-dxw: :+1:

judgments/views/detail.py
```diff
@@ -78,10 +78,18 @@ def detail(request, judgment_uri):
...
+    press_summary_suffix = "/press-summary/1"
+    if judgment_uri.endswith(press_summary_suffix):
```
dragon-dxw, line 82: I think this logic should probably be on the Judgment class.

dragon-dxw (approved):

judgments/tests/test_detail.py
```diff
@@ -226,3 +226,141 @@ def test_non_canonical_uri_redirects(self, mock_judgment):
         assert isinstance(response, HttpResponseRedirect)
         assert response.status_code == 302
         assert response.url == "/test/1234/567"
+
+
+class TestViewRelatedDocumentButton(TestCase):
```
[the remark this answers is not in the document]
dragon-dxw: :+1:

judgments/views/detail.py
anthonyhashemi: yep, 100% agree. Was planning on making a card to expose a relationship between the 2 in the api client. Needs a bit more thinking on the best way

PR #795, opened 2023-06-30
Refer to press summary when viewing press summary

anthonyhashemi:
Note: new version of https://github.com/nationalarchives/ds-caselaw-public-ui/pull/789, since it wouldnt update with my force pushed changes.

## Changes in this PR:
Since we were adding the buttons to go back and forth between judgment and press summary, I thought it was only right that we fix up the press summary html pages to show the correct static text.

Adds conditional logic to the detail view we now share between judgment and press summary to display "press summary" in text when its a press summary!

### Before
Screenshot 2023-06-28 at 17 02 21

### After
Screenshot 2023-06-28 at 17 05 08

### DIFF
--- a/judgments/tests/test_detail.py
+++ b/judgments/tests/test_detail.py
@@ -4,7 +4,7 @@ from unittest.mock import patch
 import pytest
 from caselawclient.errors import JudgmentNotFoundError
 from django.http import Http404, HttpResponseRedirect
-from django.test import TestCase
+from django.test import Client, TestCase
 from factories import JudgmentFactory
 
 from judgments.views.detail import (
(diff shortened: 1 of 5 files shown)


judgments/tests/test_detail.py
```diff
@@ -131,6 +131,53 @@ def test_pdf_link_with_no_size(self, mock_judgment, mock_pdf_size):
...
+        THEN html response should contain the download options div
+        AND this contains the Download PDF button
+        AND this contains the Download XML button
+        AND the descriptions refer to the document's type
+        """
+        mock_judgment.return_value = JudgmentFactory.build(uri=uri, is_published=True)
+        mock_get_pdf_size.return_value = "(112KB)"
+        mock_get_pdf_uri.return_value = "http://example.com/test.pdf"
+        client = Client()
+        response = client.get(f"/{uri}")
```
anthonyhashemi, line 158: Once routed into the view - the `judgement_uri` drops the starting slash. Is that an issue with a routing? It causes us to have to remove the / when creating our JudgmentFactory object

dragon-dxw: So I think the canonical forms of the URL fragment are generally `ewhc/2009/1` and `/ewhc/2009/1.xml` depending on whether you're in Python or MarkLogic space, so I think this is correct. JudgmentFactory should probably be permissive about accepting `ewhc/2009/1`, though.

dragon-dxw (approved):

judgments/tests/test_detail.py
anthonyhashemi: interesting, yeah Ill try to edit the factory in a followup pr

PR #798, opened 2023-07-03
update breadcrumbs press summary and add tests

Donna-H:
## Changes in this PR:
Updated breadcrumbs added press summary and completed tests
### Before
before

### After
after

### DIFF
--- a/ds_judgements_public_ui/templates/includes/header_nav.html
+++ b/ds_judgements_public_ui/templates/includes/header_nav.html
@@ -21,7 +21,10 @@
               {% if search_context.query %}for "{{ search_context.query }}"{% endif %}</a>
             </li>
           {% endif %}
-          {% if title and link %}<li>{{ title }}</li>{% endif %}
+          {% if title and link %}
+            <li>{{ title }}</li>
+            {% if context.document_type == "press_summary" %}<li>Press Summary</li>{% endif %}
+          {% endif %}
         </ol>
       </nav>
     </div>
--- a/ds_judgements_public_ui/templates/includes/header_nav.html
+++ b/ds_judgements_public_ui/templates/includes/header_nav.html
@@ -9,7 +9,7 @@
       <ol>
         <li>
           <span class="page-header__breadcrumb-you-are-in">You are in:</span>
-          {% if current == "home" %}
+          {% if title == "home" %}
             {% translate "common.findcaselaw" %}
           {% else %}
             <a href="{% url 'home' %}">{% translate "common.findcaselaw" %}</a>
@@ -17,8 +17,10 @@
         </li>
         {% if search_context %}
           <li>
-            <a href="{{ search_context.search_url }}">{% translate "breadcrumbs.search" %}
-              {% if search_context.query %}for "{{ search_context.query }}"{% endif %}</a>
+            <a href="{{ search_context.search_url }}">
+              {% translate "breadcrumbs.search" %}
+              {% if search_context.query %}for "{{ search_context.query }}"{% endif %}
+              </a>
             </li>
           {% endif %}
           {% if title and link %}
--- a/ds_judgements_public_ui/templates/includes/header_nav.html
+++ b/ds_judgements_public_ui/templates/includes/header_nav.html
@@ -23,7 +23,7 @@
               </a>
             </li>
           {% endif %}
-          {% if title and link %}
+          {% if link %}
             {% if context.document_type == "press_summary" %}
               <li>
                 {% if context.linked_document_uri %}
--- a/ds_judgements_public_ui/templates/includes/header_nav.html
+++ b/ds_judgements_public_ui/templates/includes/header_nav.html
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load i18n document_utils %}
 <div class="page-header__nav">
   <div class="page-header__site-logo">
     <a href="{% url "home" %}" id="home-link">{% translate "home" %}</a>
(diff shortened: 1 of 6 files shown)


Donna-H:
[PUI- Add to PS view - breadcrumbs](https://trello.com/c/saiNDw7H/1092-pui-add-to-ps-view-breadcrumbs)

anthonyhashemi (approved):

ds_judgements_public_ui/templates/includes/header_nav.html
```diff
@@ -9,19 +9,34 @@
       <ol>
         <li>
           <span class="page-header__breadcrumb-you-are-in">You are in:</span>
-          {% if current == "home" %}
+          {% if title == "home" %}
```
anthonyhashemi, line 12: current and title were taking in the same thing, so dropped current

ds_judgements_public_ui/templates/includes/header_nav.html
```diff
@@ -9,19 +9,34 @@
...
           {% else %}
             <a href="{% url 'home' %}">{% translate "common.findcaselaw" %}</a>
           {% endif %}
         </li>
         {% if search_context %}
           <li>
-            <a href="{{ search_context.search_url }}">{% translate "breadcrumbs.search" %}
-              {% if search_context.query %}for "{{ search_context.query }}"{% endif %}</a>
+            <a href="{{ search_context.search_url }}">
+              {% translate "breadcrumbs.search" %}
```
anthonyhashemi, line 21: formatting

ds_judgements_public_ui/templates/includes/header_nav.html
```diff
@@ -9,19 +9,34 @@
...
+              {% if search_context.query %}for "{{ search_context.query }}"{% endif %}
+              </a>
             </li>
           {% endif %}
-          {% if title and link %}<li>{{ title }}</li>{% endif %}
+          {% if link %}
```
anthonyhashemi, line 26: title being truthy was redundant here - we're never going to have a valid document with empty string as the FRBRName - or would we!?

Can add back in if we think we need. Will think of a test for it..

ds_judgements_public_ui/templates/includes/header_nav.html
```diff
@@ -9,19 +9,34 @@
...
+            {% if context.document_type == "press_summary" %}
+              <li>
+                {% if context.linked_document_uri %}
+                  <a href="{% url 'detail' context.linked_document_uri %}">{{ title|get_title_to_display_in_html:context.document_type }}</a>
```
anthonyhashemi, line 30: !!!!!!!!main change!!!!!!

ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
```diff
@@ -1,10 +1,10 @@
-{% load i18n %}
+{% load i18n document_utils %}
 <div class="judgment-toolbar">
   <div class="judgment-toolbar__container">
     {% if context.document_type == "press_summary" %}
       <p class="judgment-toolbar__press-summary-title">Press Summary</p>
     {% endif %}
-    <h1 class="judgment-toolbar__title">{{ context.judgment_title }}</h1>
+    <h1 class="judgment-toolbar__title">{{ context.page_title|get_title_to_display_in_html:context.document_type }}</h1>
```
anthonyhashemi, line 7: For PS with name Press Summary of Judgment A. heading of the html should be showing "Judgment A" as we have the press summary label above to show it is a PS. Not really part of the breadcrumbs PR, but since the logic to strip the prefix had to come somewhere, it made sense to fix this as we are here.

anthonyhashemi: I think I still need a test for it though

ds_judgements_public_ui/templates/layouts/base.html
```diff
@@ -36,7 +36,7 @@
     <header class="page-header">
       {% block header %}
         <div class="page-header__flex-container container py-1">
-          {% include "includes/header_nav.html" with current=context.page_title title=context.page_title link=request.path %}
+          {% include "includes/header_nav.html" with title=context.page_title link=request.path %}
```
anthonyhashemi, line 39: as mentioned above, current was redudnant

ds_judgements_public_ui/templates/layouts/judgment.html
```diff
@@ -30,7 +30,7 @@
     {% include "includes/phase_banner.html" %}
     <header class="page-header">
       <div class="page-header__flex-container container py-1">
-        {% include "includes/header_nav.html" with current=context.page_title title=context.page_title link=request.path %}
+        {% include "includes/header_nav.html" with title=context.page_title link=request.path %}
```
anthonyhashemi, line 33: again- redundant

judgments/templatetags/document_utils.py
```diff
@@ -0,0 +1,10 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def get_title_to_display_in_html(document_title, document_type):
```
anthonyhashemi, line 7: filter so we remove the "Press Summary of " prefix where we need to, but still keep the title of theml including it so we see it in the tab of a browser. Did consider passing through two context variables, one with stripped prefix but I thought this was cleaner.

judgments/tests/test_detail.py
```diff
@@ -280,7 +280,7 @@ class TestViewRelatedDocumentButton:
             ("eat/2023/1", "View Press Summary", "eat/2023/1/press-summary/1"),
         ],
     )
-    def test_view_judgment_button_when_press_summary_with_judgment(
+    def test_view_related_document_button_when_document_with_related_document(
```
anthonyhashemi, line 283: just realised these names werent updated before

anthonyhashemi:
temporarily closing whilst splitting into separate prs

anthonyhashemi:
Reopening

ds_judgements_public_ui/templates/includes/header_nav.html
anthonyhashemi: Manually set the FRBRName of a press summary on staging to "" (empty string) and it looks like this. Screenshot 2023-07-05 at 17 59 54

We should never get to this scenario though since we should not be publishing docs without names! And its better we see the PUI as obviously broken with these weird breadcrumb arrows with missing name, than trying to make it look nice in that case!

jacksonj04 (approved):

ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
anthonyhashemi: have moved to https://github.com/nationalarchives/ds-caselaw-public-ui/pull/805

PR #805, opened 2023-07-05
Sort out press summary title and headings

anthonyhashemi:
## Changes in this PR:
- Don't display press summaery prefix in heading oh html.
-  However continue to show "Press Summary of " prefix in the html title/ browser tabs.
Screenshot 2023-07-05 at 17 51 28

https://trello.com/c/WNIrNHhy/1126-dont-display-press-summary-of-prefix-in-heading-of-html

### Before
Screenshot 2023-07-05 at 17 50 47

### After
Screenshot 2023-07-05 at 17 50 14

### DIFF
--- a/judgments/views/detail.py
+++ b/judgments/views/detail.py
@@ -94,7 +94,6 @@ def detail(request, judgment_uri):
     # TODO: All references to `judgment` here need to be updated to the more general `document`
     context["judgment"] = judgment.content_as_html("")  # "" is most recent version
     context["judgment_uri"] = judgment.uri
-    context["judgment_title"] = judgment.name
     context["page_title"] = judgment.name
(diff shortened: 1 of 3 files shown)


jacksonj04 (approved):

judgments/views/detail.py
```diff
@@ -94,7 +94,6 @@ def detail(request, judgment_uri):
     # TODO: All references to `judgment` here need to be updated to the more general `document`
     context["judgment"] = judgment.content_as_html("")  # "" is most recent version
     context["judgment_uri"] = judgment.uri
-    context["judgment_title"] = judgment.name
```
anthonyhashemi, line 97: was redundant since page_title was the same as judgment_title and, judgment_title was only used once, so replaced that and removed this.

PR #812, opened 2023-07-07
Fix breadcrumbs on error pages

anthonyhashemi:
## Changes in this PR:
- Remove extra breadcrumb in error pages and fix the main shared breadcrumb in error pages

https://trello.com/c/rnT5T9iE/1105-page-not-found-double-breadcrumbs

### Before
Screenshot 2023-07-06 at 12 29 06

### After
Screenshot 2023-07-06 at 17 20 04

### DIFF
--- a/ds_judgements_public_ui/templates/layouts/base.html
+++ b/ds_judgements_public_ui/templates/layouts/base.html
@@ -36,7 +36,28 @@
     <header class="page-header">
       {% block header %}
         <div class="page-header__flex-container container py-1">
-          {% include "includes/header_nav.html" with title=context.page_title link=request.path %}
+          <div class="page-header__nav">
+            <div class="page-header__site-logo">
+              <a href="{% url "home" %}" id="home-link">{% translate "home" %}</a>
+            </div>
+            <div class="page-header__breadcrumb">
+              <nav class="page-header__breadcrumb-flex-container"
+                   aria-label="Breadcrumb">
+                <ol>
+                  <li>
+                    <span class="page-header__breadcrumb-you-are-in">You are in:</span>
+                    {% if title == "home" %}
+                      {% translate "common.findcaselaw" %}
+                    {% else %}
+                      <a href="{% url 'home' %}">{% translate "common.findcaselaw" %}</a>
+                    {% endif %}
+                  </li>
+                  {% block breadcrumbs %}
+                  {% endblock breadcrumbs %}
+                </ol>
+              </nav>
+            </div>
+          </div>
           {% include "includes/logo.html" %}
         </div>
       {% endblock header %}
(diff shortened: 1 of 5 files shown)


ds_judgements_public_ui/templates/layouts/base.html
```diff
@@ -36,7 +36,28 @@
...
+                <ol>
+                  <li>
+                    <span class="page-header__breadcrumb-you-are-in">You are in:</span>
+                    {% if title == "home" %}
+                      {% translate "common.findcaselaw" %}
+                    {% else %}
+                      <a href="{% url 'home' %}">{% translate "common.findcaselaw" %}</a>
+                    {% endif %}
+                  </li>
+                  {% block breadcrumbs %}
```
anthonyhashemi, line 55: sadly have had to copy this out from header_nav.html.

Have left it as it was in judgment.html for now but once judgment.html and base.html have been deduplicated we will just have the one copy of this code.

It was a shame that I couldnt figure out a way to mix inheritance blocks and includes. I tried various combinations of files but it doesnt work.

At minimum, I understand that you can only ever extend from 1 parent templatye, so including another file which extends from another file will not work.

Any suggestions much appreciated 🙏

judgments/tests/test_detail.py
```diff
@@ -424,6 +427,44 @@ def test_breadcrumb_when_judgment(self, mock_judgment, mock_get_pdf_size):
         </div>"""
         assert_contains_html(response, breadcrumb_html)
 
+    @patch("judgments.views.detail.get_pdf_size")
+    @patch("judgments.views.detail.get_judgment_by_uri")
+    @pytest.mark.parametrize(
+        "http_error,expected_breadcrumb",
+        [
+            (JudgmentNotFoundError, "Page not found"),
+            (Exception, "Server Error"),
```
anthonyhashemi, line 436: havent added test for 403 Forbidden even though we do have a template for that.

We dont have any pages that are authenticatd so we should get rid of that.

On the other hand we should be adding a 400 Bad request page for cases detailed in https://trello.com/c/k5JDtcKL/1129-dont-show-500-on-invalid-uris-show-400-bad-request-in-pui

ds_judgements_public_ui/templates/layouts/base.html
jacksonj04: Possibly this is more effort to fix than it is to just scrub `judgment.html` in favour of a better-behaving `base.html`.

jacksonj04: Possibly this is more effort to fix than it is to just scrub `judgment.html` in favour of a better-behaving `base.html`.

jacksonj04 (approved):

ds_judgements_public_ui/templates/layouts/base.html
anthonyhashemi: yeah definitely need to fix up the base/judgment html duplication but it would still be nice to be able to abstract some of the base html, but I will leave for now

PR #815, opened 2023-07-10
Generalise judgment variables with document

anthonyhashemi:
## Changes in this PR:

- Generalise judgment variables with document

No end user changes - just internal renaming. The view args changed from judgement_uri to document_uri but that is only exposed in our url routing - we don't have external apis, so that is fine.

https://trello.com/c/s6nPDKx6/1152-chore-generalise-pui-references-to-judgment-to-document

### DIFF
--- a/config/middleware.py
+++ b/config/middleware.py
@@ -73,6 +73,8 @@ class FeedbackLinkMiddleware:
             params["type"] = response.context_data["feedback_survey_type"]
 
         if "feedback_survey_document_uri" in response.context_data:
+            # TODO: update the survey to allow for generalisation to `document`
+            # https://trello.com/c/l0iBFM1e/1151-update-survey-to-account-for-judgment-the-fact-that-we-have-press-summaries-as-well-as-judgments-now
             params["judgment_uri"] = response.context_data[
                 "feedback_survey_document_uri"
             ]
--- a/config/middleware.py
+++ b/config/middleware.py
@@ -72,9 +72,9 @@ class FeedbackLinkMiddleware:
         if "feedback_survey_type" in response.context_data:
             params["type"] = response.context_data["feedback_survey_type"]
 
-        if "feedback_survey_judgment_uri" in response.context_data:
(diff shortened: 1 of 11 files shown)


config/middleware.py
```diff
@@ -72,9 +72,9 @@ def process_template_response(self, request, response):
         if "feedback_survey_type" in response.context_data:
             params["type"] = response.context_data["feedback_survey_type"]
 
-        if "feedback_survey_judgment_uri" in response.context_data:
+        if "feedback_survey_document_uri" in response.context_data:
             params["judgment_uri"] = response.context_data[
```
jacksonj04, line 76: Does this line also want tweaking?

ds_judgements_public_ui/templates/judgment/detail.html
```diff
@@ -1,8 +1,8 @@
 {% extends "layouts/judgment.html" %}
 {% block content %}
-  {% if context.judgment %}
+  {% if context.document %}
```
jacksonj04, line 3: Is this a good point to move this from `context['context']['judgment']` to just `context['document']`?

jacksonj04 (approved):
Looks solid, couple of comments but nothing I would consider blocking (especially given that we're gonna have another couple of rounds of renaming and abstracting anyway).

config/middleware.py
anthonyhashemi: I think not yet, I should have aded the comment 

```
# TODO: update the survey to allow for generalisation to `document`
            # https://trello.com/c/l0iBFM1e/1151-update-survey-to-account-for-judgment-the-fact-that-we-have-press-summaries-as-well-as-judgments-now
```

like I did for https://github.com/nationalarchives/ds-caselaw-public-ui/pull/815/files#diff-5c9236772e72d720d3783256e03d3441839b3958cf5f891688c0ca8ae964f4d6R115

anthonyhashemi: have added the comment to here too

PR #816, opened 2023-07-11
Utilise generalised document objects and methods from upgrade apiclie…

anthonyhashemi:
…nt version 12.0.0

### After


judgments/tests/factories.py
```diff
@@ -1,11 +1,10 @@
 from typing import Any
 from unittest.mock import Mock
 
-from caselawclient.models.judgments import Judgment
+from caselawclient.models.documents import Document
 
 
-class JudgmentFactory:
-    # "name_of_attribute": ("name of incoming param", "default value")
+class DocumentFactory:
```
anthonyhashemi, line 7: When replacing this I noticed that the factory had `judgment_date_as_string` which we still haven't decided on what the analogous for a press summary would be.

Similarly `has_supplementary_materials` might not be appropriate for a press summary. Potentially having a general document object with all the generic properties makes sense, but then some other objects which link them together such as.


Decision -> JudgmentDocument (Inherits Document) , Press Summary (Inherits Document)

A decision wouldnt have a uri but could be stored in django/postgres maybe? It could hold the ncn, whether it has supplementary materials, decision_date_as_string etc. ?

And eventually (once we sort case numbers and case pages etc. ): Case -> Decision 1, Decision 2, Decision 3....

Similar to what @timcowlishaw suggested the other day?

Im sure there are wholes in my particualr example but I think it would be good to sit down and have a proper think about the way we want to relate them a bit more.

jacksonj04: Yeah, there are definitely _lots_ of cases where we need to figure out if a behaviour is inherent to all documents, or if it's class-specific. It might be that `judgment_date` and `document_publication_date` are different things we care about... or it might not.

For now I think we just plough on and we can fix up if needed.

jacksonj04 (approved):

judgments/utils.py
```diff
@@ -203,13 +203,12 @@ def parse_date_parameter(params, param_name, default_to_last=False):
...
     api_client = MarklogicApiClient(
         host=settings.MARKLOGIC_HOST,
         username=settings.MARKLOGIC_USER,
         password=settings.MARKLOGIC_PASSWORD,
         use_https=settings.MARKLOGIC_USE_HTTPS,
     )
 
-    # raises a JudgmentNotFoundError if it doesn't exist
-    return Judgment(document_uri, api_client)
+    return Document(document_uri, api_client)
```
jacksonj04, line 214: I think this is still really a `Judgment`, since we don't have any way of interpreting different document types right now.

judgments/tests/factories.py
anthonyhashemi: Yeah agreed for here, but its worth us finalising the metadata ticket to have clear AC for what metadata we want to expose on a press summary and be able to edit in the EUI.

judgments/utils.py
anthonyhashemi: Not quite sure what you mean, will reach out on slack

PR #837, opened 2023-07-13
Chore/rename judgment specific templates to document/1152

anthonyhashemi:
## Changes in this PR:

Rename `judgment` specific html templates, css classes and translation msgids to refer to `document`

Update: also made it so static text about downloading the docs refer to document rather than needing to dynamically determine whether to display the same text but with `judgment` or `press_summary`

https://trello.com/c/s6nPDKx6/1152-chore-generalise-pui-references-to-judgment-to-document


ds_judgements_public_ui/templates/pages/about_this_service.html
```diff
@@ -161,7 +161,7 @@ <h2 id="section-data">Judgments and decisions as data</h2>
   <p>
     Judgments and decisions are sent to us as Word documents using the National Archive’s Transfer Digital Records
     service.
-    This is the digital record of the publication version of the judgment. We then create a digital surrogates of the
+    This is the digital record of the publication version of the document. We then create a digital surrogates of the
```
anthonyhashemi, line 164: I renamed a few references of judgment to document in the static pages. There probably is more changes to these static pages that we will like to make detailing more specifics about press summaries but I think these generalisations are fine for now. Happy to revert if people want.

ds_judgements_public_ui/templates/includes/document_text_download_options.html
```diff
@@ -0,0 +1,21 @@
+{% load i18n %}
+<div id="download-options" class="document-download-options">
+  <h2 class="document-download-options__header">{% translate "document.downloadoptions.title" %}</h2>
+  <div class="document-download-options__options-list">
+    <div class="document-download-options__download-option">
+      <h3>
+        <a href="{{ context.pdf_uri }}">
+          {% translate "document.downloadoptions.pdf.cta" %}
```
anthonyhashemi, line 8: no longer using the conditional 

`{% if context.document_type == 'press_summary' %}` to determine whether to show press_summary or judgment, but instead always show "document" in the download static text. Unless others feel strongly against that. It keeps things simpler this way.

locale/en_GB/LC_MESSAGES/django.po
```diff
@@ -129,48 +129,28 @@ msgstr "Help us improve this service"
...
-msgstr "Download this press summary as a PDF"
-
-#: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
-msgid "judgment.downloadoptions.pdf.description"
-msgstr ""
-"The original format of the judgment as handed down by the court, for "
-"printing and downloading."
-
-#: ds_judgements_public_ui/templates/includes/judgment_text_download_options.html
-msgid "press_summary.downloadoptions.pdf.description"
```
anthonyhashemi, line 151: here you can see the removal of 2 trnaslations (1 for each of judgment and press summary) to 1 that uses the general 'document' term.

ds_judgements_public_ui/sass/includes/_document_text.scss
```diff
@@ -1,4 +1,4 @@
-.judgment {
```
timcowlishaw, line 1: Just wanted to check - do we need to make a corresponding change in EUI to make sure that this CSS gets pulled across correctly on build?

timcowlishaw: (see the note in the readme)

timcowlishaw (changes requested):

timcowlishaw:
All looks good to me, thank you - just one little gotcha to check, just in case you weren't aware :-)

ds_judgements_public_ui/sass/includes/_document_text.scss
anthonyhashemi: https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1073

ds_judgements_public_ui/templates/includes/document_text_download_options.html
jacksonj04: Happy with this. We also have `{{ document.document_noun }}` available, but then we need to muck about with string inclusion in translations.

jacksonj04 (approved):

PR #852, opened 2023-08-02
Upgrade to customapi 13

dragon-dxw:


dragon-dxw:
Don't merge until there's a patch for the 
`Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration.` bug on custom api client.

judgments/views/detail.py
```diff
@@ -123,7 +123,7 @@ def detail(request, document_uri):
 def detail_xml(_request, document_uri):
     document = get_published_document_by_uri(document_uri)
 
-    document_xml = document.content_as_xml()
+    document_xml = document.content_as_xml
```
anthonyhashemi, line 126: Are you going to update the `detail` view as part of this PR on a followup?

In particular, we can replace the conditional
```
    press_summary_suffix = "/press-summary/1"
    if document_uri.endswith(press_summary_suffix):
```
With the usage of `get_document_from_uri` and then check the type of the returned instance.

and we can now use 

`context["judgment_ncn"] = document.neutral_citation` even for press summaries.

dragon-dxw: Yeah, I wanted to separate those things out.

dragon-dxw:
merged into https://github.com/nationalarchives/ds-caselaw-public-ui/pull/855

PR #855, opened 2023-08-07
Use document noun

dragon-dxw:
Update the public UI to use document noun to identify press summaries etc.

### DIFF
--- a/judgments/views/detail.py
+++ b/judgments/views/detail.py
@@ -94,16 +94,11 @@ def detail(request, document_uri):
         linked_doc_url = document_uri + press_summary_suffix
 
     try:
-        linked_document = get_published_document_by_uri(linked_doc_url)
-        context["linked_document_uri"] = linked_document.uri
+        context["linked_document_uri"] = get_published_document_by_uri(
+            linked_doc_url
+        ).uri
     except Http404:
-        linked_document = None
         context["linked_document_uri"] = ""
-
-    if context["document_noun"] == "press summary" and linked_document:
-        if linked_document.best_human_identifier is None:
-            raise NoNeutralCitationError(linked_document.uri)
-        context["judgment_ncn"] = linked_document.best_human_identifier
 
     context["document"] = document.content_as_html("")  # "" is most recent version
     context["document_uri"] = document.uri


dragon-dxw:
Whilst the tests pass, smoketests have a failure due to custom-api-client quirks. Fixing...

judgments/tests/test_detail.py
```diff
@@ -492,8 +512,8 @@ def test_document_headings_when_press_summary(
         self, mock_get_document_by_uri, mock_get_pdf_size
     ):
         """
-        GIVEN a press summary and related judgment
-        WHEN a request is made with the press summary URI
+        WHEN a request is made with to a document detail page
```
anthonyhashemi, line 515: could these 2 lines be swapped to be consistent with the Given-When-Then syntax?

judgments/utils.py
```diff
@@ -212,4 +212,4 @@ def get_document_by_uri(document_uri: str) -> Judgment:
     )
 
     # raises a JudgmentNotFoundError if it doesn't exist
```
anthonyhashemi, line 214: It now raises `DocumentNotFoundError`.

```suggestion
    # raises a DocumentNotFoundError if it doesn't exist
```

judgments/views/detail.py
```diff
@@ -79,23 +84,26 @@ def detail(request, document_uri):
     context = {}
 
     press_summary_suffix = "/press-summary/1"
-    if document_uri.endswith(press_summary_suffix):
-        context["document_type"] = "press_summary"
+    context["document_noun"] = document.document_noun
+    if document.best_human_identifier is None:
+        raise NoNeutralCitationError(document.uri)
```
anthonyhashemi, line 89: I think we should never reach the scenario of NoNeutralCitationError in a published document but this will make it easier for us to debug any edge cases that slip through for whatever reason.

Could we add tests for when we raise this then:

i.e for line 89 and 106.



even though we should not reach this scenario can we add a test for this scenario

judgments/views/detail.py
```diff
@@ -79,23 +84,26 @@ def detail(request, document_uri):
...
         context["linked_document_uri"] = ""
-
-    if context["document_type"] == "press_summary" and linked_document:
-        context["judgment_ncn"] = linked_document.neutral_citation
+    else:
+        if context["document_noun"] == "press summary" and linked_document:
+            if linked_document.best_human_identifier:
+                context["judgment_ncn"] = linked_document.best_human_identifier
+            else:
+                raise NoNeutralCitationError(linked_document.uri)
```
anthonyhashemi, line 106: as mentioned in other comment, could we test this

judgments/utils.py
```diff
@@ -212,4 +212,4 @@ def get_document_by_uri(document_uri: str) -> Judgment:
...
-    return Judgment(document_uri, api_client)
+    return api_client.get_document_by_uri(document_uri)
```
anthonyhashemi, line 215: looks like we don't currently test
`get_document_by_uri` and we mock it out in any test of functions that call it.

Can we add the following tests for it:


```
import unittest
from judgments.utils import get_document_by_uri

class TestGetDocumentByUriOrRaise(unittest.TestCase):
    @mock.patch('judgments.utils.MarklogicApiClient')
    def test_get_existing_document(self, mock_api_client):

        document = Mock()
        mock_api_client.return_value.get_document_by_uri.return_value = document

        result = get_document_by_uri_or_raise('sample_uri')

        self.assertEqual(result, document)

    @mock.patch('judgments.utils.MarklogicApiClient')
    def test_get_nonexistent_document(self, mock_api_client):
        mock_api_client.return_value.get_document_by_uri.side_effect = DocumentNotFoundError

        with self.assertRaises(DocumentNotFoundError):
            get_document_by_uri_or_raise('nonexistent_uri')
```

judgments/views/detail.py
```diff
@@ -79,23 +84,26 @@ def detail(request, document_uri):
...
 
-    linked_document = None
     try:
         linked_document = get_published_document_by_uri(context["linked_document_uri"])
     except Http404:
         context["linked_document_uri"] = ""
-
-    if context["document_type"] == "press_summary" and linked_document:
-        context["judgment_ncn"] = linked_document.neutral_citation
+    else:
```
anthonyhashemi, line 101: I think we can skip the else to reduce the levels of nesting

the line `if context["document_noun"] == "press summary" and linked_document:` 

checks for the linked_document being non Falsey which can only be the case if the try block succeeded.

judgments/views/detail.py
```diff
@@ -79,23 +84,26 @@ def detail(request, document_uri):
...
+        if context["document_noun"] == "press summary" and linked_document:
+            if linked_document.best_human_identifier:
```
anthonyhashemi, line 103: On lone 88 we check for `if document.best_human_identifier is None:`

can we make these 2 checks consistent?

anthonyhashemi: I probably wouldn't go as far as verifying the ncn is valid, but that does make me think again that we have validation in EUI to block publishing if ncn is invalid in any way, so maybe we could assume that best_human_identifier is valid for judgments and press summaries in the PUI and so NoNeutralCitationError is unnecessary?

jacksonj04 (commented):
Drive-by comment so I know I've said it somewhere, rather than it should be addressed in this PR: all the places we're explicitly checking if the noun is "press summary" should probably be regarded as smelly in the mid-term, since this isn't extensible. The fix will probably vary case-by-case though, maybe type-specific templates or pushing logic to a different layer.

judgments/tests/test_detail.py
```diff
@@ -601,3 +633,17 @@ def test_html_title_when_judgment(
         response = self.client.get("/eat/2023/1")
         html_title = "<title>Judgment A - Find case law</title>"
         assert_contains_html(response, html_title)
+
+
+class TestNoNeutralCitation(TestCase):
```
anthonyhashemi, line 638: I was about to say that this just covers line 89 in views/detail.py and line 106 is still untested.

However, we can actually just remove that whole block.

i.e remove  lines 103-106

```
    if context["document_noun"] == "press summary" and linked_document:
        if linked_document.best_human_identifier is None:
            raise NoNeutralCitationError(linked_document.uri)
        context["judgment_ncn"] = linked_document.best_human_identifier
```

since line 89 since we now use best_human_identifier

anthonyhashemi (changes requested):
All looks good now apart from we can remove the now redundant lines 103 to 106 in `detail.py`

judgments/views/detail.py
```diff
@@ -79,24 +84,22 @@ def detail(request, document_uri):
...
-        context["judgment_ncn"] = document.neutral_citation
+        linked_doc_url = document_uri + press_summary_suffix
 
-    linked_document = None
     try:
-        linked_document = get_published_document_by_uri(context["linked_document_uri"])
+        linked_document = get_published_document_by_uri(linked_doc_url)
+        context["linked_document_uri"] = linked_document.uri
     except Http404:
+        linked_document = None
```
anthonyhashemi, line 100: so now we don't need `linked_document` as a variable in the except case as it is now onlyu usef used for `linked_document_uri` so we can remove this line 100 `linked_document = None`

anthonyhashemi (approved):

PR #866, opened 2023-08-14
Use bookworm to fix version woes

dragon-dxw:
Comparable to https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/1107 -- upgrade Debian to upgrade Node to support upgrades of packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG PYTHON_VERSION=3.11-slim-bullseye
+ARG PYTHON_VERSION=3.11-slim-bookworm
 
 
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -41,7 +41,7 @@ ENV BUILD_ENV ${BUILD_ENVIRONMENT}
 WORKDIR ${APP_HOME}
 
 RUN addgroup --system django \
-    && adduser --system --ingroup django django
+    && adduser --system --ingroup django --home /home/django django
```
dragon-dxw, line 44: It works without this, but I've put this in as well to try to preserve editor/public parity.

I don't understand why this doesn't fail the same way any more.

anthonyhashemi (approved):

PR #872, opened 2023-08-24
Fix the use of the shared module in PUI

anthonyhashemi:
## Changes in this PR:

- Fix the usage of the shared scss.
In particular we had forgotten to update the scss import in `ds_judgements_public_ui/sass/judgmentpdf.scss`

I have also just dropped `.scss` from the import as that is the convention we use in this repo.

I hadn't verified this properly before as I just ran `npm run start-sass` but not `npm run build`

`npm run build` did give a complaint about `@import "includes/judgment_text";` since it was still referenced in `ds_judgements_public_ui/sass/judgmentpdf.scss`

https://trello.com/c/WkhvV1VO/1223-eui-pui-set-up-shared-front-end-code-repo-and-add-documenttext-css-to-it

### DIFF
--- a/ds_judgements_public_ui/sass/judgmentpdf.scss
+++ b/ds_judgements_public_ui/sass/judgmentpdf.scss
@@ -1,6 +1,6 @@
 @import "includes/variables";
 @import "includes/mixins";
-@import "includes/judgment_text";
+@import "../../node_modules/@nationalarchives/ds-caselaw-frontend/src/main";
 
 html {
   font-size: 12px;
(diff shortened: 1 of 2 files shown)


jacksonj04:
I'm worried that this implies we can ship invalid SCSS and it makes it all the way through testing and linting! Maybe we need to add a build step to the checks?

jacksonj04 (approved):

dragon-dxw (approved):

anthonyhashemi:
> I'm worried that this implies we can ship invalid SCSS and it makes it all the way through testing and linting! Maybe we need to add a build step to the checks?

definitely. will make card to address this in next sprint

PR #876, opened 2023-08-31
(FCL-138) Upgrade to Django 4

dragon-dxw:
Just updating the dependency.


dragon-dxw:
Jumping to Django 5 instead.

PR #891, opened 2023-09-12
Judgment template now uses common wrapper

jacksonj04:
This gets rid of the custom wrapper and navigation templates for judgments, meaning we only have to maintain one set of common page chrome.


timcowlishaw (approved):
Aha ace, much simpler. Thanks @jacksonj04!

PR #896, opened 2023-09-19
Breadcrumbs now render from structured data

jacksonj04:
The breadcrumbs now use the `breadcrumbs` array in the context to build and render a list from structured data.

For templates with hardcoded breadcrumbs, these can be overridden with the `breadcrumbs` block as previously.

### DIFF
--- a/judgments/views/detail.py
+++ b/judgments/views/detail.py
@@ -110,11 +110,27 @@ def detail(request, document_uri):
         else reverse("detail_pdf", args=[document.uri])
     )
 
+    if document.document_noun == "press summary":
+        breadcrumbs = [
+            {
+                "url": "/" + context["linked_document_uri"],
+                "text": document.name.removeprefix("Press Summary of "),
+            },
+            {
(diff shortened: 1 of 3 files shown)


judgments/views/detail.py
```diff
@@ -110,11 +110,27 @@ def detail(request, document_uri):
         else reverse("detail_pdf", args=[document.uri])
     )
 
+    if document.document_noun == "press summary":
+        breadcrumbs = [
+            {
+                "url": "/" + context["linked_document_uri"],
```
anthonyhashemi, line 116: why do we need url in the case of press summaries?

jacksonj04: This is keeping behaviour the same - at the moment it's structured:

Find Case Law > The Case of Foo vs Bar > Press Summary

anthonyhashemi: I had forgotten the fact that the middle section needs the url for it to be a link

anthonyhashemi (approved):

PR #905, opened 2023-09-29
Search date parser now returns a real `date` instead of a yyyy-mm-dd string

Donna-H:
## Changes in this PR:
Bug fix : update date filters  from yyyy-mm-dd to dd-mm-yyyy
https://trello.com/c/eSE97UiL/1094-pui-date-filters-read-yyyy-mm-dd-should-read-dd-mm-yyyy
### Before
before

### After
after

### DIFF
--- a/judgments/utils.py
+++ b/judgments/utils.py
@@ -1,7 +1,7 @@
 import math
 import re
 from calendar import monthrange
-from datetime import datetime
+from datetime import date, datetime
 from typing import Optional, TypedDict
 from urllib.parse import parse_qs, urlparse
 
@@ -190,7 +190,7 @@ def parse_parameter_as_int(params, parameter_name, default=None):
         return default
 
 
-def parse_date_parameter(params, param_name, default_to_last=False):
+def parse_date_parameter(params, param_name, default_to_last=False) -> Optional[date]:
     year_param_name = f"{param_name}_year"
     month_param_name = f"{param_name}_month"
     day_param_name = f"{param_name}_day"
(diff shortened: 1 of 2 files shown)


Donna-H:
[[PUI] Date filters read yyyy-mm-dd; should read dd-mm-yyyy](https://trello.com/c/eSE97UiL/1094-pui-date-filters-read-yyyy-mm-dd-should-read-dd-mm-yyyy)

Donna-H:
[[PUI] Date filters read yyyy-mm-dd; should read dd-mm-yyyy](https://trello.com/c/eSE97UiL/1094-pui-date-filters-read-yyyy-mm-dd-should-read-dd-mm-yyyy)

judgments/utils.py
```diff
@@ -205,7 +205,7 @@ def parse_date_parameter(params, param_name, default_to_last=False):
         day = parse_parameter_as_int(params, day_param_name, default=default_day)
 
         dt = datetime(year, month, day)
-        return dt.strftime("%Y-%m-%d")
+        return dt.strftime("%d-%m-%Y")
```
anthonyhashemi, line 208: worth adding a test for this bug fix

jacksonj04: Looks like we already have a failing test which just needs updating! @Donna-H let me know if you need a hand.

Donna-H: @jacksonj04 updated thanks for help please review the changes

jacksonj04 (approved):
Behaviour is now to format the date as "dd mon yyyy" (the default short date format), because we're passing a structured date object into the template and it's doing `.__str__` behind the scenes.

I'm happy with this, @HLLindon is happy with this. We can change the actual template behaviour later if we _really_ want dd-mm-yyyy.

Screenshot 2023-10-03 at 10 15 10

PR #908, opened 2023-10-03
Validate that inputted search years are within court date range

timcowlishaw:
## Changes in this PR:

If users search for judgments outside the range of our collection, warn them early. The copy may need a tweak

`./manage.py recalculate_court_dates --write` needs to have been run in the past for this to work (it should be run by cron on staging and prod, but may lead to some head-scratching locally if no errors are raised)

https://trello.com/c/wZwFBPAH/1086-pui-dont-allow-years-in-search-outside-our-collection-date
## Screenshots of UI changes:
Screenshot 2023-10-03 at 18 34 51

### DIFF
--- a/judgments/utils.py
+++ b/judgments/utils.py
@@ -190,7 +190,9 @@ def parse_parameter_as_int(params, parameter_name, default=None):
         return default
 
 
-def parse_date_parameter(params, param_name, default_to_last=False) -> Optional[date]:
+def parse_date_parameter(
+    params, param_name, default_to_last=False, start_year=None, end_year=None
+) -> Optional[date]:
     year_param_name = f"{param_name}_year"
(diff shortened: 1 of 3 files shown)


judgments/utils.py
```diff
@@ -190,14 +190,23 @@ def parse_parameter_as_int(params, parameter_name, default=None):
         return default
 
 
-def parse_date_parameter(params, param_name, default_to_last=False) -> Optional[date]:
+def parse_date_parameter(
```
jacksonj04, line 193: I think we're perhaps starting to hang too much on this function. I'm happy with this for now, but we should consider this as a candidate for refactoring how we validate the fields (eg by splitting "is this year valid" into its own validator)

jacksonj04 (approved):

PR #909, opened 2023-10-04
re-order filters on search pages

Donna-H:
## Changes in this PR:
Re-order filters on search pages
https://trello.com/c/GIFCMk3B/1039-pui-re-order-filter-boxes-so-court-lists-are-lower
### Before
before-structured-search
before-filter-search

### After
Screenshot from 2023-10-05 13-43-49

after-filter-page

### DIFF
--- a/ds_judgements_public_ui/templates/includes/results_filters_inputs.html
+++ b/ds_judgements_public_ui/templates/includes/results_filters_inputs.html
@@ -82,7 +82,6 @@
     </fieldset>
   </div>
 </div>
-{% include "includes/structured_search_additional_inputs.html" %}
 <div class="structured-search__multi-fields-panel">
   <div class="structured-search__specific-field-container">
     <label for="party_name"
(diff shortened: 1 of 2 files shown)


Donna-H:
[PUI: Re-order filter boxes so court lists are lower](https://trello.com/c/GIFCMk3B/1039-pui-re-order-filter-boxes-so-court-lists-are-lower)

ds_judgements_public_ui/templates/includes/results_filters_inputs.html
```diff
@@ -82,6 +74,15 @@
...
 </div>
+<div class="structured-search__single-field-panel">
+  <div class="structured-search__limit-to-container">
+    <fieldset>
+      <legend class="structured-search__limit-to-label">From specific courts or tribunals</legend>
+      <div class="structured-search__court-options" id="court">{% include "includes/courts_options.html" %}</div>
+    </fieldset>
+  </div>
+</div>
+{% include "includes/structured_search_additional_inputs.html" %}
```
timcowlishaw, line 85: Just wanted to check - did this line need moving? This adds the 'neutral citation' and 'keywords' fields on the structured search page, which should show up *below* the courts there, but not show up at all in the filters on the results page.

timcowlishaw (changes requested):
Hey donna - one quick query - I think something else might have been moved that didn't need to be. If you could just confirm that the *neutral citation* and *keywords* fields:

- Show up on the structured search page *below* the courts list
- *Don't show up at all* in the filters on results pages

... then i think we're good to go. Thank you!!

timcowlishaw (approved):

PR #911, opened 2023-10-05
First pass at heirarchical courts list

timcowlishaw:
## Changes in this PR:

Show the courts heirarchy on the structured search page and in the filters list.

This doesn't include 'select all' functionality, and needs some tidying up - both presentationally here, and in terms of names / structure of courts in the utils library.

In particular, we need to
 - Group the courts so that courts show in the left column, and tribunals in the right (work in both places)
 - Rename court divisions / tribunals so that they don't repeat the group heading (eg "Family division" vs "Family division of the High Court", "Lands Chamber" vs "Upper Tribunal Lands Chamber") (work in utils library - this may involve giving each court a 'grouped' and 'ungrouped' name, and using the correct one depending on context)

https://trello.com/c/Gj60XUIp/1380-investigate-court-hierarchy-work-finalize-design

### Before
Screenshot 2023-10-05 at 11 30 13

### After
Screenshot 2023-10-04 at 18 02 31

### DIFF
--- a/config/views.py
+++ b/config/views.py
@@ -70,7 +70,8 @@ class StructuredSearchView(TemplateViewWithContext):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["context"]["courts"] = courts.get_selectable_groups()
+        context["context"]["courts"] = courts.get_grouped_selectable_courts()
+        context["context"]["tribunals"] = courts.get_grouped_selectable_tribunals()
         context["feedback_survey_type"] = "structured_search"
         return context
 
--- a/config/views.py
+++ b/config/views.py
@@ -70,7 +70,7 @@ class StructuredSearchView(TemplateViewWithContext):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["context"]["courts"] = courts.get_selectable()
+        context["context"]["courts"] = courts.get_selectable_groups()
         context["feedback_survey_type"] = "structured_search"
         return context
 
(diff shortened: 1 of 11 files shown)


timcowlishaw:
[Investigate Court Hierarchy work - finalize design](https://trello.com/c/Gj60XUIp/1380-investigate-court-hierarchy-work-finalize-design)

jacksonj04 (commented):
At risk of doing design work, would this look better with more vertical padding between groups to make it clearer that they're a contained list?

timcowlishaw:
> At risk of doing design work, would this look better with more vertical padding between groups to make it clearer that they're a contained list?

aha, you are 100% correct, and thank you for pointing it out! - however, after some experimentation, I'm gonna do this after doing the work to put the courts in the left column and the tribunals in the right. The issue is that we need to use `margin-top` on the group headings (as we don't want extra space between top-level courts), and that currently has the side-effect of pushing the top item in the right column down in an unsightly way (as we're using CSS multi-column layout which unfortunately doesn't let you target the first element of a column for styling differently). Once I've changed to have an explicit 'courts column' and 'tribunals column' it'll be easy :-)

timcowlishaw:
Screenshot 2023-10-10 at 18 20 47
Screenshot 2023-10-10 at 18 12 49
Screenshot 2023-10-10 at 18 12 24
Screenshot 2023-10-10 at 18 11 01

Updated version after addressing all the 'snagging' issues on trello, tests will pass after an API release and version bump.

timcowlishaw:
Update: EAT is now at the top of the tribunals list! 🥳

PR #948, opened 2023-10-18
Highlight query results on judgment page

timcowlishaw:
## Changes in this PR:

This adds highlighting of query terms in the judgment body, and a UI for navigating between them, in order that users can better identify at a glance how relevant a judgment is to their query, and also navigate to relevant sections easily.

We also switch to using `waffle` for feature flags instead of `django-flags`, and put this behind a feature flag - **the flag `highlight_search_query` will need to be created and enabled on staging and production to make this visible**.

Tests will pass once [v17.2.0 of the API client is released](https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/448) and the SCSS will build once [this shared frontend change](https://github.com/nationalarchives/ds-caselaw-frontend/pull/17) is merged!

https://trello.com/c/tnpX6eJW/1412-search-highlight-term-within-a-judgment

#### Before
Screenshot 2023-11-09 at 08 14 05
#### After
Screenshot 2023-11-09 at 08 13 17

#### Before
Screenshot 2023-11-09 at 08 14 13

#### After
Screenshot 2023-11-09 at 08 13 34


timcowlishaw:
[SEARCH: Highlight term within a judgment](https://trello.com/c/tnpX6eJW/1412-search-highlight-term-within-a-judgment)

timcowlishaw:
Couple of good codeclimate suggestions there - will do a quick refactor tomorrow AM

timcowlishaw:
> Couple of good codeclimate suggestions there - will do a quick refactor tomorrow AM

Actually, i take that back - I think the explicitness of the current version would be lost if i refactored, and it's not a huge amount of repetition. I vote to leave as is, unless anyone objects?

pyproject.toml
```diff
@@ -3,3 +3,4 @@
 ignore="H021,H023,H030,H031"
 profile="django"
 indent=2
+custom_blocks="flag"
```
jacksonj04, line 6: \o/

judgments/migrations/0002_alter_courtdates_options.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.2.22 on 2023-11-09 05:34
```
jacksonj04, line 1: This is a bit concerning; where does this change come from?

jacksonj04 (changes requested):

judgments/views/detail.py
```diff
@@ -75,16 +80,17 @@ def get_best_pdf(request, document_uri):
 
 
 def detail(request, document_uri):
+    query = request.GET.get("query")
     document = get_published_document_by_uri(document_uri)
 
     # If the document_uri which was requested isn't the canonical URI of the document, redirect the user
     if document_uri != document.uri:
```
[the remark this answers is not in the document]
timcowlishaw: aha good shout - death to unneeded complexity! I'll remove.

judgments/migrations/0002_alter_courtdates_options.py
timcowlishaw: ok, i think there was an earlier change somewhere on `main` which would have generated a migration, but for whatever reason we never ran the migration generation command at the time, so therefore when we added waffle and ran the migrations for that, it generated and applied the old migration too. Maybe this should come out of this PR and I'll open a new one for it on its own?

ds_judgements_public_ui/static/js/src/modules/document_navigation_links.js
```diff
@@ -45,18 +105,109 @@ $(() => {
...
+                root: null,
+                rootMargin: `-25% 0px -25% 0px`,
+            },
+        );
+
+        $("#start-of-document-link").text("Start");
+        $("#end-of-document-link").text("End");
+    }
+
+    $("#end-of-document-link").on("click", (event) => {
```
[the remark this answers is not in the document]
timcowlishaw: oh yep, that's a lovely solution - nothing had occured to me that didn't make the JS a little too obscure while also removing the dupication. I'll do that this AM!

config/settings/base.py
```diff
@@ -72,15 +69,18 @@
...
+    "waffle",
 ]
 
 LOCAL_APPS = [
     # Your stuff: custom apps go here
     "judgments.apps.JudgmentsConfig",
 ]
 # https://docs.djangoproject.com/en/dev/ref/settings/#installed-apps
-INSTALLED_APPS = DJANGO_APPS + LOCAL_APPS
+INSTALLED_APPS = DJANGO_APPS + THIRD_PARTY_APPS + LOCAL_APPS
```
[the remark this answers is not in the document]
timcowlishaw: 🙌

judgments/migrations/0002_alter_courtdates_options.py
timcowlishaw: As discussed, i'll open another PR with that change!

jacksonj04 (approved):
This gets the green light to merge and deploy, _but_ we shouldn't set up the feature flag on PUI in any kind of deployment mode (ie only opt in specific users via admin) until we've had the changes made to the cookie banner.

Issue #985, opened 2023-11-15
Homepage blocked by non-Google and Microcost bots, and other `robots.txt` oddness


symroe:
The current `robots.txt` file allows `Googlebot` and `bingbot` to crawl the home page, but disallows other bots from doing the same thing:

(code block snipped to highlight lines)
```
User-Agent: Googlebot
User-Agent: bingbot
Allow: /
...
Disallow: /$

User-Agent: *
Disallow: /
```

This is somewhat confusing, especially the seemingly contradictory `Disallow: /$` line for `Googlebot` and `bingbot`.

I think the intent is to allow all robots to access the home page (`/`) and the list of allowed pages. All other pages should be disallowed. If that's true, I think it can be simplified to:

```
User-Agent: *
Allow: /
Allow: /transactional-licence-form
Allow: /about-this-service
Allow: /how-to-use-this-service
Allow: /accessibility-statement
Allow: /open-justice-licence
Allow: /terms-of-use
Disallow: /*/
```

As an aside, when looking at this I found two other URLs that are indexed:

* `api.caselaw.nationalarchives.gov.uk` isn't excluded from crawling
* `assets.caselaw.nationalarchives.gov.uk` has a `robots.txt` file excluding `/` but results are showing for it on a [Google `site:` search](https://www.google.com/search?hl=en&q=site%3Ahttps%3A%2F%2Fcaselaw.nationalarchives.gov.uk%2F#ip=1). I think this is meant to be `Disallow: *` for all bots?

dragon-dxw:
assets robots.txt is now
```
User-Agent: *
Disallow: *
```

jacksonj04:
> `Disallow: *` is not the right syntax -- it should be `Disallow: /`. Put both in assets.

Was just about to say the same! Just to confirm:

- [ ] `api.caselaw` has `User-Agent: *` and `Disallow: *`
- [ ] `assets.caselaw` has `User-Agent: *` and `Disallow: *`

timcowlishaw:
Another one that seems to be being indexed by goodle:

https://caselaw.nationalarchives.gov.uk/ewca/civ/2023/1360

See [This search query](https://www.google.com/search?client=firefox-b-d&q=joseph+kyle+coventry+city+council)  - it's the top result.

timcowlishaw:
It might be worth checking the common crawl url index too at some point, (when it's working, it's giving a 500 this morning): http://urlsearch.commoncrawl.org/

[...]

PR #991, opened 2023-11-17
Fix some robots.txt issues

dragon-dxw:
Addresses issues with `robots.txt` raised in #985.

### DIFF
--- a/ds_judgements_public_ui/templates/robots.txt
+++ b/ds_judgements_public_ui/templates/robots.txt
@@ -6,4 +6,3 @@ Allow: /accessibility-statement
 Allow: /open-justice-licence
 Allow: /terms-of-use
 Disallow: /
-Disallow: *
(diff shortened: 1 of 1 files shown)


ds_judgements_public_ui/templates/robots.txt
```diff
@@ -1,19 +1,9 @@
...
 User-Agent: *
-Disallow: /
 Allow: /transactional-licence-form
 Allow: /about-this-service
 Allow: /how-to-use-this-service
 Allow: /accessibility-statement
 Allow: /open-justice-licence
 Allow: /terms-of-use
+Disallow: /
+Disallow: *
```
jacksonj04, line 9: I don't think this `Disallow: *` line is valid, we can ditch it.

jacksonj04 (approved):
One line to remove, otherwise looks good!

PR #1001, opened 2023-11-23
Support ukait court URLs

dragon-dxw:
Understand /ukait is a court URL


jacksonj04 (approved):

jacksonj04:
We want to find a better way to handle this as we expand coverage, but this is good for now.

PR #1008, opened 2023-11-29
PUI update cookie banner top

Donna-H:
## Changes in this PR:
The FCL cookie policy panel currently sits on top of interactive content therefore moved so that it complies with WCAG 2.2.
https://trello.com/c/l73pLZQ9/1514-cookie-policy-wcag-22-requirement
### Before
[before.webm](https://github.com/nationalarchives/ds-caselaw-public-ui/assets/75584408/c362e0e5-dd26-4748-a5f4-7f72268e1e48)

### After
[after.webm](https://github.com/nationalarchives/ds-caselaw-public-ui/assets/75584408/4aece466-79ac-4bce-bf97-db03b54f559c)


Donna-H:
[Cookie Policy WCAG 2.2 requirement](https://trello.com/c/l73pLZQ9/1514-cookie-policy-wcag-22-requirement)

jacksonj04:
This feels like something we should be pushing back upstream to TNA as a whole, since this is their banner and it's going to need fixing across the entire estate.

timcowlishaw:
Hey @Donna-H - on reflection, I think @jacksonj04 perhaps has a point here - we've maybe been a bit hasty in trying to fix this ourselves. However, one useful thing to do might be to open an Issue on https://github.com/nationalarchives/ds-cookie-consent for to get the ball rolling on discussing it? we could then propose a PR over there to fix it.

ahosgood:
I think this hotfix is fine for the time being. We recognise that the current cookie banner is not WCAG 2.2 compliant but we don't have to be compliant for a little while yet.

We are working on a new cookie banner (https://nationalarchives.github.io/design-system/components/cookie-banner/) that should eventually be easier to add to your services. This is still in development though so for now, we have made a similar fix for the DS cookie banner in beta.nationalarchives.gov.uk: https://github.com/nationalarchives/ds-wagtail/blob/develop/sass/etna.scss#L427-L436

dragon-dxw (approved):

PR #1017, opened 2023-12-04
Turn JSON Cookie errors into 400 errors

dragon-dxw:
Deliberately malformed cookies give a lot of false positives. More correctly flag these as being suspicious behaviour and then deprioritise in rollbar.

### DIFF
--- a/judgments/context_processors.py
+++ b/judgments/context_processors.py
@@ -1,6 +1,8 @@
 import json
 from urllib.parse import unquote
 
+from django.core.exceptions import SuspiciousOperation
+
 from config.settings.base import env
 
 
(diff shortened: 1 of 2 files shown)


judgments/context_processors.py
```diff
@@ -12,7 +14,11 @@ def cookie_consent(request):
 
     if cookie_policy:
         decoder = json.JSONDecoder()
-        decoded = decoder.decode(unquote(cookie_policy))
+        try:
+            decoded = decoder.decode(unquote(cookie_policy))
+        except json.JSONDecodeError:
+            raise SuspiciousOperation("Cookie tampered with: not valid JSON")
+            raise BadRequest("Cookie tampered with: not valid JSON")
```
jacksonj04, line 21: Does this work? Surely we'll never reach the `BadRequest` here because we've already raised `SuspiciousOperation`?

jacksonj04 (approved):

judgments/context_processors.py
dragon-dxw: No, the Suspicious Operation raises a Bad Request, that's gubbins.

PR #1042, opened 2023-12-19
ensure that the letter "v" is correctly stopworded

timcowlishaw:
## Changes in this PR:

This re-adds 'v' to the stopword list, and ensures that the normalisation of various common ways of writing 'v' happens before, so that they're all caught and stopped.

Note that we don't stopword single-word queries, so searching for 'v' explicitly will still return results (it took me a while to realise that this was happening!)


timcowlishaw:
[PUI: SEARCH Check if 'V' is acting as a stopword](https://trello.com/c/za5evnGr/1576-pui-search-check-if-v-is-acting-as-a-stopword)

dragon-dxw (approved):

dragon-dxw:
Should probably be in alphabetical order?

PR #1068, opened 2024-01-18
Catch MarklogicNotPermitted errors

dragon-dxw:
If a judgment was published and its press summary existed but was not published, the press summary would raise a MarklogicNotPermitted error. We didn't catch that.

The function returns a Http404 which feels like the wrong type for something returning a Document. But that's not worth trying to fix right now, I think.

### DIFF
--- a/judgments/views/detail.py
+++ b/judgments/views/detail.py
@@ -2,7 +2,7 @@ import logging
 import os
 
 import requests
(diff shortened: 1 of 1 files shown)


jacksonj04:
This doesn't feel right to me; there shouldn't be this kind of coupling going on between documents, so this to me suggests we've got something architecturally wrong elsewhere.

PR #1071, opened 2024-01-22
Bump utils to 1.3.5 to get EWFC (B) fix

dragon-dxw:


dragon-dxw:
[PUI: Family Court filter - displaying 'b' when it should not be](https://trello.com/c/FRTNXptE/1629-family-court-filter-displaying-b-when-it-should-not-be)

jacksonj04 (approved):

PR #1094, opened 2024-02-09
Optionally run a single test

pezholio:
This allows us to optionally specify a single test to run using `fab test` - if provided, we just run `pytest` with the test argument.


dragon-dxw (approved):
I have been looking for this off and on for months. Thank you.

PR #1097, opened 2024-02-13
(1704) Handle press summary redirects and multiple summaries

pezholio:
This adds a resolver to handle the multiple formats that judgements may or may not have and sends the details to the relevant view. It also handles calls to `$documentUri/press-summary`, fetching all possible press summaries for a document and:

- If there is 1 press summary, we redirect
- If there are no press summaries, we 404
- If there are multiple press summaries, we return a list

Note that we expect case 3 to be very much an edge case, so the design is a bit rudimentary!

## Screenshot

image

### DIFF
--- a/judgments/resolvers/document_resolver_engine.py
+++ b/judgments/resolvers/document_resolver_engine.py
@@ -1,0 +1,25 @@
+from typing import Optional
+
+from django.http.request import HttpRequest
+from django.views.generic import View
+
+from judgments.views.detail import detail, detail_xml, get_best_pdf, get_generated_pdf
+
+
+class DocumentResolverEngine(View):
+    def dispatch(
+        self,
+        request: HttpRequest,
+        document_uri: str,
+        file_format: Optional[str] = None,
+    ):
+        fileformat_lookup = {
+            "data.pdf": get_best_pdf,
+            "generated.pdf": get_generated_pdf,
+            "data.xml": detail_xml,
+            "data.html": detail,
+        }
+        if file_format:
+            return fileformat_lookup[file_format](request, document_uri)
+
+        return detail(request, document_uri)
--- a/judgments/resolvers/document_resolver_engine.py
+++ b/judgments/resolvers/document_resolver_engine.py
@@ -4,6 +4,7 @@ from django.http.request import HttpRequest
 from django.views.generic import View
 
 from judgments.views.detail import detail, detail_xml, get_best_pdf, get_generated_pdf
+from judgments.views.press_summaries import press_summaries
 
 
 class DocumentResolverEngine(View):
@@ -12,6 +13,7 @@ class DocumentResolverEngine(View):
         request: HttpRequest,
         document_uri: str,
         file_format: Optional[str] = None,
+        component: Optional[str] = None,
     ):
         fileformat_lookup = {
             "data.pdf": get_best_pdf,
@@ -19,7 +21,13 @@ class DocumentResolverEngine(View):
             "data.xml": detail_xml,
             "data.html": detail,
         }
+        component_lookup = {
+            "press-summary": press_summaries,
+        }
         if file_format:
             return fileformat_lookup[file_format](request, document_uri)
+
+        if component:
+            return component_lookup[component](request, document_uri)
 
         return detail(request, document_uri)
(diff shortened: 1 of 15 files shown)


judgments/urls.py
```diff
@@ -38,22 +38,10 @@
         name="feed",
     ),
     re_path(
-        r"^(?P<document_uri>.*/\d{4}/\d+.*)/data.pdf$",
-        get_best_pdf,
-        name="detail_pdf",
+        r"^(?P<document_uri>.*/\d{4}/\d+.*)/?$",
```
dragon-dxw, line 41: I'm struggling to see how this works at the moment

judgments/resolvers/document_resolver_engine.py
```diff
@@ -0,0 +1,20 @@
...
+class DocumentResolverEngine(View):
+    def dispatch(self, request: HttpRequest, document_uri: str):
+        response_lookup = {
+            "/data.pdf": get_best_pdf,
+            "/generated.pdf": get_generated_pdf,
+            "/data.xml": detail_xml,
+            "/data.html": detail,
+        }
+        for key in response_lookup:
+            if document_uri.endswith(key):
```
dragon-dxw, line 18: I feel like this significantly distorts the meaning of document_uri which is typically meaning "something of the form uksc/2024/1 or uksc/2024/1/press-summary/1, which has a direct correspondance to marklogic. I think I'd like to separate out the filename and path components..

judgments/utils.py
```diff
@@ -244,3 +244,15 @@ def formatted_document_uri(document_uri: str, format: Optional[str] = None) -> s
...
+
+def linked_doc_url(document: Document):
+    press_summary_suffix = "/press-summary/1"
+    if document.document_noun == "press summary":
+        return document.uri.removesuffix(press_summary_suffix)
+    else:
+        return document.uri + press_summary_suffix
+
+
+def linked_doc_title(document: Document):
```
dragon-dxw, line 264: can we say that document is in fact a Press Summary? I think this is more specific than a Document. I'll need to take a look at how this winds up getting used.

Should these be on the Document class; or PressSummary and Judgment classes?

dragon-dxw: Arguably the linked-doc of a Judgment should have this prefix added.

judgments/views/press_summaries.py
```diff
@@ -0,0 +1,41 @@
...
+            "judgement_name": judgement_name,
+            "breadcrumbs": press_summary_list_breadcrumbs(press_summaries[0]),
+            "press_summaries": press_summaries,
+        },
+    )
+
+
+def fetch_summaries(document_uri) -> list[PressSummary]:
+    press_summaries = get_press_summaries_for_document_uri(document_uri)
+    if len(press_summaries) == 0:
```
dragon-dxw, line 38: It feels odd to handle 0 here with a HTTP404 but 1 in the other function. I think we should fold these functions together.

judgments/resolvers/document_resolver_engine.py
pezholio: I think you're right here - I think it was a bit of naivety on my part coming to the project cold. I'll give it another go. Think I've got some ideas brewing 👍

judgments/urls.py
pezholio: I've made some tweaks to the urls and the resolver to make it easier to grok now

judgments/views/press_summaries.py
pezholio: Yeah, I had that originally, but CodeClimate didn't like it 😢

judgments/utils.py
pezholio: I tried to tie these helpers more closely to the classes, but got a bit tangled (and arguably, there's a lot more utils that should probably be on a class instance rather than a util). I've now tweaked this so both helpers support PressSummary and Judgment classes.

pezholio: I might do some work on this this afternoon in a seperate PR - I've got some ideas

dragon-dxw: I'll be happier when this is gone, but happy for now. :)

dragon-dxw (approved):

judgments/utils.py
pezholio: I’ve got something in the pipeline that should tidy a lot of this up

PR #1098, opened 2024-02-14
Serve schemas at /schema/akn-modified.xsd, caselaw.xsd and xml.xsd from Github.

dragon-dxw:
Questions to answer:

Is these the right URL?

> Also, the “schemaLocation” on legislation.gov.uk is http://www.legislation.gov.uk/schema/legislation.xsd, so maybe we could use https://caselaw.nationalarchives.gov.uk/schema/filename.xsd -- Jim

Is this the right approach?

Should /akn redirect here?

### DIFF
--- a/config/converters.py
+++ b/config/converters.py
@@ -1,0 +1,8 @@
+class SchemaFileConverter:
+    regex = r"(?:xml|caselaw|akn-modified)\.xsd"
+
+    def to_python(self, value):
+        return str(value)
+
+    def to_url(self, value):
+        return str(value)
--- a/config/tests.py
+++ b/config/tests.py
@@ -1,4 +1,10 @@
+from unittest.mock import patch
+
+import pytest
+from django.http import Http404
 from django.test import TestCase
+
+from config import views
 
 
 class TestCacheHeaders(TestCase):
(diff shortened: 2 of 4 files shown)


config/converters.py
```diff
@@ -0,0 +1,8 @@
+class SchemaFileConverter:
+    regex = r"(?:xml|caselaw|akn-modified)\.xsd"
```
pezholio, line 2: Could we add `https://raw.githubusercontent.com/nationalarchives/ds-caselaw-marklogic/main/src/main/ml-schemas/` as a constant here? (Something like `SCHEMA_ROOT`) then return the full URL in `to_url`?

pezholio (commented):
Just one comment about making the Schema Root a constant, but looks good I think.

I was thinking about potentially pulling down the schema on deploy, but actually caching is probably a more foolproof way of doing things.

config/converters.py
dragon-dxw: Rewritten the failing tests and added SCHEMA_ROOT. They failed because I turned caching on...

pezholio (approved):
LGTM 👍

PR #1104, opened 2024-02-19
Add model for pdf logic

pezholio:
This attempts to start tidying up some of the logic around PDF handling and have it in one model class. There's still some tidying up to do around the presentation layer, which will come later.

### DIFF
--- a/judgments/admin.py
+++ b/judgments/admin.py
@@ -1,6 +1,6 @@
 from django.contrib import admin
 
-from judgments.models import CourtDates
+from judgments.models.court_dates import CourtDates
 
 
 @admin.register(CourtDates)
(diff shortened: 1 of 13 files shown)


judgments/models.py
```diff
@@ -1,8 +1,54 @@
...
 
+import environ
+import requests
 from django.db import models
 from django.db.models import Max, Min
 
+from judgments.utils import formatted_document_uri
+
+
+class DocumentPdf:
```
dragon-dxw, line 13: How much overlap is there with models/utilities/aws.py ? Feels like the internals might more properly belong there, or those functions here.

I think I want this to be a separate file. Happy if we import DocumentPDF here just to provide models.DocumentPDF instead of models.document_pdf.DocumentPDF

dragon-dxw (approved):

judgments/models.py
pezholio: Cool, I've separated out the models into their own files now

PR #1106, opened 2024-02-20
SPIKE of transactional license form...

timcowlishaw:
## Changes in this PR:

...using django formtools wizards and gov.uk notify for email sending

### After


adam-murray:
Closing as it looks like this was a SPIKE for code that is already on main - feel free to correct me @timcowlishaw

PR #1107, opened 2024-02-20
Attempt to fix broken PDF fetching

dragon-dxw:
### After


pezholio (approved):
LGTM - sorry I caused so much chaos on my way out the door! 😊

PR #1108, opened 2024-02-21
Revert "(Chore) Add presentation logic to a Presenter class"

dragon-dxw:
I was having problems with this locally so want to just get back to a sensible state first.

Reverts nationalarchives/ds-caselaw-public-ui#1105


pezholio (approved):
Reluctantly approving this 😢

dragon-dxw:
We'll get it back in soon enough I hope! :)

PR #1109, opened 2024-02-21
Reinstate (Chore) Add presentation logic to a Presenter class"

dragon-dxw:
Reverts nationalarchives/ds-caselaw-public-ui#1108
Reinstates Reverts nationalarchives/ds-caselaw-public-ui#1105


dragon-dxw:
There's nothing like `linked_doc_url` on the Document class, but there is a function in judgments/utils.py.

The Judgment / PressSummary classes should have access to that function.

And it needs tests.

PR #1110, opened 2024-02-21
weasy_pdf no longer has name, hardcode reference instead for now

dragon-dxw:
Not sure what the best fix for the redirect is -- should probably be a reverse, but that's not possible any more.

The generated.xml endpoint had a name, but that went away in the refactor, so now we need to be a bit more explicit.


pezholio (approved):

CristinaRO (approved):
Approved on the understanding that Dragon knows what they're doing, even if I lack enough context to really understand the underlying complexities!

PR #1126, opened 2024-02-29
Show message when no exact match found for an NCN search

CristinaRO:
## Changes in this PR:
- Show an informational message on the first page of search results for a neutral citation-shaped query for which FCL did not find an exact match

https://trello.com/c/42XNOrBq/1732-implement-anthonys-designs-for-no-results-found

### Before
Screenshot 2024-02-29 at 17 19 37

### After
Screenshot 2024-02-29 at 17 18 35

### DIFF
--- a/ds_judgements_public_ui/static/css/includes/ncn_not_found.css
+++ b/ds_judgements_public_ui/static/css/includes/ncn_not_found.css
@@ -1,0 +1,12 @@
+.tna-message{
+    margin-bottom: 2rem;
+    padding: 1rem;
+    display: flex;
+    gap: 1rem;
+    align-items: flex-start;
+    background-color: rgb(249, 247, 226);
+    /* border-left: 0.5rem #26262a solid; */
+    border-left: 0.5rem var(--keyline-dark) solid;
+    border-left-color: rgb(255, 204, 0);
+
+}
--- a/judgments/templatetags/search_results_filters.py
+++ b/judgments/templatetags/search_results_filters.py
@@ -1,6 +1,6 @@
 from django import template
 
-from judgments.utils import is_exact_ncn_match as _is_exact_ncn_match
+import judgments.utils
 from judgments.utils import preprocess_title
 
 register = template.Library()
@@ -18,7 +18,7 @@ def is_exact_title_match(result, query):
 
 @register.filter
 def is_exact_ncn_match(result, query):
-    return _is_exact_ncn_match(result, query)
+    return judgments.utils.is_exact_ncn_match(result, query)
 
 
 @register.filter
--- a/judgments/templatetags/search_results_filters.py
+++ b/judgments/templatetags/search_results_filters.py
@@ -1,24 +1,9 @@
-import re
-
 from django import template
 
-from judgments.utils import normalise_spaces, preprocess_query
+from judgments.utils import is_exact_ncn_match as _is_exact_ncn_match
+from judgments.utils import preprocess_title
 
 register = template.Library()
-
-not_alphanumeric = re.compile("[^a-zA-Z0-9]")
-
-
-def replace_parens(string):
-    return normalise_spaces(re.sub("\\(.+\\)", "", string))
-
-
-def preprocess_title(string):
-    return preprocess_query(replace_parens(string)).lower().strip()
-
-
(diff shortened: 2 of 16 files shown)


ds_judgements_public_ui/static/css/includes/ncn_not_found.css
```diff
@@ -0,0 +1,12 @@
+.tna-message{
```
dragon-dxw, line 1: I think the filename of this should be something more generic since it's not actually NCN not found specific, perhaps?

judgments/templatetags/search_results_filters.py
```diff
@@ -33,7 +18,7 @@ def is_exact_title_match(result, query):
 
 @register.filter
 def is_exact_ncn_match(result, query):
-    return preprocess_ncn(query) == preprocess_ncn(result.neutral_citation)
+    return _is_exact_ncn_match(result, query)
```
dragon-dxw, line 21: I'm not sure why I'm not liking this name. I think I'd prefer

```
import judgments.utils
return judgments.utils.is_exact_ncn_match
```

judgments/tests/test_utils.py
```diff
@@ -81,3 +83,65 @@ def test_get_press_summaries_for_document_uri(self, mock_api_client):
...
+            mock.configure_mock(neutral_citation=ncn)
+            return mock
+
+        query_text = "walrus"
+        page = "1"
+        search_results = [
+            mock_judgment(ncn) for ncn in ["[2013] EWSC 123", "[2014] EWHC 4122 (Fam)"]
+        ]
+        result = show_no_exact_ncn_warning(search_results, query_text, page)
+        self.assertFalse(result)
```
dragon-dxw, line 147: Surprised there's no positive test.

judgments/views/advanced_search.py
```diff
@@ -138,13 +141,18 @@ def advanced_search(request):
             context["per_page"] = per_page
             context["filtered"] = has_filters(context["query_params"])
             context["page_title"] = gettext("results.search.title")
+            context["show_no_exact_ncn_warning"] = show_no_exact_ncn_warning(
+                search_response.results, query_text, page
+            )
+            # TODO: rethink nonono logic?
```
dragon-dxw, line 147: We should have a think about these TODOs.

judgments/tests/test_utils.py
```diff
@@ -81,3 +83,65 @@ def test_get_press_summaries_for_document_uri(self, mock_api_client):
 
         result = get_press_summaries_for_document_uri("sample_uri")
         self.assertEqual(result, summaries)
+
+    def test_show_no_exact_ncn_warning_is_true_when_no_exact_results_for_ncn_shaped_query(
```
dragon-dxw, line 87: I think I'd prefer shorter, more cryptic function names and docstrings; there's a lot of negatives!

dragon-dxw (approved):

ds_judgements_public_ui/static/css/includes/ncn_not_found.css
dragon-dxw: In hindsight I think `tna-message` this might literally be the name of this component -- the message component from the tna design system.
https://nationalarchives.github.io/design-system/components/message/

So `copied-from-tna-design-system.css` ?

judgments/templatetags/search_results_filters.py
dragon-dxw: > In Python, single underscore `_` before an object name means that the object is meant to be private, and shouldn't be directly accessed from outside the class.

is I think why I don't like it; to me it's implying it's a private variable, rather than a name-mangled public externa one.

judgments/templatetags/search_results_filters.py
```diff
@@ -33,7 +18,7 @@ def is_exact_title_match(result, query):
 
 @register.filter
 def is_exact_ncn_match(result, query):
-    return preprocess_ncn(query) == preprocess_ncn(result.neutral_citation)
+    return judgments.utils.is_exact_ncn_match(result, query)
```
CristinaRO, line 21: @dragon-dxw Is this what you meant?

PR #1128, opened 2024-03-04
Include tna-frontend component library.

timcowlishaw:
## Changes in this PR:

This was a little more involved than I thought - it required both an upgrade to the version of sass we're using, and some changes to the specificity of various rules (as tna-frontend seems to override the link colour in a manner that takes precedence over the CSS cascade in ways that are mysteriour to me).

This should however allow us to start including tna-frontend components in our pages though, and I think we should work towards further integration with tna-frontend, the next step would be to use the grid and the theming capability for layout, typography, and colours as much as possible.

https://trello.com/c/fD1rbQD2/1741-discussion-around-using-tna-frontend-component-library-for-ui-needed-for-ncn-match-homepage-work

### DIFF
--- a/ds_judgements_public_ui/sass/main.scss
+++ b/ds_judgements_public_ui/sass/main.scss
@@ -1,8 +1,17 @@
 // Setup
-@import "includes/variables";
-@import "includes/mixins";
+@use "includes/variables" as *;
+
+// Override variables in tna-frontend
+@use "../../node_modules/@nationalarchives/frontend/nationalarchives/variables/colour"
+  with (
+  $link-colour: $color-link-blue
+);
+// Include all of tna-frontend
+@use "../../node_modules/@nationalarchives/frontend/nationalarchives/all";
 
 // Core
+@import "includes/variables";
+@import "includes/mixins";
 @import "includes/layout";
 @import "includes/spacing";
 @import "includes/typography";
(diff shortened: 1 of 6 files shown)


timcowlishaw:
Tagging in @ahosgood for information - if i'm doing anything wildly stupid or odd-looking here, please let me know!  I'll also raise an issue on the tna-frontend repo with some feedback

ds_judgements_public_ui/sass/main.scss
```diff
@@ -1,8 +1,17 @@
...
+// Override variables in tna-frontend
+@use "../../node_modules/@nationalarchives/frontend/nationalarchives/variables/colour"
+  with (
+  $link-colour: $color-link-blue
+);
+// Include all of tna-frontend
+@use "../../node_modules/@nationalarchives/frontend/nationalarchives/all";
 
 // Core
+@import "includes/variables";
```
ahosgood, line 13: Is this now a duplicate import of line 1?

timcowlishaw:
Aha it is - i think we need to @use it above in order to pass it to
tna-frontend, but also @include it so it's available (unscoped) for our
other CSS? I may have misunderstood though.

ahosgood:
`@use` should import the code in just the same way that `@import` does. The difference is that you can use many `@use` statements in your codebase so everything has access to the module but the code that's imported will only be included once.

You _should_ be fine to just use the `@use` but I haven't looked very deep into your setup.

ahosgood:
Looking at it, I don't think it matters. You have no CSS in your `includes/_variables.scss` so there shouldn't be an issue. 👍🏼

ahosgood (approved):

PR #1129, opened 2024-03-04
Bugfix/date filter lost on change sort

timcowlishaw:
## Changes in this PR:

 - Fixed bug where date filters were lost on changing the page sort order / number of results per page. I also fixed an underlying bug in `parse_date_parameter` whereby if the date was provided as a string parameter (eg. `from=2021-01-01`) the parse function would return a string, but if i provided seperate date part parameters (eg. `from_year=2021&from_month=01&from_day=01`) then it would return a `date`. The function will now return a date object in all cases.
 - Made some tests less brittle by making them ignore whitespace when matching against http responses

https://trello.com/c/xNlAGB0Z/1746-date-filters-broken-on-pui-on-hold-till-ncn-match-work-done

### DIFF
--- a/ds_judgements_public_ui/templates/includes/recent_judgments.html
+++ b/ds_judgements_public_ui/templates/includes/recent_judgments.html
@@ -13,7 +13,7 @@
           </span>
           <span>
             <span class="judgment-listing__neutralcitation">{{ item.neutral_citation }}</span>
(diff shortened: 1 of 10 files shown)


timcowlishaw:
[date filters broken on pui - on hold till NCN match work done](https://trello.com/c/xNlAGB0Z/1746-date-filters-broken-on-pui-on-hold-till-ncn-match-work-done)

PR #1130, opened 2024-03-05
New courts index page links to search instead of court homepage

timcowlishaw:
## Changes in this PR:

As discussed in sprint review 4/3/24, to ensure users can refine and filter after arriving at court page, until the persistent search work has been advanced.


timcowlishaw:
aha it's not actually visible yet until the homepage changes go live (which
will be feature flagged)

PR #1134, opened 2024-03-07
Rmove tna-frontend and reimplement the advice message component

timcowlishaw:
## Changes in this PR:

tna-frontend was causing more problems than it was solving, so this reverts all the work to include it, as well as adding some of our own css for the 'NCN warning', which i've also abstracted out into a reusable 'advice message' component.

## Screenshots of UI changes:
Screenshot 2024-03-07 at 08 32 40


CristinaRO (approved):
I would prefer it if the ripping out of the styling and the extraction of the 'advice message' component were in separate commits, for better readability and narrative clarity, but it's not a blocker.

ds_judgements_public_ui/templates/judgment/results.html
```diff
@@ -10,14 +10,9 @@ <h1>{% translate "results.search.title" %}</h1>
...
         {% if context.show_no_exact_ncn_warning %}
-          <div class="tna-message">
-            <h class="tna-message__heading">
-            <span class="tna-message__heading-icon">I</span>mportant information
-            </h>
-            <div class="tna-message__message">
-              There is no judgment with the Neutral Citation of {{ context.query }} in our database.
-            </div>
-          </div>
+          {% with message="There is no judgment with the Neutral Citation of %s in our database."|interpolate:context.query %}
```
dragon-dxw, line 13: I dislike this crafting of an interpolate function profoundly but reluctantly admit that it seems to be an established way of doing things.  (There is an `|add:`built-in, but apparently `"2"|add:"2"` is `"4"`)

timcowlishaw:
> I would prefer it if the ripping out of the styling and the extraction of the 'advice message' component were in separate commits, for better readability and narrative clarity, but it's not a blocker.

aah yeah, this occured to me like 5 seconds after i squashed them together, sorry.

ds_judgements_public_ui/templates/judgment/results.html
timcowlishaw: Yeah, i'm not a fan either tbh, but it seems like the 'django way' (`add` i considered but it doesn't let us interpolate the NCN into the middle of the string, as well as being confusingly overloaded on strings of numbers)

PR #1135, opened 2024-03-07
if a court has no crest, dont display anything

timcowlishaw:
## Changes in this PR:

Previously, i'd added the ministry of justice crest as a default and as a test in development, which we don't want. This removes it.

### Before
Screenshot 2024-03-07 at 14 38 39

### After

Screenshot 2024-03-07 at 14 38 55


ds_judgements_public_ui/templates/pages/court_or_tribunal.html
```diff
@@ -9,10 +9,9 @@
 {% block content %}
   <div class="container court-or-tribunal">
     <div class="court-or-tribunal__crest">
```
CristinaRO, line 11: Assuming it's ok if this div is going to be empty for crest-less courts?

CristinaRO (approved):

dragon-dxw (approved):

PR #1153, opened 2024-03-27
(1751) Multi-faceted search

adam-murray:
## Depends on:
https://github.com/nationalarchives/ds-caselaw-marklogic/pull/58
https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/575
https://github.com/nationalarchives/ds-caselaw-public-ui/pull/1169

## Changes in this PR:
This PR implements handling for court and year search facets.

https://trello.com/c/XMzUM1jX/1751-🏠🎛-homepage2-faceted-search-concept-on-staging-for-testing-courtyear

### Before
image

### After
image


judgments/utils/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .utils import *
```
dragon-dxw, line 1: There's a lot of complaints on the internet about how `*` imports are bad, but I think it's fine in this context. We could be explicit about `__all__` in the child modules.

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,19 @@
+from ds_caselaw_utils import courts as all_courts
+from ds_caselaw_utils.courts import CourtNotFoundException
+
+
+def process_facets(facets: dict):
+    """
+        Util for processing court facets
```
dragon-dxw, line 7: """Filter court facets so only real courts remain"""

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,19 @@
+from ds_caselaw_utils import courts as all_courts
+from ds_caselaw_utils.courts import CourtNotFoundException
+
+
+def process_facets(facets: dict):
```
dragon-dxw, line 5: Probably wants to be `process_court_facets`?

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+        Util for processing court facets
+    """
+    unprocessed_facets = facets.copy()
+    court_facets = {}
+    for facet_key, facet_value in facets.items():
+        try:
+            court_code = all_courts.get_by_code(facet_key)
```
dragon-dxw, line 13: I think this could be a lot simpler if we got all the court codes ourselves: we could maybe write something like:

court_codes = [court.get("code") for court in all_courts]
valid_facets = [facet for facet in facet if facet_key in court_codes]
invalid_facets = [facet for facet in facet if facet_key not in court_codes]
return invalid_facets, valid_facets

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Separates facets dict into non-year facets,
+    and year facets
+    """
+    year_facets = {
+        facet_key: facet_value
+        for facet_key, facet_value in facets.items()
+        if facet_key in _valid_years()
+    }
+    unprocessed_facets = {
```
CristinaRO, line 63: Is there no Pythonic way of "just" doing the diff between `facets` and `year_facets`? 😅 

I think part of me is still thinking of filtering the facets by the `name` attribute under which they were returned from MarkLogic, and then we'd know for sure that invalid year keys could be discarded/ignored. (And same for court keys.)

adam-murray: Althought counter-intutitive, this is actually more efficient. Modifying existing dictionaries using pop is a very expensive operation compared to constructing dictionaries.

CristinaRO:
[🏠🎛 Homepage#2: Faceted search concept on staging for testing (court&year)](https://trello.com/c/XMzUM1jX/1751-%F0%9F%8F%A0%F0%9F%8E%9B-homepage2-faceted-search-concept-on-staging-for-testing-courtyear)

judgments/templatetags/query_filters.py
```diff
@@ -68,3 +68,17 @@ def remove_court(query_params, court):
...
+
+
+@register.filter
+def replace_year_in_query(query_params, year):
+    params = dict(query_params)
+    params["from"] = None
+    params["from_day"] = None
+    params["from_month"] = None
+    params["from_year"] = year
+    params["to"] = None
```
dragon-dxw, line 80: Could we delete these parameters instead (if they exist) to make the parameters nicer?

dragon-dxw: Probably not essential for POC.

ds_judgements_public_ui/templates/includes/facets/year_search_facets.html
```diff
@@ -0,0 +1,8 @@
+{% load query_filters court_utils %}
```
dragon-dxw, line 1: check if you need both of these (in multiple places)

ds_judgements_public_ui/templates/includes/facets/court_search_facets.html
```diff
@@ -0,0 +1,10 @@
+{% load query_filters court_utils %}
+<div style="text-align: left;">
+  <b>Courts relevant to your query:</b>
+  {% for court, item_count in context.court_facets.items %}
+    {% if court.code not in context.query_params.court %}
+      <a href="{% url "search" %}?{{ request.GET.urlencode }}&court={{ court.code }}">{{ court.grouped_name }} ({{ item_count }})</a>
```
dragon-dxw, line 6: This feels clunky and we should probably make it tidier in the real thing but OK for now

ds_judgements_public_ui/templates/includes/facets/year_search_facets.html
```diff
@@ -0,0 +1,8 @@
...
+<div style="text-align: left;">
+  <b>Years relevant to your query:</b>
+  {% for year, item_count in context.year_facets.items %}
+    <a href="{% url "search" %}?{{ context.query_params|replace_year_in_query:year }}">{{ year }} ({{ item_count }})</a>
```
dragon-dxw, line 5: This feels a lot cleaner than court_search_facets.

dragon-dxw: I assume it gives things in a nice useful format.

judgments/tests/template_tags/test_query_filters.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class TestQueryFilters(unittest.TestCase):
+    def test_replace_year_in_query(self):
+        query_params = {
+            "from": "",
+            "from_day": "3",
+            "from_month": "9",
+            "from_year": "2009",
+            "to": "",
```
dragon-dxw, line 13: we're modifying "to" and "from" so we should probably test they're removed/blanked as well.

judgments/tests/test_utils.py
```diff
@@ -145,3 +148,28 @@ def mock_judgment(ncn):
...
+        eat_court_code = all_courts.get_by_code("EAT")
+
+        remaining_facets, court_facets = process_court_facets(unprocessed_facets)
+
+        self.assertEqual(remaining_facets, {"INVALID": "5", "": "1"})
+        self.assertEqual(court_facets, {eat_court_code: "3"})
+
+    def test_process_year_facets(self):
+        """
+        process_year_facets returns year facets, and invalid facets.
```
dragon-dxw, line 168: as above

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+ALL_COURT_CODES = [court.code for court in all_courts.get_all()]
+
+
+def _valid_years():
+    """
+    Generate a list of valid years as strings.
+    """
+    today = datetime.date.today()
+    valid_years = range(2003, today.year)
```
dragon-dxw, line 13: This will probably bite us when we import older stuff.

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    return [f"{year}" for year in valid_years]
+
+
+def _get_value_as_number(dict_item):
+    """
+    Given a single dictionary item,
```
dragon-dxw, line 19: `"""Given a facet tuple ("name", "12") return the value as an integer"""`

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    return the value as an integer.
+    """
+    return int(dict_item[1])
+
+
+def _sort_by_number_in_value(unsorted_dict: dict):
+    """
+    Sorts a dictionary by value where
+    the values contain numbers as strings
+    """
```
dragon-dxw, line 29: Does suggest we might just want to make those values ints early in the process.

judgments/utils/search_utils.py
```diff
@@ -0,0 +1,69 @@
...
+    sorted_items = sorted(unsorted_dict.items(), key=_get_value_as_number, reverse=True)
+    return dict(sorted_items)
+
+
+def process_court_facets(facets: dict, current_courts: dict = {}):
+    """
+    Separates facets dict into non-court facets,
+    and court facets.
+    """
+    court_facets = {
```
dragon-dxw, line 39: Already discussed that we want to get court/year out of the original data, but we still want the validation. Okay for now.

ds_judgements_public_ui/templates/includes/facets/year_search_facets.html
CristinaRO: The assumption was wrong at the time, hence the live demo fail 😅 . But we've fixed it now! :D

judgments/utils/search_utils.py
CristinaRO: Agree to do this in future, maybe alongside changes to preserve the facet name?

judgments/tests/test_search.py
```diff
@@ -194,6 +195,74 @@ def test_search_breadcrumbs_with_query_string(
...
+        self, mock_search_judgments_and_parse_response, mock_api_client
+    ):
+        mock_search_judgments_and_parse_response.return_value = FakeSearchResponse()
+        eat_court_code = all_courts.get_by_code("EAT")
+        response = self.client.get("/judgments/search?query=example+query")
+
+        # Desired court_facet is present
+        assert response.context["context"]["court_facets"] == {eat_court_code: "3"}
+        # Blank keys are not present
+        assert {"": "3"}.items() not in response.context["context"][
```
dragon-dxw, line 217: I'd probably write `assert "" not in complicated_dict.keys()`; this will always pass because the value of `""` is `5`.

judgments/tests/test_search.py
```diff
@@ -194,6 +195,74 @@ def test_search_breadcrumbs_with_query_string(
...
+            "court_facets"
+        ].items()
+        # Keys that don't match existing courts are not present
+        assert {"invalid_court": "10"}.items() not in response.context["context"][
```
dragon-dxw, line 221: ditto except the number is fine

judgments/tests/test_search.py
```diff
@@ -194,6 +195,74 @@ def test_search_breadcrumbs_with_query_string(
...
+    def test_populated_year_facets(
+        self, mock_search_judgments_and_parse_response, mock_api_client
+    ):
+        mock_search_judgments_and_parse_response.return_value = FakeSearchResponse()
+        response = self.client.get("/judgments/search?query=example+query")
+
+        # Desired year_facet is present
+        assert response.context["context"]["year_facets"] == {"2010": "103"}
+        # Keys that don't match valid years are not present
+        assert {"1900": "4"}.items() not in response.context["context"][
```
dragon-dxw, line 251: similarly here.

judgments/utils/search_utils.py
dragon-dxw: yep, that's fine.

ds_judgements_public_ui/templates/includes/facets/court_search_facets.html
CristinaRO: Fixed! 😅

ds_judgements_public_ui/templates/includes/facets/year_search_facets.html
CristinaRO: It looks like we do, alas.

PR #1168, opened 2024-04-08
Style Guide PUI

Donna-H:
## Changes in this PR:
Created a Style guide for PUI by customising GDS components based on requirements for FCL Transactional Licence below:
Word counter on text areas
'Form fields' components
Important information component
Information boxes & static info pages for form
Expanding detail box
## Jira Ticket / Rollbar error (etc)
Word counter on text areas
https://national-archives.atlassian.net/browse/FCL-39?atlOrigin=eyJpIjoiZmExNWQxMzZkNzg4NDIyYmE2Mjg2MWVmMDIzMjU0ODQiLCJwIjoiaiJ9
'Form fields' components
https://national-archives.atlassian.net/browse/FCL-23?atlOrigin=eyJpIjoiY2VhYzRjNWY2NGNlNGYxOTg1YmY0ZTI4MTViMjI3YWUiLCJwIjoiaiJ9
Important information component
https://national-archives.atlassian.net/browse/FCL-40?atlOrigin=eyJpIjoiNjhlZDM3MjJmOWFmNDgyMThiN2JjZjVhYWY4YmExYmIiLCJwIjoiaiJ9
Information boxes & static info pages for form
https://national-archives.atlassian.net/browse/FCL-24?atlOrigin=eyJpIjoiZTRkMDY2OTU0MmZjNDBkM2E2NTlmMmM2ODFkYjg0YmUiLCJwIjoiaiJ9
Expanding detail box
https://national-archives.atlassian.net/browse/FCL-25?atlOrigin=eyJpIjoiNzdmM2EyMjBkYjRlNDU5NWI4MDhiNWQ5NmZiYmNkMjIiLCJwIjoiaiJ9
### Before 
no style guide
### After


ds_judgements_public_ui/sass/includes/_gds_page.scss
```diff
@@ -0,0 +1,34 @@
+.govuk-fieldset__legend {
```
timcowlishaw, line 1: Actually one minor organizational thing - i think all of the rules in this file could go in _gds_overrides.scss, and this file be deleted. Would that be ok? it seems simpler!

timcowlishaw: .... in fact, it appears all the rules are already there (and this one isn't being included anywhere) - i think you can just delete it!

timcowlishaw (changes requested):
Apart from one comment, this looks ace to me, Donna.

@dragon-dxw am i right in thinking that /style-guide won't be indexed by search engines unless we *explicitly* include it in robots.txt (which we haven't)? It definitely shouldn't be indexed, but just double checking that that's the default behaviour!

thanks!

Tim

ds_judgements_public_ui/sass/includes/_gds_page.scss
Donna-H: deleted it reappeared after conflicts were being sorted @timcowlishaw

ds_judgements_public_ui/templates/layouts/base.html
```diff
@@ -22,15 +22,11 @@
...
-    {% block javascript %}
-      <script defer src="{% static 'js/dist/app.js' %}"></script>
-      <script defer src="{% static 'js/dist/cookie_consent.js' %}"></script>
-    {% endblock javascript %}
     {% include "includes/gtm/gtm_head.html" %}
   </head>
   <body>
     <script>
-      document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
+      document.body.className += ' js-enabled' + ('noModule' in HTMLScriptElement.prototype ? ' govuk-frontend-supported' : '');
```
dragon-dxw, line 29: I don't follow this

package.json
```diff
@@ -35,7 +35,8 @@
     "IE 11"
   ],
   "dependencies": {
-    "@nationalarchives/ds-caselaw-frontend": "github:nationalarchives/ds-caselaw-frontend#v2.0.2",
+    "@nationalarchives/ds-caselaw-frontend": "github:nationalarchives/ds-caselaw-frontend#v2.0.1",
```
dragon-dxw, line 38: Is there a reason for the lower version?

timcowlishaw (approved):
ok great - confirmed that the page won't be indexed. You might have to rebase  from`main` again, but i think we're good to merge!

dragon-dxw:
Minor issue -- character count and word count appear to be conflated in the examples.

ds_judgements_public_ui/templates/layouts/base.html
timcowlishaw: we now add both the same js-enabled classname, and the govuk-fronted-supported classame, as suggested in [their docs](https://design-system.service.gov.uk/styles/page-template/). The ternary is kinda cryptic, but in general where we have inline scripts i tend to favour brevity so happy with it as is

PR #1177, opened 2024-04-15
Add gitignores for nixOS development environments

timcowlishaw:
## Changes in this PR:

I've finally got a new computer that works! It's not a mac this time, and I'm running [NixOS](https://nixos.org/) on it (which is fantastic! but i won't bore you with the details now.). One thing of note however is that it allows isolated and reproducible per-project build environments which is extremely useful, however, it involves creating a couple of extra files in the project directory. This commit adds those files to the .gitignore :-)


dragon-dxw (approved):

timcowlishaw:
Chucking this in my global gitignores instead!

PR #1184, opened 2024-04-24
Static pages 1-4 and temporary download link to word doc application on the Temporary download version 

timcowlishaw:
## Changes in this PR:
Static pages 1-4 with Page 3 - Temporary download version (download link to word doc application)
https://national-archives.atlassian.net/browse/FCL-83
### After
**Pages 1-3** - temporary download link to word doc application on the Temporary download version page 3

https://github.com/nationalarchives/ds-caselaw-public-ui/assets/75584408/d40d4a2d-11d8-4464-8b06-fa46bbb7f299

**Page 4** - ready for when form is completed to link to Page 3 -Online form version 
page-4


ds_judgements_public_ui/sass/includes/_confirmation.scss
```diff
@@ -0,0 +1,17 @@
+h1.confirmation-steps {
```
timcowlishaw, line 1: I'd like to think a bit about the scope of these rules, and  whether they live in the right place. Are they general overrides to the gov.uk styles to be used everywhere? in which case they should probably go in the gds overrides sass. Otherwise, if they're specific styles used in these pages, we should scope them as such, and use BEM, creating a new `.transactional-licence-form-content` block (or something like that, and scoping these within them), and putting them in a sass file which contains everything for that block.

ds_judgements_public_ui/sass/includes/_start.scss
```diff
@@ -0,0 +1,12 @@
+.start-form {
```
timcowlishaw, line 1: See above comments on _confirmation.scss, the same applies here.

transactional_licence_form/templates/start.html
```diff
@@ -1,5 +1,193 @@
...
+          or by including it in your own product or application, all with certain conditions.
+        </p>
+        <p>You can operate under the Open Justice Licence without needing to apply for one.</p>
+        <p>
+          <em>This licence does not cover computational analysis in bulk.</em>
+        </p>
+      </div>
+      <details class="govuk-details">
+        <summary class="govuk-details__summary">
+          <span class="govuk-details__summary-text">What is possible under the Open Justice Licence</span>
```
timcowlishaw, line 41: This is a more general issue with the gds components that we have to look at, but this element is an example of it: the "focused" style is the standard GDS yellow highlight, rather than our own blue outline (annoyingly i can't screengrab a focused state to show you). We should add something to the gds overrides to use our own focus styles everywhere - I can help with this.

ds_judgements_public_ui/sass/includes/_confirmation.scss
```diff
@@ -0,0 +1,17 @@
...
+  margin-bottom: -1.5rem;
+}
+
+.govuk-warning-text {
```
timcowlishaw, line 5: A few  more tweaks we'll need to make here:

image


 -  The horizontal margins are inconsistent, we should tidy this up.
 -  The icon  should be vertically flush with the top of the text
 -  (I *thought* the icon for the red warning box was an exclamation mark rather than an "i" but i'm going to double check that, as i think if it was so, it should just work)
 - We have two different "warning box" styles within the form (see also ".govuk-notification-banner_red". What difference are we intending to communicate here, if any? if they are being used for different purposes, i don't think that's coming across (the difference is confusing). If they're not being used for different purposes, we should settle on one and use it consistently. This is a design thing rather than a dev thing, so not a criticism of your work, but it'd be good to deal with this while we're here.

ds_judgements_public_ui/sass/includes/_start.scss
```diff
@@ -0,0 +1,12 @@
...
+  font-size: 1rem;
+}
+
+.start-form__box {
```
timcowlishaw, line 5: Not convinced this is the right name for this, it's used to contain _example responses_ and the classname should reflect that.  Also, it could do with a shade more padding:

image

transactional_licence_form/templates/start3.html
```diff
@@ -1,5 +1,351 @@
...
+              </div>
+            </div>
+          </details>
+          <a role="button"
+             tabindex="0"
+             draggable="false"
+             class="govuk-button"
+             data-module="govuk-button"
+             value="I want to perform computational analysis"
+             href="{% url 'transactional-licence-form-page-2' %}">Download as PDF</a>
```
timcowlishaw, line 318: The URL here is wrong, but i suspect that's because we don't have the PDF yet? just noting down so we remember to change it, and to get the pdf!

transactional_licence_form/templates/confirmation.html
```diff
@@ -1,5 +1,88 @@
...
+    <div class="govuk-form-group">
+      <fieldset class="govuk-fieldset" aria-describedby="waste-hint">
+        <legend class="govuk-fieldset__legend">
+          <strong>Have you got details of who is applying for this licence?</strong>
+        </legend>
+        <div class="govuk-checkboxes--small" data-module="govuk-checkboxes">
+          <div class="govuk-checkboxes__item">
+            <input class="govuk-checkboxes__input"
+                   id="orginisation"
+                   name="orginistation"
```
timcowlishaw, line 33: Even though this isn't a "real form", these inputs all need unique ids and names (ie different from each other), and the label `for` property for each should contain the id of the correct one. At the moment, if you click on **any** of the labels, the **first** checkbox toggles.

timcowlishaw (commented):
here you go @Donna-H - a couple of minor snags and a couple more general thoughts, but overall this looks good. There's a couple of things for us to look at together, so just shout when you get to them.

timcowlishaw:
BTW also @Donna-H i've rebased this from my current licence form feature branch so before you start working we should make sure you're all up to date locally. Shout me for help with that!

transactional_licence_form/templates/start.html
timcowlishaw: oops sorry, another tiny thing here: the colour of the text on the dropdown component is slightly off: 
image

Donna-H: Hey Tim colour of text in dropdown header is blue as per the design by Bunnyfoot https://o88kd3.axshare.com/introduction_b.html

timcowlishaw: Hey donna, yep, i think i'd treat the bunnyfoot designs more like wireframes for this type of thing: lets use the colour palette we have already!

ds_judgements_public_ui/sass/includes/_confirmation.scss
Donna-H: @timcowlishaw is this confirmation page obsolete in this branch now ? Is it now suppose to be part of the actual form as step 1 as in the updates in Miro shows crossed off with post it stating first page in application form - if so shall I delete this page on the branch ?
image

ds_judgements_public_ui/sass/includes/_start.scss
Donna-H: @timcowlishaw updated sass file as it is specific _transactional_licence_form_content.scss and BEM 
 block named transactional_licence_form_content.

ds_judgements_public_ui/sass/includes/_start.scss
Donna-H: @timcowlishaw as comment above -updated sass file as it is specific _transactional_licence_form_content.scss and BEM  block named transactional_licence_form_content. Therefore this is now transactional-licence-form-content__example-responses .  Padding added to text in box and between heading and box

Donna-H:
> BTW also @Donna-H i've rebased this from my current licence form feature branch so before you start working we should make sure you're all up to date locally. Shout me for help with that!

@timcowlishaw as discussed in our planning I had no working laptop but today I have started on this by pulling this branch onto my new laptop and making the edits so Tuesday we can discuss

transactional_licence_form/templates/confirmation.html
Donna-H: @timcowlishaw please confirm you have these checkboxes in the actual form then this confirmation.html page will be deleted

timcowlishaw:
Reopening #1198 to merge into the transactional licence form branch, before merging to main.

timcowlishaw:
Hey @Donna-H ,

This looks broadly good to me: I'm going to quickly go through and proof-read the copy (and possibly tweak a couple of small layout things) but the bulk of it is there! Leave it with me and i'll finish merging it shortly

PR #1218, opened 2024-05-28
Use djhtml to fix indentation issues instead of djlint

dragon-dxw:


adam-murray (approved):
Approved - left one small comment, but happy to merge.

PR #1254, opened 2024-06-21
Basic infrastructure for playwright end to end tests

timcowlishaw:
## Changes in this PR:

End-to-end tests with [playwright](https://playwright.dev)!

They live in the `e2e_tests/` folder, and are run in a nicely dockerized fashion with `fab e2etest` as  well as in CI (by default against local, but you can run against staging, prod, or some other URL just by providing an argument).

The only missing piece of the puzzle is how to test things that rely on Marklogic from CI, an issue which i do not know how to fix. However, this will allow us to test eg. the transactional licence form, and hopefully start adding tests for other parts of the site!

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -193,6 +193,25 @@ def test(c, test=None):
         subprocess.run(["docker", "compose", "exec", "django", "pytest", test])
 
 
+@task(optional=["baseUrl"])
+def e2etest(c, baseUrl="http://django:3000"):
+    """
+    Run end-to-end playwright tests against the given base url -
+    the default is the running local django web container.
+    """
+    subprocess.run(
+        [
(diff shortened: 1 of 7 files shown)


fabfile.py
```diff
@@ -193,6 +193,25 @@ def test(c, test=None):
...
+    """
+    Run end-to-end playwright tests against the given base url -
+    the default is the running local django web container.
+    """
+    subprocess.run(
+        [
+            "docker",
+            "compose",
+            "build",
+            "e2e_tests",
```
dragon-dxw, line 207: docker compose build caches, so this shouldn't take forever locally.

dragon-dxw (approved):

PR #1264, opened 2024-06-24
Computational licence form tests

timcowlishaw:
## Changes in this PR:

Add tests for transactional licence form. These consist of a playwright (almost) end-to-end test which drives
  the form UI up to the review screen, asserts that the responses are
  visible, and tests the editing and navigation functionality, plus some
  unit tests for the actual form submission (which are tricky to do in
  test due to their reliance on sending email to a third party)


transactional_licence_form/tests/test_tlf_utils.py
```diff
@@ -0,0 +1,146 @@
...
+
+from transactional_licence_form import utils
+
+
+class TestTLFUtils(unittest.TestCase):
+    @mock.patch("transactional_licence_form.utils.settings")
+    @mock.patch("transactional_licence_form.utils.EmailSender.__new__")
+    @mock.patch("transactional_licence_form.utils.sanitize_and_format_response_as_xml")
+    def test_send_form_response(self, mock_sanitize, mock_email_sender, mock_settings):
+        from_email = "from@example.com"
```
adam-murray, line 12: Probably worth sorting this into AAA? Could add comments to break up the arranges, because I understand there's a lot of them!

transactional_licence_form/tests/test_tlf_utils.py
```diff
@@ -0,0 +1,146 @@
...
+    def test_sanitize_and_format_response_as_xml(self):
+        """
+        Tests:
+            - that the form data fields are correctly mapped to the tags expected in the CRM system,
+            - that country names are correctly resolved (agent_country)
+            - That the licence holder contact is correctly defaulted i
+                (licence_holder_email, licence_holder_lastname)
+            - That angle brackets are correctly escaped (agent_companyname)
+        """
+        form_data = {
```
adam-murray, line 59: Might be worth having large data structures in a separate file and import it? Makes it a bit easier to read, as well as keep the structure more easily.

adam-murray (approved):
Approved, with some nit picky optional changes!

PR #1268, opened 2024-06-25
Setup Marklogic for end-to-end tests in CI

timcowlishaw:
## Changes in this PR:

This moves the end to end tests to a seperate action, and sets up
marklogic (as in the development environment), populating it with
some sample judgments from the live site, in order to run real
end-to-end tests.

It also includes a minimum-viable end-to-end test which exercises
marklogic (test_home_page.py) - marklogic is called to populate the
"recent judgments" on the homepage, so the homepage would 500 if it
was not available, and this test would fail.

I'm not sure how to make e2e-tests passing a "required status" on PRs - 
is this something a github admin has to do?


.github/workflows/ci.yml
```diff
@@ -83,13 +83,79 @@ jobs:
...
+      - name: Run marklogic docker container
+        run: docker-compose up -d
+        working-directory: "./marklogic"
+
+      - name: Deploy marklogic configuration
+        run: sleep 30 && gradle mlDeploy
+        working-directory: "./marklogic"
+
+      - name: Populate marklogic DB with test data
+        run: python ./development_scripts/populate_from_caselaw.py
```
jacksonj04, line 146: Not strictly my favourite because this isn't guaranteed to be deterministic, but 1,000% better than what we have at the moment!

.github/workflows/ci.yml
```diff
@@ -83,13 +83,79 @@ jobs:
...
+
+      - name: Install requests
+        run: "pip install requests"
+
+      - name: Run marklogic docker container
+        run: docker-compose up -d
+        working-directory: "./marklogic"
+
+      - name: Deploy marklogic configuration
+        run: sleep 30 && gradle mlDeploy
```
jacksonj04, line 142: (as per [Slack convo](https://dxw.slack.com/archives/C04FQ3GFGUQ/p1719300443189719)) we should be able to ditch this `sleep` by adding a [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) to the Dockerfile pointing to the [MarkLogic healthcheck](https://docs.marklogic.com/11.0/guide/release-notes/en/new-features-in-marklogic-11/enhanced-health-check-monitoring.html), and then sticking `--wait` on the above `docker-compose up`.

This should cause the entire process to pause until everything is reporting green.

jacksonj04 (changes requested):

anthonyhashemi:
This looks good. I guess we'll have to see how much of a blocker this will be on PRs if the e2e test suite becomes large / spinning up and down the containers takes a while. Should be fine for now, if devs get frustrated at sone point with the length it takes we could then move this to just run on merge to main against deployed instances.

In terms of required checks, I can require ci.yml to pass as a check.

pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-addopts = --ds=config.settings.test --reuse-db --disable-socket --ignore=e2e_tests/
+addopts = --ds=config.settings.test --reuse-db --disable-socket --ignore=e2e_tests/ --ignore=marklogic/
```
anthonyhashemi, line 2: why do we want pytest to ignore the marklogic folder?

timcowlishaw: aha the marklogic folder only exists in CI when we checkout the marklogic repo: there's a file called `test_` something in there which isnt actually a pytest test, and which causes pytest to fail.

The marklogic tests are still run when we merge marklogic to main, so i don't think it's necessary to run them here

jacksonj04 (approved):
🚀

dragon-dxw (approved):

PR #1269, opened 2024-06-25
Ignore addons-dev (which might contain api-client tests which should not be tested)

dragon-dxw:


adam-murray (approved):

jacksonj04 (changes requested):
Should we add this to `.gitignore` as well?

jacksonj04 (approved):
🚢

PR #1270, opened 2024-06-25
[FCL-136] restyle the no results message using the govuk-warning-text

jlhdxw:
## Changes in this PR:

 - Changed the styling of the no results message to match the other style being used
 - Removed the specific "no results" styling
 - Brought the other "advice message" in line with the style guide
 - Removed the specific "advice message" styling
 - Adjusting the results layout to handle when there is 2 warning boxes

https://national-archives.atlassian.net/browse/FCL-136

### Before

image

### After

image


ds_judgements_public_ui/sass/includes/_results.scss
```diff
@@ -53,4 +53,10 @@
     padding: $spacer-unit calc($spacer-unit * 2);
     margin: $spacer-unit 0;
   }
+
+  &__list-layout-container {
+    .govuk-warning-text {
+      margin-bottom: 1.5rem;
```
jacksonj04, line 59: Should this be a multiplier of `$spacer-unit`, or is this actually a fixed unit?

jacksonj04 (approved):
One small consideration (I don't know the answer to, it could be entirely correct), but other than that it LGTM!

ds_judgements_public_ui/sass/includes/_results.scss
jlhdxw: Updated this :)

PR #1275, opened 2024-06-26
[FCL-145] Removed nested uses of `context`

jacksonj04:
## Changes in this PR:

Remove nested uses of `context` which make the code harder to understand and reason about.

FCL-145

### DIFF
--- a/config/middleware.py
+++ b/config/middleware.py
@@ -66,8 +66,8 @@ class FeedbackLinkMiddleware:
             "full_url": request.build_absolute_uri(),
         }
 
-        if "query" in response.context_data.get("context", {}):
-            params["search_term"] = response.context_data["context"]["query"]
+        if "query" in response.context_data:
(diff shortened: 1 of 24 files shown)


dragon-dxw (approved):
The tests haven't run properly -- they probably need poking to run again -- and I think a few might fail due to some missing contexts (or that might just be an out of date test)

Looks good!

dragon-dxw:
These look a little suspicious -- can you check if they're still right?
[found by `ag '"context"'`]
judgments/tests/test_search.py
350:        assert "" not in response.context["context"]["court_facets"].keys()
352:        assert "invalid_court" not in response.context["context"]["court_facets"].keys()
366:        assert "" not in response.context["context"]["tribunal_facets"].keys()
369:            "invalid_court" not in response.context["context"]["tribunal_facets"].keys()
399:        assert "1900" not in response.context["context"]["year_facets"].keys()

PR #1278, opened 2024-06-27
FCL-137 | add judgment title and ncn to sticky bar

jlhdxw:
## Changes in this PR:

Adds the judgment title and ncn to the sticky bar. Had to re-jig the sticky bar slightly to allow for other stuff to be put in there.

https://national-archives.atlassian.net/browse/FCL-137

### Before

Desktop without query:
image

Mobile without query:
image

Desktop with query:

image

Mobile with query:

image

### After

Desktop without query:

image

Mobile without query:

image

Desktop with query:

image

Mobile with query:

image


jacksonj04 (approved):
This looks great!

My only thought is... now we have Playwright hooked up, do we want to start adding tests for things like this? Happy to merge this in and then quickly spike on a test, I don't want to hold up an actual fix if getting everything else in order will take days.

ds_judgements_public_ui/templates/judgment/detail.html
```diff
@@ -46,13 +46,20 @@
...
-         href="#start-of-document">Back to start</a>
-    </span>
+  <div class="document-navigation" id="js-document-navigation">
+    <div class="document-navigation__summary">
+      <h2>{{ context.page_title }}</h2>
+      <p>{{ context.judgment_ncn }}</p>
+    </div>
+
+    <div id="js-document-navigation-links-end"
+         class="document-navigation-links {% if context.query %}with-query{% endif %}">
```
dragon-dxw, line 56: this will need to be just "query" -- Nick's removing all the 'context-named-context' bits

dragon-dxw (commented):
LGTM, not approving since @jacksonj04 should be able to get the big refactor out without more contexts popping up!

ds_judgements_public_ui/templates/judgment/detail.html
jlhdxw: Have rebased against main and sorted out the context stuff :) thanks for pointing out

jlhdxw:
@jacksonj04 I saw your comment on slack about maybe not needing a e2e test for this, but I've added one anyway as it would be good to have at least something in place for that page. So it loads a judgment that exists on staging, checks the extra title is not visible, then scrolls and checks it is visible. We can add more checks to this e2e as we do more changes :)

jlhdxw:
@jacksonj04 looks like the e2e tests can't connect to staging - do you know if when it runs on CI it'll be able to access staging data?

jlhdxw:
Spoke with @dragon-dxw and I'm going to look at getting the marklogic instance to run some scripts to populate data but also I'll need to update the populate scripts to also publish the cases which I'll look at doing 1 of 2 ways - either importing the custom api client to publish them after importing, or figuring out what url would allow publishing the judgment on insert.

PR #1289, opened 2024-06-28
[FCL-180] Save user input when pressing previous in licence application form

Dranoel2:
## Changes in this PR:

The previous button in the transactional licence application form was not saving the user's answers, which has now been fixed.

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136/backlog?selectedIssue=FCL-180

### DIFF
--- a/transactional_licence_form/views.py
+++ b/transactional_licence_form/views.py
@@ -34,15 +34,6 @@ class FormWizardView(NamedUrlSessionWizardView):
         return redirect(url)
 
     def post(self, *args, **kwargs):
-        # We override this to ensure that the form is saved when returning to the review page
-
-        wizard_goto_step = self.request.POST.get("wizard_goto_step", None)
-        if not self.in_review():
-            # If we are not reviewing, we follow the default behaviour: we do not validate or save the form
-            # and go directly to the step.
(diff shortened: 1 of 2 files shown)


transactional_licence_form/views.py
```diff
@@ -58,22 +49,30 @@ def post(self, *args, **kwargs):
...
+
+        self.storage.set_step_data(self.steps.current, self.process_step(form))
+        self.storage.set_step_files(self.steps.current, self.process_step_files(form))
+
+        # Going to a step out of order takes priority
+        if wizard_goto_step and wizard_goto_step in self.get_form_list():
+            return self.render_goto_step(wizard_goto_step)
+
+        # Since the user isn't jumping, they must be moving forwards
+        if form.is_valid():
```
jlhdxw, line 66: We could continue the style of single `if` statements and remove the `else` later on by making this:

```pythong
if not form.is_valid():
    return self.render(form)
```

then we just follow up with:

```
if self.steps.current == self.steps.last:
    return self.render_done(form, **kwargs)
```

Finally:

```
return self.render_next_step(form)
```

jlhdxw (approved):
1 minor comment but otherwise LGTM

PR #1297, opened 2024-07-04
version1 updated header for hompage

Donna-H:
## Changes in this PR:
version1 updated header for hompage
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-107
### Desktop view
before-desktop

###  Mobile view
before360
before-375
before-430

### Desktop view
desktop-after

### Mobile view

menu-open-360

after375
open-menu375

after430
menu-open-430

menu-open-430


ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
   margin-bottom: 0;
 }
+
+.govuk-header {
+  border-bottom: none;
+}
+
+.govuk-header__link--homepage:hover,
+.govuk-header__link--homepage:active {
+  border-bottom: $space-1 solid $color-black;
```
jlhdxw, line 196: Thanks for using the new variables :) but let's only use the spacing variables for spacing related CSS - as this is a sizing, let's just set this to 1rem if that's what we would like it to be.
```suggestion
  border-bottom: 0.25rem solid $color-black;
```

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+}
+
+.govuk-header__navigation-item--active a,
+.govuk-header__navigation-item--active a:hover,
+.govuk-header__navigation-item--active a:visited {
+  color: $color-yellow;
+  font-weight: bold !important;
```
jlhdxw, line 203: If we make this styling more specific, we can remove the `!important`:

```suggestion
.govuk-header__navigation-item--active .govuk-header__link,
.govuk-header__navigation-item--active .govuk-header__link:hover,
.govuk-header__navigation-item--active .govuk-header__link:visited {
  color: $color-yellow;
  font-weight: bold;
```

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+
+  &:focus {
+    background-color: transparent;
+    box-shadow: none;
+  }
+}
+
+.govuk-header__navigation-item--active .govuk-header__link,
+.govuk-header__navigation-item .govuk-header__link:hover {
```
jlhdxw, line 212: This styling and the styling above should be merged together as they're targeting the same thing.

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+  color: $color-yellow;
+  font-weight: bold;
+  border-bottom: $space-1 solid $color-yellow;
```
jlhdxw, line 215: Same comment here with regard to spacing - let's just set this to 0.25rem.
```suggestion
  border-bottom: 0.25rem solid $color-yellow;
```

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+  text-decoration: none;
+}
+
+.govuk-header__navigation-item .govuk-header__link {
+  font-size: 0.875rem;
```
jlhdxw, line 220: Can we use `$typography-sm-text-size` here?
```suggestion
  font-size: $typography-sm-text-size;
```

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+    content: attr(title);
+    font-weight: bold;
+    height: 0;
+    overflow: hidden;
+    visibility: hidden;
+  }
+}
+
+a.govuk-header__link {
+  border-bottom: 1px solid $color-white;
```
jlhdxw, line 238: Due to the bold on hover affect we have, the border-bottom looks a bit weird:
image

There's 2 ways I can think of to resolve this:

 - Use `text-decoration: underline` - do you know why we decided to change to border bottom instead of text-decoration? for when we don't have the big yellow border bottom it does look fine with `text-decoration: underline`. If the issue is we want to move the underline down slightly, we can use `text-underline-offset` to move it down a bit.
 - Centerally aligned the text. This does mean we would still have the extra lines on the left and right though:

image

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+  @media only screen and (max-width: $grid-breakpoint-medium) {
+    border-bottom: 1px solid $color-black;
```
jlhdxw, line 240: Rather than setting the border to black, we should either set it to transparent (in case the background colour decides to change in the future) or ideally just set the border-bottom to none here.

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+  }
+}
+
+.govuk-header__menu-button:focus {
+  outline: $space-1 solid transparent;
```
jlhdxw, line 245: Let's change this to the actual size rather than using spacing values
```suggestion
  outline: 0.25rem solid transparent;
```

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+  color: #ffffff;
+  background-color: transparent;
+  text-decoration: none;
+  -webkit-box-decoration-break: clone;
+  box-decoration-break: clone;
```
jlhdxw, line 250: As far as I can tell this doesn't seem to do anything in this instance.

ds_judgements_public_ui/sass/includes/_phase_banner.scss
```diff
@@ -26,6 +26,17 @@
     font-weight: 700;
   }
 
+  &__beta-phase {
+    background-color: $color-yellow;
+    color: $color-black;
+    text-transform: uppercase;
+    padding: $space-1 $space-1;
+    margin-left: $space-1;
+    vertical-align: $space-1;
```
jlhdxw, line 35: If we style `.service-introduction__header` with flex-box and align everything centrally, we can guarantee this will be central in all devices and in all instances/accessibility settings, i.e.:

```
.service-introduction__header {
    display: flex;
    align-items: center;
    justify-content: center;
    gap: 0.25rem;
}
```

Then this can become:

```
&__beta-phase {
    background-color: $color-yellow;
    color: $color-black;
    text-transform: uppercase;
    padding: $space-1;
    font-weight: 700;
    font-size: $typography-sm-text-size;
}

File: ds_judgements_public_ui/sass/includes/_service_introduction.scss [hunk 11]
```diff
@@ -1,16 +1,22 @@
...
+
     width: 100%;
     font-family: $font-roboto;
     font-weight: normal;
     color: $color-white;
-    margin: 0;
+    margin: 0 $space-1;
+    @media only screen and (max-width: $grid-breakpoint-medium) {
+      display: inline-block;
+      float: right;
```
jlhdxw, line 18: We should avoid using floats as much as possible - we shouldn't need it in this situation, especially if we style with flex box as mentioned in the previous comment

ds_judgements_public_ui/sass/includes/_service_introduction.scss
```diff
@@ -1,16 +1,22 @@
 .service-introduction {
   background-color: $color-almost-black;
```
jlhdxw, line 2: I've noticed this is different to the main background colour. It's hard to see but if you look closely it's not the same:

image

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -186,3 +186,90 @@
...
+}
+
+.govuk-frontend-supported .govuk-header__menu-button {
+  right: $space-16;
```
jlhdxw, line 254: I've noticed when trying this out and clicking around on mobile it's very easy to click the logo instead of the menu as the clickable area for the logo is so big:

image

(The blue shows the clickable area)

We should look at reducing that size on mobile.

Donna-H: @jlhdxw thanks updated with an override to gov logo link sizing

Donna-H (commented):
@jlhdxw please do another review after I have updated based on your comments and @Terry-Price feedback earlier

ds_judgements_public_ui/sass/includes/_govuk_overrides.scss
```diff
@@ -193,35 +193,29 @@
...
 
-.govuk-header__navigation-item--active .govuk-header__link,
-.govuk-header__navigation-item .govuk-header__link:hover {
-  color: $color-yellow;
-  font-weight: bold;
-  border-bottom: $space-1 solid $color-yellow;
-  text-decoration: none;
-}
-
 .govuk-header__navigation-item .govuk-header__link {
```
jlhdxw, line 212: I think we need to add a focus style somewhere around here as when you click a nav item it goes black and we can't see it anymore. Also the hover affect has disappeared where it goes yellow and bold.

Donna-H: @jlhdxw checked with Terry and hover should only be taking away the underline not making it bold yellow so updated it thanks for spotting

jlhdxw (approved):
Unit test will come in a follow up PR. Approving to get this in :)

PR #1307, opened 2024-07-10
bug fix header spacing and helper text

Donna-H:
## Changes in this PR:
bug fix header spacing and helper text
https://national-archives.atlassian.net/browse/FCL-214
### Before
Helper text 
before-original
logo block spacing
before-spaceV2

### After
Helper text 
after
logo block spacing
after-space

### DIFF
--- a/ds_judgements_public_ui/templates/pages/home.html
+++ b/ds_judgements_public_ui/templates/pages/home.html
@@ -23,7 +23,7 @@
           <div class="service-introduction">
             <span class="service-introduction__separator"></span>
             <h1 class="service-introduction__header">Find case law<span class="phase-banner__beta-phase">Beta</span></h1>
-            <p class="service-introduction__helper-text">{% translate "home.useservice" %}</p>
+            <p class="service-introduction__helper-text">Judgments and decisions from 2001 onwards</p>
           </div>
           <div class="govuk-header__content">
             <nav aria-label="Menu" class="govuk-header__navigation">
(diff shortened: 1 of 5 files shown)


ds_judgements_public_ui/templates/pages/home.html
```diff
@@ -73,7 +73,7 @@ <h1 class="service-introduction__header-beta">Find case law</h1>
         <div class="service-introduction">
           <span class="service-introduction__separator"></span>
           <h1 class="service-introduction__header">Find case law</h1>
-          <p class="service-introduction__helper-text">{% translate "home.useservice" %}</p>
```
jlhdxw, line 76: More a question than a comment - why are we translating anything? Is there a plan to make the site have more than 1 language?

jlhdxw (approved):

ds_judgements_public_ui/templates/pages/home.html
dragon-dxw: Long version -- https://github.com/nationalarchives/ds-find-caselaw-docs/blob/main/doc/adr/0015-translations.md
Short version -- we did it early on and then decided it was a bad idea, so whilst we haven't committed time to removing all the translations, we're generally in favour of swapping it out for plain text.

Welsh language UI is not on the roadmap. It's not impossible that we get a dual Welsh/English judgment at some point -- we might just wind up concatenating them.

PR #1308, opened 2024-07-10
FCL-147 | Extract breadcrumbs out of base.html

jlhdxw:
## Changes in this PR:

 - Extracts breadcrumbs out of base.html into it's own include
 - Removes the need for blocks (they don't work with includes in django) within templates that extend the base.html

https://national-archives.atlassian.net/browse/FCL-147

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -12,6 +12,10 @@ from .converters import SchemaFileConverter
 
 register_converter(SchemaFileConverter, "schemafile")
 
+handler404 = views.NotFoundView.as_view()
+handler500 = views.ServerErrorView.as_view()
+handler403 = views.PermissionDeniedView.as_view()
+
 urlpatterns = [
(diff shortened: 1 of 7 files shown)


config/urls.py
```diff
@@ -12,6 +12,10 @@
 
 register_converter(SchemaFileConverter, "schemafile")
 
+handler404 = views.NotFoundView.as_view()
```
dragon-dxw, line 15: These don't seem to get used?

dragon-dxw (approved):

config/urls.py
jlhdxw: So the way these works in django is that when they are defined here, they are used over the defaults. If you check the code below, that's only set when in debug mode (which is useful for testing but not actually used in reality). If you don't set these explicitly, the normal django error views are used

PR #1310, opened 2024-07-11
FCL-147 | Improve HTML assertion helper

jlhdxw:
## Changes in this PR:

Changes our html assertion checker for tests to use their actual nodes and check them rather than comparing strings with whitespace removed.

https://national-archives.atlassian.net/browse/FCL-215

### DIFF
--- a/judgments/tests/test_assertions.py
+++ b/judgments/tests/test_assertions.py
@@ -1,0 +1,54 @@
+import pytest
+from django.test import TestCase
+from judgments.tests.utils.assertions import assert_contains_html, assert_not_contains_html
+
+
+class MockResponse:
+    def __init__(self, content):
+        self.content = content.encode()
+
+
+class TestAssertions(TestCase):
+    contained_html = "<title>Judgment A - Find case law - The National Archives</title>"
+    response_contains = MockResponse("<html>" + contained_html + "</html>")
+    response_not_contains = MockResponse("<html>Something else</html>")
+
+    def test_assert_contains_html(self):
+        try:
+            assert_contains_html(self.response_contains, self.contained_html)
+        except AssertionError:
+            pytest.fail("assert_contains_html raised an AssertionError")
+
+    def test_assert_not_contains_html(self):
+        try:
+            assert_not_contains_html(self.response_not_contains, self.contained_html)
+        except AssertionError:
+            pytest.fail("assert_not_contains_html raised an AssertionError")
+
+    def test_assert_contains_html_fails(self):
+        with pytest.raises(AssertionError):
+            assert_contains_html(self.response_not_contains, self.contained_html)
+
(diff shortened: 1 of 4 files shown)


judgments/tests/utils/assertions.py
```diff
@@ -2,22 +2,55 @@
...
-    assert html.replace(" ", "").replace("\n", "") in response.content.decode().replace(" ", "").replace("\n", "")
+def parse_html(html_string):
+    return html.fromstring(html_string)
 
 
-def assert_not_contains_html(response, html):
-    """
-    Asserts that the given HTML is not contained within the response content.
+def parse_html_fragment(html_string):
+    return html.fragments_fromstring(html_string)
```
dragon-dxw, line 13: Ooh. I did not know about this, very useful. I've always taken to stripping off the outer XML.

Worried about the mismatch in fragment/fragments: do we mean one or many.

judgments/tests/utils/assertions.py
```diff
@@ -2,22 +2,55 @@
...
 
-    Raises:
-        AssertionError: If the HTML is found in the response content.
-    """
-    assert html.replace(" ", "").replace("\n", "") not in response.content.decode().replace(" ", "").replace("\n", "")
+
+def normalise_text(text):
+    return " ".join(text.split())
```
dragon-dxw, line 17: That's pretty elegant.

judgments/tests/utils/assertions.py
```diff
@@ -2,22 +2,55 @@
...
+
+
+def elements_match(element_1, element_2):
```
dragon-dxw, line 20: This feels complicated and could do with some tests. Yes this is starting to feel recursive :D

dragon-dxw (approved):

judgments/tests/utils/assertions.py
jlhdxw: My thinking here was that we're passing in a fragment, and this function will return a list of fragments, maybe I could rename it something like `parse_html_fragment_as_fragments` ?

judgments/tests/test_assertions.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TestAssertions(TestCase):
+    contained_html = "<title>Judgment A - Find case law - The National Archives</title>"
+    response_contains = MockResponse("<html>" + contained_html + "</html>")
+    response_not_contains = MockResponse("<html>Something else</html>")
+
+    def test_assert_contains_html(self):
+        try:
+            assert_contains_html(self.response_contains, self.contained_html)
+        except AssertionError:
```
dragon-dxw, line 19: Not sure we gain anything from these trys around the assertions

dragon-dxw (approved):

PR #1312, opened 2024-07-12
Fix per_page control having selected value selected

dragon-dxw:
Changes to how user-input numbers are handled means the number of options per page are  no longer strings in the template, so comparing against "50" is wrong.


jacksonj04 (approved):
This is what we get for using strict typing 😆

PR #1325, opened 2024-07-15
fix(Search): Fix 'no exact match' regression (so it appears)

dragon-dxw:
Pages are no longer strings. Have searched for `page.*'/d` (and `"` variant) to try to catch all other occurances (mostly tests)

Not sure if it's the right behaviour that the no-no-no box still appears... probably.

image

### DIFF
--- a/judgments/tests/test_search.py
+++ b/judgments/tests/test_search.py
@@ -38,6 +38,17 @@ class TestBrowseResults(TestCase):
         self.assertContains(response, "A SearchResult name!", html=True)
 
 
+class TestNoNCN(TestCase):
+    @patch("judgments.views.browse.api_client")
+    @patch("judgments.views.advanced_search.search_judgments_and_parse_response")
+    def test_browse_results(self, mock_search_judgments_and_parse_response, mock_api_client):
+        mock_search_judgments_and_parse_response.return_value = FakeSearchResponseNoResults()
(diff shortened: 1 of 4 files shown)


judgments/tests/test_search.py
```diff
@@ -38,6 +38,17 @@ def test_browse_results(self, mock_search_judgments_and_parse_response, mock_api
...
 
 
+class TestNoNCN(TestCase):
+    @patch("judgments.views.browse.api_client")
+    @patch("judgments.views.advanced_search.search_judgments_and_parse_response")
+    def test_browse_results(self, mock_search_judgments_and_parse_response, mock_api_client):
+        mock_search_judgments_and_parse_response.return_value = FakeSearchResponseNoResults()
+        response = self.client.get(r"/judgments/search?query=[2024]%20EAT%209999")
+        self.assertContains(
+            response, "There is no judgment with the Neutral Citation of [2024] EAT 9999 in our database.", html=True
```
jlhdxw, line 48: If you care about where this appears on the page, you can use the assert_response_contains_text assertion helper

jlhdxw (approved):
LGTM

judgments/tests/test_search.py
dragon-dxw: Ooh. I can.

PR #1339, opened 2024-07-23
Do not cache transactional licence

dragon-dxw:
The difficulties we've been with the transactional licence not loading in the data might very well be caused by caching -- only happens remotely (no caching locally), is slightly erratic, **appears fixable by cache-busting query parameters**. Either way, it's not good to cache pages that might emit user data.


dragon-dxw:
Nick recommends https://docs.djangoproject.com/en/5.0/topics/http/decorators/#django.views.decorators.cache.never_cache

jlhdxw (approved):

PR #1347, opened 2024-07-30
FCL-234 | Handle multiple courts in params when rebuilding query params

jlhdxw:
## Changes in this PR:

Query params where there were multiple values for a param were getting lost. This ensures we keep them.

https://national-archives.atlassian.net/browse/FCL-234

### DIFF
--- a/judgments/views/advanced_search.py
+++ b/judgments/views/advanced_search.py
@@ -66,7 +66,7 @@ def advanced_search(request: HttpRequest) -> HttpResponse:
         return HttpResponseBadRequest("GET requests only")
     else:
         form: AdvancedSearchForm = AdvancedSearchForm(request.GET)
-        params: dict = request.GET
+        params = request.GET
         """
         Form should be valid unless there is a critical issue
         with the submission (i.e. Month > 12)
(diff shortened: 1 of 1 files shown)


judgments/views/advanced_search.py
```diff
@@ -66,7 +66,7 @@ def advanced_search(request: HttpRequest) -> HttpResponse:
         return HttpResponseBadRequest("GET requests only")
     else:
         form: AdvancedSearchForm = AdvancedSearchForm(request.GET)
-        params: dict = request.GET
+        params = request.GET
```
dragon-dxw, line 69: did the typing cause a problem?

judgments/views/advanced_search.py
```diff
@@ -163,8 +163,14 @@ def advanced_search(request: HttpRequest) -> HttpResponse:
                 ) = process_court_facets(search_response.facets, courts_and_tribunals)
                 unprocessed_facets, year_facets = process_year_facets(unprocessed_facets)
 
-            changed_queries = {key: value for key, value in params.items() if value is not None and not key == "page"}
+            changed_queries = {}
+            for key in params:
+                if params.getlist(key) and key != "page":
```
dragon-dxw, line 168: I'd consider setting params.getlist(key) to a variable, but this is splitting hairs.

dragon-dxw: The typing here feels weird -- either things are multielement lists or not a list at all.

dragon-dxw (approved):

judgments/views/advanced_search.py
jlhdxw: Yep - params is not a dict (dict's don't have a function getitems)

judgments/views/advanced_search.py
jlhdxw: Good point - have simplified this

PR #1350, opened 2024-08-01
Set up context processors for waffle for multi-directional feature flags

dragon-dxw:
Waffle and django do not natively support using flags as variables. But we can use a context processor to convert between the two.


dragon-dxw:
Has issues with caching, maybe?

dragon-dxw (commented):
Think this is ready for review now.

jacksonj04 (changes requested):
I think the `misc/poetry.lock` has been committed by accident, and one comment (not crucial but nice for completeness) about a code path not being exercised by tests.

PR #1358, opened 2024-08-06
Remove old feature flags.

dragon-dxw:
These flags can be removed from prod to no effect.
https://national-archives.atlassian.net/browse/FCL-183


dragon-dxw:
image
image
:+1:

Donna-H (approved):
all okay !

PR #1362, opened 2024-08-07
Fix inconsistent capitalisation errors in dockerfiles

dragon-dxw:
There were errors complaining about "FROM x as y" in docker files.


jlhdxw (approved):
LGTM

PR #1375, opened 2024-08-12
[FCL-255] Disable global cache behaviours

jacksonj04:
This PR disables global caching behaviours to allow for serving different responses on a request-by-request basis. This is necessary for allowing A/B testing of the homepage and the header variations.

The intent is that this will be rolled out to identify if/where any actual improved caching is needed, with a rapid rollback in case of resource issues.

## Jira

FCL-255

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -131,7 +131,6 @@ MIDDLEWARE = [
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
-    "config.middleware.CacheHeaderMiddleware",
     "config.middleware.RobotsTagMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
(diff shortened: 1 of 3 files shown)


config/settings/base.py
```diff
@@ -259,7 +258,7 @@
 # Your stuff...
 # ------------------------------------------------------------------------------
 APPEND_SLASH = False
-WHITENOISE_MAX_AGE = 900  # set Cache-Control header on static-served images
+WHITENOISE_MAX_AGE = 3600  # set Cache-Control header on static-served images to one hour
```
dragon-dxw, line 261: This is a very sensible change,

config/tests.py
```diff
@@ -8,20 +8,10 @@
 
 
 class TestCacheHeaders(TestCase):
-    def test_static_headers(self):
```
dragon-dxw, line 11: We could `@xfail` these.

dragon-dxw (approved):
I'd approach it very differently, leaving the substantial bits in, but this approach is legit.

PR #1378, opened 2024-08-16
FCL-228 | add document paragraph anchors links when they are on the page

jlhdxw:
## Changes in this PR:

Adds a script that sets all the paragraph sections that have paragraph numbers with IDs to have their number turned into a link that links to that section.

I'm not sure if there are some reports where this may cause issues? If anyone knows of any I can bake that into the script.

Are we happy with how the paragraph numbers look like links? We could style this as text but still have them as links so those who know are able to get the links easily.

## Screenshots of UI changes:

Makes the paragraph numbers look like this:

image


dragon-dxw:
I'd be tempted to suggest a copy icon rather than a link: image

I think the numbers not being visibly hyperlinks might be an improvment; it's a lot of visual clutter.

dragon-dxw (approved):

PR #1386, opened 2024-08-22
FCL-228 | add waffle tags to the window so we can toggle features on/…

jlhdxw:
## Changes in this PR:

This loops through all the waffle flags and sets them on the window as waffleFlags so we can do javascript based a/b testing.

We have an object explicitly set with some specific flags in this same context preprocessor - as a follow up, it might be worth refactoring that to use this waffle_flags object, but I didn't want to do all that in this PR.


dragon-dxw (approved):
I think I'd like to see a test; "does `{"flag": true, "other-flag": false}` appear?"

I've used override_flag (see https://waffle.readthedocs.io/en/stable/testing/automated.html) in similar tests.

PR #1388, opened 2024-08-22
FCL-228 | update tooltip variant text

jlhdxw:
## Changes in this PR:

Adds the tooltip variant for the paragraph linking.

## Screenshots of UI changes:

New variant looks like this:

image

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_judgment_document_paragraph_anchors.scss
+++ b/ds_judgements_public_ui/sass/includes/_judgment_document_paragraph_anchors.scss
@@ -7,7 +7,8 @@
     }
 
     &:hover {
-      .judgment-body__copy-link {
+      .judgment-body__copy-link,
+      .judgment-body__copy-tooltip {
         opacity: 1;
       }
     }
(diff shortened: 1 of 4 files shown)


ds_judgements_public_ui/sass/includes/_judgment_document_paragraph_anchors.scss
```diff
@@ -7,7 +7,8 @@
     }
 
     &:hover {
-      .judgment-body__copy-link {
+      .judgment-body__copy-link,
+      .judgment-body__copy-tooltip {
```
dragon-dxw, line 11: Should this one be helptext? We've search-replaced tooltip elsewhere.
(Might also be worth a code search for tooltip to see if we're missing any)

jlhdxw: This one is correct. Basically we have tooltip and helptext. Helptext in the "icon" variation is just text that explains what to do. The "tooltip" version is the thing you actually have to click.

PR #1403, opened 2024-08-27
[FCL-262] increase cache TTL for WhiteNoise-served assets

jacksonj04:
These static assets are currently being cached for 900 seconds (15 minutes). Increase to 1 hour to improve cache hit rates for these.


dragon-dxw (approved):
Doesn't quite catch all the test changes, but good in principle.

PR #1406, opened 2024-08-28
[FCL-266] Add in-memory cache for getting documents

jacksonj04:
Use `lru_cache` to keep the most recently requested 256 documents in application memory.

Use [this workaround](https://stackoverflow.com/questions/31771286/python-in-memory-cache-with-time-to-live) to emulate cache TTL, by introducing a new parameter which 'rolls over' every 900 seconds, giving a maximum lifetime of 15 minutes for each document before it will be re-requested from MarkLogic.

## Linked documents

Detecting where a document doesn't exist is currently done using a `DocumentNotFoundError` exception, which _isn't_ cached. We add a new parameter which will make `get_document_by_uri` return `None`, which can be cached, in the case of responses where we consider a "this doesn't exist" state to be more robust, such as in the case of linked documents.

## Check page

The page at `/check` now returns some telemetry about the state of the document cache, instead of just a plain "OK" response.

```json
{
  "status": "ok",
  "document_cache": {
    "hits": 4,
    "misses": 11,
    "current_size": 6,
    "max_size": 256
  }
}
```

## Jira

FCL-266


judgments/utils/utils.py
```diff
@@ -186,9 +186,17 @@ def has_filters(query_params, exclude=["order", "per_page"]):
...
 
 
-def get_document_by_uri(document_uri: str) -> Document:
+def get_document_by_uri(document_uri: str, max_ttl: int = 900) -> Document:
+    ttl_hash = round(time() / max_ttl)
+    return get_document_by_uri_from_cache(DocumentURIString(document_uri), ttl_hash=ttl_hash)
+
+
+@lru_cache(maxsize=256)
+def get_document_by_uri_from_cache(document_uri: DocumentURIString, ttl_hash: int = 0) -> Document:
```
dragon-dxw, line 195: Docstrings would be appreciated, this is quite subtle. Recommend a link to the SO answer.

judgments/utils/utils.py
```diff
@@ -186,9 +186,17 @@ def has_filters(query_params, exclude=["order", "per_page"]):
     return len(set(k for (k, v) in query_params.items() if v) - set(exclude)) > 0
 
 
-def get_document_by_uri(document_uri: str) -> Document:
+def get_document_by_uri(document_uri: str, max_ttl: int = 900) -> Document:
+    ttl_hash = round(time() / max_ttl)
+    return get_document_by_uri_from_cache(DocumentURIString(document_uri), ttl_hash=ttl_hash)
+
+
+@lru_cache(maxsize=256)
```
dragon-dxw, line 194: This feels like a reasonable number. There's a negligible risk that no-one uses the site and just asks for the Megajudgment every quarter hour and that'd be of the order of a gigabyte...

dragon-dxw (approved):

config/views/check.py
```diff
@@ -1,5 +1,19 @@
...
 
-class CheckView(TemplateViewWithContext):
-    template_name = "pages/check.html"
+
+def status(request):
+    document_cache_info = get_document_by_uri_from_cache.cache_info()
+
+    return JsonResponse(
+        {
+            "status": "ok",
```
dragon-dxw, line 11: I'd be tempted to make this `OK` so `"OK"` appears in the output, in case that was being tested for.

judgments/views/detail.py
```diff
@@ -35,9 +35,11 @@ class NoNeutralCitationError(Exception):
     logging.getLogger("weasyprint").handlers = []
 
 
-def get_published_document_by_uri(document_uri: str) -> Document:
+def get_published_document_by_uri(document_uri: str, cache_if_not_found: bool = False) -> Document:
     try:
-        document = get_document_by_uri(document_uri)
+        document = get_document_by_uri(document_uri, cache_if_not_found=cache_if_not_found)
```
dragon-dxw, line 40: Pondering whether there are security issues with the cache, since different users have different access -- no, because the editor UI and public UI are completely isolated.

dragon-dxw (approved):

judgments/views/detail.py
jacksonj04: I did wonder this - there's potentially a max 15 minute window where a document which was previously published and has been yanked would still be visible, but I'm not unduly worried about this.

In due course we should add some kind of API to the PUI which lets us trigger things like cache flushing based on EUI interactions.

PR #1409, opened 2024-08-28
[FCL-253] Fix narrow content on variant 3 homepage

jacksonj04:
This is down to some unnecessary nesting of classes which each add their own padding or margins; remove and simplify.

### Before

Find case law - The National Archives · 5 10pm · 08-28

### After

Find case law - The National Archives


dragon-dxw (approved):
Please check it looks okay on Mobile widths before merging.

PR #1411, opened 2024-08-29
FCL-264 | add search breadcrumb if search query present on judgment d…

jlhdxw:
## Changes in this PR:

Adds a search breadcrumb to the judgment breadcrumb if there is a search query.

https://national-archives.atlassian.net/browse/FCL-264

## Screenshots of UI changes:

image

### DIFF
--- a/ds_judgements_public_ui/templates/includes/judgment_listing.html
+++ b/ds_judgements_public_ui/templates/includes/judgment_listing.html
@@ -14,8 +14,8 @@
       <span class="judgment-listing__judgment">
         <span>
           <span class="judgment-listing__title">
-            {% if query %}
-              <a href="{% url 'detail' item.uri %}?query={{ query | urlencode }}">{{ item.name }}</a>
+            {% if query_param_string %}
+              <a href="{% url 'detail' item.uri %}?{{ query_param_string }}">{{ item.name }}</a>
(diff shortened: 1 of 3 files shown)


dragon-dxw (approved):

dragon-dxw:
How does this look if there's a search for one of the courts? Is it 'Search results for ""'?

jacksonj04 (changes requested):
So... this works, but I'm not convinced it actually meets the user needs, mostly because it will never include any further parameters used in the search query.

For example, an advanced search at https://caselaw.nationalarchives.gov.uk/judgments/search?per_page=10&order=relevance&query=cheese&from_date_0=&from_date_1=&from_date_2=&to_date_0=&to_date_1=&to_date_2=&court=uksc&party=&judge= only ever passes the `query=cheese` parameter to the judgment page, which in turn means anybody using the breadcrumbs to navigate back will miss a lot of that original search context.

I'm not sure what the best solution here would be; passing the whole context is messy but possibly right?

jlhdxw:
@jacksonj04 yes I missed that, good catch. I think I'll go with adding all the search query params when the user navigates to a judgment so we can add those to the breadcrumb link. I think that's the simplest approach without having to store it in some kind of state

PR #1423, opened 2024-09-02
[FCL-275] Cache the latest judgments search on the home page

jacksonj04:
This will cache the list of judgments on the home page for up to 15 minutes, reducing database load.

## Jira

FCL-275

### DIFF
--- a/judgments/views/index.py
+++ b/judgments/views/index.py
@@ -3,12 +3,24 @@ from caselawclient.client_helpers.search_helpers import (
     search_judgments_and_parse_response,
 )
 from caselawclient.search_parameters import SearchParameters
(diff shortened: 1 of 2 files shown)


judgments/views/index.py
```diff
@@ -3,22 +3,34 @@
...
 from django.views.generic import TemplateView
 from ds_caselaw_utils import courts as all_courts
+from time import time
+from functools import lru_cache
 
 from judgments.forms import AdvancedSearchForm
 from judgments.utils import api_client
 
 
+@lru_cache(maxsize=4)
```
dragon-dxw, line 17: good value of maxsize, we need to remember this for the future in case we have more variants...

dragon-dxw (approved):

PR #1442, opened 2024-09-12
[FCL-289] Search based atom feeds

dragon-dxw:
There's a few decisions to be made: what URL does this appear at, do we do content-type header or URI based etc. Also includes a refactoring of some of the search logic.

Could do with a few tests of behaviour but they probably want to wait for the URL decisions.

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-290


judgments/urls.py
```diff
@@ -26,7 +26,7 @@
         BrowseView.as_view(),
         name="browse",
     ),
-    path("atom.xml", feeds.LatestJudgmentsFeed(), name="feed"),
+    path("atom.xml", feeds.SearchJudgmentsFeed(), name="search-feed"),
     path("<court:court>/atom.xml", feeds.LatestJudgmentsFeed(), name="feed"),
```
jacksonj04, line 30: Could we make these just push a redirect to `/atom.xml` with appropriate params?

judgments/feeds.py
```diff
@@ -139,3 +115,90 @@ def feed_extra_kwargs(self, obj):
         extra_kwargs["total"] = int(obj["search_response"].total)
         extra_kwargs["page"] = obj["page"]
         return extra_kwargs
+
+
+class LatestJudgmentsFeed(JudgmentsFeed):
+    def link(self, obj):
+        page = obj.get("page", 1)
+        return f"/{obj['slug']}/atom.xml/?page={page}&order={obj.get('order')}"
```
jacksonj04, line 123: One errant slash (I think):

```suggestion
        return f"/{obj['slug']}/atom.xml?page={page}&order={obj.get('order')}"
```

judgments/forms/search_forms.py
```diff
@@ -94,9 +94,10 @@ class AdvancedSearchForm(forms.Form):
     def clean(self):
         cleaned_data = super().clean()
         # Validate that from is before to now that we have access to both fields
-        # Cleaned data can never be `None`, so supress warning
-        to_date = cleaned_data.get("to_date")  # type: ignore
-        from_date = cleaned_data.get("from_date")  # type: ignore
+        if cleaned_data is None:  # make the type checker happier
+            raise RuntimeError("Cleaned data can never be None, this should never occur")
```
jacksonj04, line 98: Could we change this to a `return None` and keep the signature intact?
```suggestion
            return None
```

judgments/tests/test_atom_feed.py
```diff
@@ -45,11 +48,54 @@ def test_feed_exists(self, mock_api_client, mock_search_judgments_and_parse_resp
...
+            decoded_response,
+        )
+
+    @patch("judgments.feeds.search_judgments_and_parse_response")
+    def test_bad_page_not_500(self, mock_search_judgments_and_parse_response):
+        # "?page=" doesn't 500
         mock_search_judgments_and_parse_response.return_value = FakeSearchResponse()
         response = self.client.get("/atom.xml?page=")
-        self.assertEqual(response.status_code, 404)
+        assert response.status_code != 500
```
jacksonj04, line 98: Shouldn't this really be a 400? The user has provided an invalid parameter (`page` is `None`)

judgments/tests/test_search.py
```diff
@@ -83,8 +83,8 @@ def test_judgment_advanced_search_with_populated_court_dates(
             SearchParameters(
                 query="waltham forest",
                 court="",
-                judge="",
-                party="",
+                judge=None,
```
jacksonj04, line 86: Hooray for more strictly correct data types!

jacksonj04 (approved):
Couple of tiny comments, but substantially it's good to go!

judgments/forms/search_forms.py
dragon-dxw: Raising should keep the return value intact?

judgments/tests/test_atom_feed.py
dragon-dxw: It's not invalid to not specify a page; due to changes in the handling, an empty page is now handled as 1, so we get a 200. I could explicitly test for that.

jacksonj04: We should probably make this match the spec as much as possible, so if we could explicitly require page (tbh any parameter) to be of the expected type that's probably preferable...

But also I don't feel massively strongly here. This just feels like a non-test; what case is it actually testing?

dragon-dxw: I think this test was added to avoid regressions of a bug where users could provoke 500 errors. But yeah, we can ditch it.

judgments/tests/test_atom_feed.py
```diff
@@ -91,17 +91,29 @@ def test_search_query_in_URL(self, mock_api_client, mock_search_judgments_and_pa
...
+    def test_feed_with_empty_date(self, mock_search):
+        mock_search.return_value = FakeSearchResponse()
 
         response = self.client.get("/atom.xml")
         decoded_response = response.content.decode("utf-8")
         self.assertEqual(response.status_code, 200)
         self.assertIn("A SearchResult name!", decoded_response)
+
+    def test_redirect_full(self):
+        response = self.client.get("/ewhc/ch/2024/atom.xml")
```
jacksonj04, line 111: Can we include a test case which covers passing in page and ordering parameters as well?

judgments/feeds.py
```diff
@@ -60,8 +60,15 @@ def root_attributes(self):
 
     def add_item_elements(self, handler, item) -> None:
         super().add_item_elements(handler, item)
-        handler.addQuickElement("tna:uri", item.get("uri", ""))
         handler.addQuickElement("tna:contenthash", item.get("content_hash", ""))
+        path = item.get("uri", "")
+        handler.addQuickElement("tna:uri", path)
+        if path:
+            path_underscore = path.replace("/", "_")
+            handler.addQuickElement("tna:xml", f"/{path}/data.xml")
```
jacksonj04, line 68: These don't feel quite right - what does the output look like? Should we be generating `link rel="alternate"` nodes?

jacksonj04 (changes requested):
Couple of tiny changes, but this is substantially Good and Awesome.

judgments/tests/test_atom_feed.py
dragon-dxw: Done.

judgments/feeds.py
```diff
@@ -60,8 +60,24 @@ def root_attributes(self):
...
+            handler.addQuickElement(
+                "link", "", {"rel": "alternate", "type": "application/akn+xml", "href": f"/{path}/data.xml"}
+            )
+            handler.addQuickElement(
+                "link",
+                "",
+                {
+                    "rel": "alternate",
+                    "type": "application/pdf",
+                    "href": f"https://assets.caselaw.nationalarchives.gov.uk/{path}/{path_underscore}.pdf",
```
jacksonj04, line 78: We should probably push this into a method on the `Document` in future, but not now!

judgments/tests/test_atom_feed.py
dragon-dxw: We deleted it.

PR #1446, opened 2024-09-15
[FCL-284] Allow crawling of the site in robots.txt and disallow indexing using meta tags

jacksonj04:
At the moment our `robots.txt` explicitly disallows crawling of the site. This is nominally to discourage search engines from indexing judgments, however search engines may still index pages where they have a high confidence that they exist and are linked to from other sites.

This means that although those pages are discoverable via search for related terms, search engines cannot scrape the page to build content previews.

The spirit is that pages should not appear in search indexes, and this should instead be achieved by relying on a `noindex` tag in the header of the page.

## Changes in this PR:

- Allow crawling of the whole site by search engines in `robots.txt`
- Make sure documents are explicitly disallowed from indexing using `noindex`, both as HTML `meta` tags and `X-Robots-Tag` HTTP headers.
- `sitemap.xml` is a `sitemapindex`, pointing in turn to maps of:
  - Static pages, enumerated manually in the `sitemaps.py`
  - Court pages, derived from courts data
  - One page per year, per court, listing all judgments and their last transformation date, derived by running a search query
- Clarify in the terms of use that crawling/spidering is fine as long as you don't build an index, and adhere to the robots directives.

## Jira

FCL-284


config/views/sitemaps.py
```diff
@@ -0,0 +1,124 @@
...
+
+class SitemapStaticView(TemplateView, TemplateResponseMixin):
+    content_type = "application/xml"
+    template_name = "sitemaps/sitemap.xml"
+
+    def get_context_data(self, **kwargs) -> dict[str, Any]:
+        context = super().get_context_data(**kwargs)
+
+        # This is a list of the names of URLs to reverse and include in the map
+        url_names = [
```
jlhdxw, line 53: Maybe not for now but it would be useful if we could somehow generate this from urls.py - I can see a case where we add a static page to the urls and forget to add it here.

config/views/sitemaps.py
```diff
@@ -0,0 +1,124 @@
...
+        year: Union[int, None] = self.kwargs.get("year")
+
+        try:
+            search_parameters = SearchParameters(
+                court=court_query if court_query else None,
+                date_from=(datetime.date(year=year, month=1, day=1).strftime("%Y-%m-%d") if year else None),
+                date_to=(datetime.date(year=year, month=12, day=31).strftime("%Y-%m-%d") if year else None),
+                order="-date",
+                page=1,
+                page_size=50000,
```
jlhdxw, line 114: Is the search API okay handling this? 😄 also are we doing this to avoid there being a chance of 2 pages? I feel like we have more than 50,000 records?

jlhdxw (approved):
Some minor comments, LGTM

config/views/sitemaps.py
jacksonj04: Agreed - static files are all over the place at the moment and it would be good to tidy them up.

config/views/sitemaps.py
jacksonj04: Yes; the intersection of court and year should never generate more than a few thousand records at worst (even for the busiest court).

PR #1450, opened 2024-09-16
[FCL-219] Add very minimal ruff configuration

jacksonj04:
This switches on ruff with a very minimal set of rules to start and enforce code consistency. We can incrementally switch on rules over time to bring the codebase into stylistic alignment with everything else.

## Jira

FCL-219


jlhdxw (approved):
Beautiful

PR #1454, opened 2024-09-18
Fcl 163/standardise templates

jlhdxw:
## Changes in this PR:

 - Replaces translations with their string equivilant
 - Fixes an incorrect email address
 - Fixes page titles not getting their translation
 - Removes translations set up


config/settings/base.py
```diff
@@ -29,18 +29,12 @@
 # though not all of them may be available with every OS.
 # In Windows, this must be set to your system time zone.
 TIME_ZONE = "UTC"
-# https://docs.djangoproject.com/en/dev/ref/settings/#language-code
-LANGUAGE_CODE = "en-gb"
```
jacksonj04, line 33: Is this actually no longer needed? Probably not, just doesn't seem directly related to I18N.

jacksonj04 (approved):
It's so much less stuff!

config/views/template_view_with_context.py
```diff
@@ -7,5 +6,5 @@ class TemplateViewWithContext(TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["page_title"] = gettext(self.page_title) if self.page_title else None
+        context["page_title"] = self.page_title
```
dragon-dxw, line 9: This might not be equivalent if self.page_title is None; that might be turned into the string "None", I think

ds_judgements_public_ui/templates/includes/beta_phase_banner.html
```diff
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load %}
```
dragon-dxw, line 1: This tag can probably go? Is this even valid?

dragon-dxw (commented):
Looks good, some minor things to check.

config/views/template_view_with_context.py
jacksonj04: I think it should be retained as a `None` until it hits the template, which should then be the layer which deals with the problem.

config/settings/base.py
jlhdxw: Looking at the docs, in django it's only used for localisation. I couldn't see any more references of it either so looks safe to remove

config/views/template_view_with_context.py
jlhdxw: Have handled everywhere page_title is used to do a None check

ds_judgements_public_ui/templates/includes/beta_phase_banner.html
jlhdxw: Removed

PR #1478, opened 2024-09-24
[FCL-322] Fix for court years being off by one in the sitemap

jacksonj04:
We were using `range()` to output all the years we consider to be valid for a court, but this is _exclusive_ of the last value.

Fix this, and also swap to using our from-database `CourtDates` model as the original source of the courts and their date ranges. This means that when years roll over we aren't inadvertently publicising a non-existent range of values for a court.

## Jira

FCL-322

### DIFF
--- a/config/views/sitemaps.py
+++ b/config/views/sitemaps.py
@@ -9,8 +9,8 @@ from caselawclient.search_parameters import SearchParameters
 from django.http import Http404
 from django.urls import reverse
 from django.views.generic.base import TemplateResponseMixin, TemplateView
-from ds_caselaw_utils import courts
 
+from judgments.models.court_dates import CourtDates
 from judgments.utils import api_client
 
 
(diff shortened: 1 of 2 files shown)


config/views/sitemaps.py
```diff
@@ -83,12 +83,8 @@ def get_context_data(self, **kwargs) -> dict[str, Any]:
...
 
         context["items"] = [
-            {
-                "url": self.request.build_absolute_uri(
-                    reverse("court_or_tribunal", kwargs={"param": court.canonical_param})
-                )
-            }
-            for court in courts.get_listable_courts()
+            {"url": self.request.build_absolute_uri(reverse("court_or_tribunal", kwargs={"param": court.param}))}
+            for court in CourtDates.objects.all()
```
dragon-dxw, line 87: Not sure I understand why we have non-listable courts in the data now.

dragon-dxw (approved):

PR #1493, opened 2024-10-03
[FCL-367] Improve how we provide canonical URLs as part of page metadata

jacksonj04:
This PR improves how we declare canonical URLs in page metadata. This allows search engines to link to the version of a page we consider to be correct, regardless of any other details which might have been appended to the incoming URL (eg search query terms, advertising tracking parameters and so-on).

As part of this we also bring our various static views into a new `static.py` file to make it easier to reason about the changes, and reorganise `urls.py` so it's easier to see where any particular URL comes from. These changes do not have any impact in terms of functionality.

## Jira

FCL-367


config/views/static.py
```diff
@@ -0,0 +1,145 @@
+from ds_caselaw_utils import courts
+
+from .template_view_with_context import TemplateViewWithContext
```
dragon-dxw, line 3: Is it worth TemplateViewWithContext handling the feedback_survey_type context item, based on a value in the subclass (if it exists?)

dragon-dxw (approved):

PR #1494, opened 2024-10-03
[FCL-368] Improve how we flag pages as not being indexable

jacksonj04:
**This is based on the work in #1493 - review/merge that one first and rebase this!**

This was previously split between an in-template HTML tag, and a HTTP tag which relied on a whitelist. It was possible for these two things to become out of sync. Now, the view decides if a page should be indexable, by explicitly setting `context["page_allow_index"]` to `True`. `TemplateViewWithContext` will simplify this process for pages which use it.

For all instances where `page_allow_index` is not explicitly set to True, the `base.html` template will inject a `noindex,nofollow` meta tag and the `RobotsTagMiddleware` will inject a `noindex,nofollow` `X-Robots-Tag` HTTP header into the `response` object.

This establishes a pattern which can be used in FCL-304 of specifying these page metadata elements once in the view, and then them being reused in various contexts as necessary.

## Jira

FCL-368

### DIFF
--- a/judgments/views/index.py
+++ b/judgments/views/index.py
@@ -32,6 +32,7 @@ class IndexView(TemplateView):
         context = super().get_context_data(**kwargs)
 
         context["page_canonical_url"] = self.request.build_absolute_uri(reverse("home"))
+        context["page_allow_index"] = True
 
         try:
             search_response = cached_recent_judgments(ttl_hash=round(time() / 900))  # Expire cache in 15 mins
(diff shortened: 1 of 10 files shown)


judgments/views/index.py
```diff
@@ -32,6 +32,7 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
         context["page_canonical_url"] = self.request.build_absolute_uri(reverse("home"))
+        context["page_allow_index"] = True
```
dragon-dxw, line 35: I'm not following the logic of why this class should support indexing yet.

dragon-dxw (approved):

PR #1495, opened 2024-10-04
[FCL-366] Fix for misaligned focus

Donna-H:
## Changes in this PR:
fix hover on the cards at bottom of static pages
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-366
### 
before
Before

### After
after

### DIFF
--- a/ds_judgements_public_ui/sass/includes/govuk_overrides/_grid.scss
+++ b/ds_judgements_public_ui/sass/includes/govuk_overrides/_grid.scss
@@ -155,7 +155,7 @@
   max-width: 23rem;
   height: 16rem;
   padding: $space-1 $space-5;
-  margin: 10px;
+  margin: $space-2;
   border-top: 4px solid $color-yellow;
   background-color: $color-light-grey;
   text-decoration: underline;
@@ -193,7 +193,7 @@
   max-width: 23rem;
   height: 16rem;
   padding: $space-1 $space-5;
-  margin: 10px;
+  margin: $space-2;
   border-top: 4px solid $color-yellow;
   background-color: $color-light-grey;
   text-decoration: underline;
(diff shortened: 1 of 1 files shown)


ds_judgements_public_ui/sass/includes/govuk_overrides/_grid.scss
```diff
@@ -155,9 +155,7 @@
   max-width: 23rem;
   height: 16rem;
   padding: $space-1 $space-5;
-  margin-top: -15px;
-  margin-top: 20px;
-  margin-left: -9px;
+  margin: 10px;
```
jlhdxw, line 158: Can we use a spacing value here to ensure our spacing is consistent? I think $space-3 should be about equivalent here. Same comment to the other 10px below

ds_judgements_public_ui/sass/includes/govuk_overrides/_grid.scss
```diff
@@ -140,10 +140,10 @@
 .govuk-grid-column-two-thirds__info_terms_use {
```
jacksonj04, line 140: Just with an eye to simplifying and reducing duplication; is there a reason these classes are so specific?

Donna-H: @jacksonj04 yes it was based on design checks with Terry to make the cards align correctly as text amopunt etc and titles are different

ds_judgements_public_ui/sass/includes/govuk_overrides/_grid.scss
Donna-H: @jlhdxw updated to $space-2 as per design review with Terry it ligns up best to 10px thank you

ds_judgements_public_ui/sass/includes/govuk_overrides/_grid.scss
jacksonj04: If we're doing something where each box needs different styles with some straightfoward text then there's probably something else we should be looking at, but that's a separate issue! 😄

jlhdxw (approved):

PR #1506, opened 2024-10-08
FCL-163 | remove document_noun from context

jlhdxw:
## Changes in this PR:

Removing explicit document_noun value from context in favour of accessing the document directly

### DIFF
--- a/ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
+++ b/ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
@@ -2,10 +2,10 @@
 
 <div class="judgment-toolbar">
   <div class="judgment-toolbar__container">
-    {% if document_noun == "press summary" %}
+    {% if document.document_noun == "press summary" %}
       <p class="judgment-toolbar__press-summary-title">Press Summary</p>
     {% endif %}
-    <h1 class="judgment-toolbar__title">{{ page_title|get_title_to_display_in_html:document_noun }}</h1>
+    <h1 class="judgment-toolbar__title">{{ document|get_title_to_display_in_html }}</h1>
     <p class="judgment-toolbar__reference">{{ judgment_ncn }}</p>
     <div class="judgment-toolbar__buttons judgment-toolbar-buttons">
       {% if linked_document_uri %}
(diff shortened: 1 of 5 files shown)


ds_judgements_public_ui/templates/includes/judgment_text_toolbar.html
```diff
@@ -2,10 +2,10 @@
 
 <div class="judgment-toolbar">
   <div class="judgment-toolbar__container">
-    {% if document_noun == "press summary" %}
+    {% if document.document_noun == "press summary" %}
       <p class="judgment-toolbar__press-summary-title">Press Summary</p>
     {% endif %}
-    <h1 class="judgment-toolbar__title">{{ page_title|get_title_to_display_in_html:document_noun }}</h1>
+    <h1 class="judgment-toolbar__title">{{ document|get_title_to_display_in_html }}</h1>
```
jacksonj04, line 8: I think I might cry with happiness.

ds_judgements_public_ui/templates/judgment/detail.html
```diff
@@ -12,9 +12,9 @@
 {% endblock precontent %}
 {% block content %}
   <div id="start-of-document"></div>
-  {% if document %}
+  {% if document_html %}
```
jacksonj04, line 15: Would `document.get_content_as_html` do the trick?

judgments/templatetags/document_utils.py
```diff
@@ -8,14 +8,14 @@
 
 
 @register.filter
-def get_title_to_display_in_html(document_title, document_noun):
-    if not document_title:
+def get_title_to_display_in_html(document):
```
jacksonj04, line 11: Super picky - could we add type hints for this?

jacksonj04 (approved):

ds_judgements_public_ui/templates/judgment/detail.html
jlhdxw: In the context it passes in:

```
MOST_RECENT_VERSION,
        query=preprocess_query(query) if query is not None else None,
```

Which I think if we added those to the context and passed them in in the template we've got the same issue.  I think having this one in the context is fine as it's not just an attribute on the document

judgments/templatetags/document_utils.py
jlhdxw: Have done that (and added to the function below too)

PR #1509, opened 2024-10-08
FCL-163 | split out details views

jlhdxw:
## Changes in this PR:

Split the details views into separate files to make it easier to read/find the appropriate views.

I also want to convert these all to classes, which will be a follow up PR

### DIFF
--- a/judgments/views/detail/detail_html.py
+++ b/judgments/views/detail/detail_html.py
@@ -1,24 +1,18 @@
 import logging
 import os
 import urllib
-from typing import Any, Union
+from typing import Any
 
-import requests
-from caselawclient.errors import DocumentNotFoundError, MarklogicNotPermittedError
-from caselawclient.models.documents import Document, DocumentURIString
-from django.conf import settings
-from django.http import Http404, HttpResponse, HttpResponseRedirect
-from django.shortcuts import redirect
+from caselawclient.models.documents import DocumentURIString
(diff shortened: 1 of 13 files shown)


judgments/views/detail/detail_html.py
```diff
@@ -37,69 +31,7 @@ class NoNeutralCitationError(Exception):
...
-    return redirect(
-        reverse(
-            "detail",
-            kwargs={"document_uri": document_uri, "file_format": "generated.pdf"},
-        )
-    )
-
-
-def detail(request, document_uri):
+def detail_html(request, document_uri):
```
jacksonj04, line 34: We should strictly speaking make this a classy view, but one problem at a time!

judgments/views/detail/detail_xml.py
```diff
@@ -0,0 +1,22 @@
...
+import os
+
+from django.http import HttpResponse
+
+from judgments.utils import (
+    get_published_document_by_uri,
+)
+
+# suppress weasyprint log spam
+if os.environ.get("SHOW_WEASYPRINT_LOGS") != "True":
```
jacksonj04, line 11: The XML rendering shouldn't need to care about Weasyprint at all.

jacksonj04 (approved):
Substantially happy with this, just a couple of tidy-up comments.

PR #1510, opened 2024-10-08
update licence info pages to new design

Donna-H:
## Changes in this PR:
update licence info pages to new design
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-356
### Before
[before.webm](https://github.com/user-attachments/assets/bab5aace-72ad-4a9f-8269-bee4ecd91c1e)

### After

[after.webm](https://github.com/user-attachments/assets/bb9caa02-f711-4868-8e8c-d0b223ba1f67)

### DIFF
--- a/ds_judgements_public_ui/templates/pages/open_justice_licence.html
+++ b/ds_judgements_public_ui/templates/pages/open_justice_licence.html
@@ -4,165 +4,176 @@
         content="Open justice is a fundamental constitutional principle and necessary for the rule of law. The purpose of this licence is to support open justice." />
 {% endblock meta_description %}
 {% block static_content %}
-  <h1>Open Justice Licence</h1>
-  <div class="anchor-links">
-    <h3 class="anchor-links__header">Information on this page:</h3>
-    <ul class="anchor-links__list">
-      <li class="anchor-links__list-option">
-        <a href="#section-purpose">Purpose of this licence</a>
-      </li>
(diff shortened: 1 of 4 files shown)


dragon-dxw (approved):
LGTM, but probably worth waiting for someone more frontend.

dragon-dxw (approved):
LGTM.

ds_judgements_public_ui/templates/pages/open_justice_licence.html
```diff
@@ -4,165 +4,176 @@
...
+              <a href="#section-definitions">Definitions</a>
+            </li>
+            <li class="anchor-links__list-option">
+              <a href="#section-about">About the Open Justice Licence</a>
+            </li>
+          </ul>
+        </div>
+      </div>
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="hidden-h1-desktop">Open Justice Licence</h1>
```
jacksonj04, line 39: I wonder if the text of the licence itself could be made into an include, because the exact same text is used here and in `start.html`?

Terry-Price (approved):
Looks good @Donna-H I like the new template for these pages.

jlhdxw (approved):
Looks good to me. As a future note - it would be nice if we could build our pages using our own classes that extend the govuk classes if needed (I'm thinking specifically the govuk-grid-row etc classes)

PR #1517, opened 2024-10-09
Fix typing on search_request_to_parameters

dragon-dxw:
## Changes in this PR:
Ensure `search_request_to_parameters` only returns `SearchParameters`, not a HTTP Response in error cases. (The outside function handles the errors anyway.)

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-377

## Screenshots of UI changes:
No change to UI -- still doesn't reflect problem in description.
image

### DIFF
--- a/judgments/utils/search_request_to_parameters.py
+++ b/judgments/utils/search_request_to_parameters.py
@@ -6,7 +6,6 @@ from django.core.exceptions import BadRequest
 from django.http import (
     HttpRequest,
 )
-from django.template.response import TemplateResponse
 
 from judgments.forms import AdvancedSearchForm
 from judgments.utils import (
@@ -17,7 +16,7 @@ from judgments.utils import (
 from judgments.utils.utils import sanitise_input_to_integer
 
 
-def search_request_to_parameters(request: HttpRequest, return_page_if_invalid="pages/structured_search.html"):
+def search_request_to_parameters(request: HttpRequest) -> SearchParameters:
     """
     The advanced search view handles any searches made in the application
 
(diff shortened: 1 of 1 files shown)


judgments/utils/search_request_to_parameters.py
```diff
@@ -17,7 +16,7 @@
 from judgments.utils.utils import sanitise_input_to_integer
 
 
-def search_request_to_parameters(request: HttpRequest, return_page_if_invalid="pages/structured_search.html"):
```
dragon-dxw, line 20: return_page_if_invalid is not present in any callers.

judgments/utils/search_request_to_parameters.py
```diff
@@ -38,71 +37,70 @@ def search_request_to_parameters(request: HttpRequest, return_page_if_invalid="p
     with the submission (i.e. Month > 12)
     """
     if not form.is_valid():
-        # If the form has errors, return it for rendering!
-        return TemplateResponse(request, return_page_if_invalid, {"form": form})
```
dragon-dxw, line 42: The caller for the search page already checks the validity before the call. This isn't a sensible type to return here for the atom feed.

dragon-dxw: The rest of the changes in this commit are just unindentation from making it an `if - return` not `if - return - else`

jacksonj04 (approved):
We definitely want to figure out a nicer way of catching and relaying API (I know this is Atom, but it's really the stub of our API behaviours) "you did something wrong" errors.

That said, that's another problem, so LGTM!

dragon-dxw:
Tested that the search behaviour remains unchanged -- court=ewhc fails with "see below" but there's nothing below, but this behaviour is unchanged.

PR #1523, opened 2024-10-11
Add breadcrumbs to static section pages in landing

Donna-H:
## Changes in this PR:
Add breadcrumbs to static section pages in landing
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-359
## Help and guidance page section cards
cards-help-guidance

### Before
before-help-1st-card
before-help-2nd-card
before-help3rd-card

### After
after-help-1st-card
after-cards-help-2nd-card
after-help-3rd-card

## Terms and policies page section cards
cards-terms-policies

### Before
before-terms-1st-card
before-terms-2nd-card
before-terms-3rd-card

### After
after-terms-1st-card
after-terms-2nd-card
after-terms-3rd-card

### DIFF
--- a/config/views/static.py
+++ b/config/views/static.py
@@ -27,7 +27,7 @@ class AccessibilityStatementView(TemplateViewWithContext):
         context = super().get_context_data(**kwargs)
         context["feedback_survey_type"] = "support"
         context["breadcrumbs"] = [
-            {"url": self.request.build_absolute_uri(reverse("terms_and_policies")), "text": "Terms and policies"},
+            {"url": (reverse("terms_and_policies")), "text": "Terms and policies"},
             {"text": "Accessibility statement"},
         ]
         return context
@@ -43,7 +43,7 @@ class ContactUsView(TemplateViewWithContext):
         context = super().get_context_data(**kwargs)
         context["feedback_survey_type"] = "contact_us"
         context["breadcrumbs"] = [
-            {"url": self.request.build_absolute_uri(reverse("help_and_guidance")), "text": "Help and guidance"},
+            {"url": (reverse("help_and_guidance")), "text": "Help and guidance"},
             {"text": "Contact us"},
         ]
         return context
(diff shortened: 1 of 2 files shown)


config/views/static.py
```diff
@@ -146,6 +159,10 @@ class TermsOfUseView(TemplateViewWithContext):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["feedback_survey_type"] = "support"
+        context["breadcrumbs"] = [
+            {"url": "/terms-and-policies", "text": "Terms and policies"},
+            {"text": "Terms of Use"},
```
jacksonj04, line 164: Should this be lowercase "use" to match everything else being sentence case?

jacksonj04 (changes requested):
Nice to see these making it in! I wonder if it's worth using `reverse()` for the URLs rather than adding them explicitly, just to improve robustness if we move things in future?

Donna-H:
> Nice to see these making it in! I wonder if it's worth using `reverse()` for the URLs rather than adding them explicitly, just to improve robustness if we move things in future?

config/views/static.py
Donna-H: @jacksonj04 thanks for spotting this will update

Donna-H:
> @jacksonj04 never done breadcrumbbs before Tim use to do this so have no idea about reverse
> 
> > Nice to see these making it in! I wonder if it's worth using `reverse()` for the URLs rather than adding them explicitly, just to improve robustness if we move things in future?

@jacksonj04  okay after reading https://docs.djangoproject.com/en/5.0/ref/urlresolvers/ I have found where we have used reverse() previously in our codebase and will update it now then you can check and review it again

jacksonj04 (approved):
If you wanted to tighten it up the `self.request.build_absolute_uri()` isn't strictly necessary. All this will do is add the `https://caselaw.nationalarchives.gov.uk` to the front of the URL, but they'll work fine without it.

PR #1531, opened 2024-10-14
Allow short court codes (e.g. ewhc)

### LINKED ISSUE #1520: Atom feed court filter for `ewhc` results in server error
With the current Atom feed i noticed that only court+subdivision is working but not court directly

using just `ewhc` doesn't work and yields a “Server Error”
❌ https://caselaw.nationalarchives.gov.uk/atom.xml?court=ewhc 

where the example from the docs works returns the expected atom feed
✅ https://caselaw.nationalarchives.gov.uk/atom.xml?court=ewhc/fam

related documentation
https://nationalarchives.github.io/ds-find-caselaw-docs/public#tag/Reading-documents/operation/atomFeed

is this a bug?

dragon-dxw:
Short court codes which have subcourts such as ewhc haven't been valid, and this is a problem for the atom feed. 

Fixes #1520

### DIFF
--- a/judgments/forms/search_forms.py
+++ b/judgments/forms/search_forms.py
@@ -10,6 +10,8 @@ from judgments.utils import preprocess_query
 from .fields import DateRangeInputField
 from .widgets import CheckBoxSelectCourtWithYearRange
 
+court_choices_dict = dict[Union[Optional[str], CourtParam], Union[str, dict[CourtParam, str]]]
+
 
 def _get_choices_by_group(courts: list[CourtGroup]):
     """
@@ -24,20 +26,20 @@ def _get_choices_by_group(courts: list[CourtGroup]):
         "key2": "value2"
     }
     """
-    options: dict[Union[Optional[str], CourtParam], Union[str, dict[CourtParam, str]]] = {}
+    options: court_choices_dict = {}
     for group in courts:
         if group.display_heading:
-            option_a: dict[Union[Optional[str], CourtParam], Union[str, dict[CourtParam, str]]] = {
+            court_group: court_choices_dict = {
                 group.name: {
                     court.canonical_param: court.grouped_name for court in group.courts if court.canonical_param
                 }
             }
-            options = options | option_a
+            options = options | court_group
         else:
-            option_b: dict[Union[Optional[str], CourtParam], Union[str, dict[CourtParam, str]]] = {
+            isolated_court: court_choices_dict = {
                 court.canonical_param: court.grouped_name for court in group.courts if court.canonical_param
             }
-            options = options | option_b
+            options = options | isolated_court
     return options
 
 
--- a/judgments/forms/search_forms.py
+++ b/judgments/forms/search_forms.py
@@ -1,6 +1,9 @@
+from typing import Optional, Union
+
 from django import forms
 from django.forms import ValidationError
 from ds_caselaw_utils import courts as all_courts
+from ds_caselaw_utils.courts import CourtGroup, CourtParam
 
 from judgments.utils import preprocess_query
 
@@ -8,7 +11,7 @@ from .fields import DateRangeInputField
 from .widgets import CheckBoxSelectCourtWithYearRange
 
 
-def _get_choices_by_group(courts):
+def _get_choices_by_group(courts: list[CourtGroup]):
     """
     Given a list of court objects, construct a dictionary containing
     {"key1": "value1"} for non-grouped courts, and {"court_group": {"key1": "value1"}}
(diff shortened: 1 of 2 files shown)


judgments/forms/search_forms.py
```diff
@@ -35,6 +35,24 @@ def _get_choices_by_group(courts):
 TRIBUNAL_CHOICES = _get_choices_by_group(all_courts.get_grouped_selectable_tribunals())
 
 
+def get_short_identifiers(top_level_choices):
+    all_shorts = set()
+    for key, value in top_level_choices:
+        if isinstance(value, list):
+            all_shorts = all_shorts | set(x[0].partition("/")[0] for x in value)
+    return all_shorts
```
jlhdxw, line 43: We could move this under the `CourtTribunalField` class as it's only being used there.

judgments/forms/search_forms.py
```diff
@@ -35,6 +35,24 @@ def _get_choices_by_group(courts):
...
+
+
+class CourtTribunalField(forms.MultipleChoiceField):
+    def validate(self, *args, **kwargs):
+        short_identifiers = get_short_identifiers(self.choices)
+        try:
+            return super().validate(*args, **kwargs)
+        except ValidationError as e:
+            if e.params.get("value") in short_identifiers:
+                return  # acceptable value
```
jlhdxw, line 53: Is it okay to return nothing here? Should we explicitly return `None` or whatever this expects us to return?

judgments/forms/search_forms.py
jlhdxw, line 43: What is value in this instance? it's hard to figure out if `set(x[0].partition("/")[0] for x in value)` might fail under certain types of data. If it's typed it might tell you that there could be a potential time when accessing [0] on x would fail, or partition could fail

dragon-dxw: I saw this as being directly analogous to "_get_choices_by_group" (and probably should have had an underscore); it doesn't operate on a CourtTribunalField so I'm not sure it belongs there.

judgments/forms/search_forms.py
dragon-dxw: Sure, this is purely stylistic, and explicit is better than implicit.

judgments/forms/search_forms.py
dragon-dxw: `top_level_choices` is essentially a list of key-value pairs as a list; [('ewca/civ', 'Civil Division'), ('ewca/crim', 'Criminal Division')]), coming out of a dictionary, so x[0] is almost certainly valid. As long as it's a string, partition("/") will always return three things, the first being "all the text before the `/`", even if there isn't one.

jlhdxw: Okay cool, thanks for explaining

judgments/forms/search_forms.py
```diff
@@ -21,20 +24,47 @@ def _get_choices_by_group(courts):
         "key2": "value2"
     }
     """
-    options: dict = {}
+    options: dict[Union[Optional[str], CourtParam], Union[str, dict[CourtParam, str]]] = {}
```
jlhdxw, line 27: Shall we move this `dict[Union[Optional[str], CourtParam], Union[str, dict[CourtParam, str]]]` to a type alias so we can use it here and on option_a/option_b?

jacksonj04: Agree with moving this to a concrete class, this looks like it's complex enough we want it to stand on its own.

judgments/forms/search_forms.py
```diff
@@ -21,20 +24,46 @@ def _get_choices_by_group(courts):
...
     for group in courts:
         if group.display_heading:
-            option = {group.name: {court.canonical_param: court.grouped_name for court in group.courts}}
+            option_a: dict[Union[Optional[str], CourtParam], Union[str, dict[CourtParam, str]]] = {
```
jacksonj04, line 30: These `option_a` and `option_b` names could be clearer, possibly as part of moving the `options` type definition into an alias so it's clearer what's going on here.

judgments/forms/search_forms.py
```diff
@@ -21,20 +24,46 @@ def _get_choices_by_group(courts):
...
+            if isinstance(value, list):
+                all_shorts = all_shorts | set(x[0].partition("/")[0] for x in value)
+        return all_shorts
+
+    def validate(self, *args, **kwargs):
+        short_identifiers = self._get_short_identifiers()
+        try:
+            return super().validate(*args, **kwargs)
+        except ValidationError as e:
+            if e.params and e.params.get("value") in short_identifiers:
```
jacksonj04, line 61: This path could do with a test.

judgments/forms/search_forms.py
```diff
@@ -21,20 +24,46 @@ def _get_choices_by_group(courts):
...
+            }
+            options = options | option_b
     return options
 
 
 COURT_CHOICES = _get_choices_by_group(all_courts.get_grouped_selectable_courts())
 TRIBUNAL_CHOICES = _get_choices_by_group(all_courts.get_grouped_selectable_tribunals())
 
 
+class CourtTribunalField(forms.MultipleChoiceField):
```
jacksonj04, line 48: I'm not 100% sold on this name, it kinda implies it's the `Tribunal` belonging to a `Court`, which makes no sense.

It's really some kind of horrible abstract `JudicialBodyField`... but I don't like that either.

jacksonj04 (changes requested):
Couple of tiny comments - also just for ease of what relates to what, are these really court codes or are they params?

judgments/forms/search_forms.py
dragon-dxw: Changed to CourtOrTribunalField

judgments/forms/search_forms.py
dragon-dxw: Renamed to isolated_court and court_group

judgments/forms/search_forms.py
dragon-dxw: Tests added at the HTML level.

judgments/forms/search_forms.py
dragon-dxw: Done.

jacksonj04 (approved):
I still don't like the whole court code/param mess, but this is definitely a step closer to it being solved!

PR #1537, opened 2024-10-16
FCL-373 | add and style atom feed button to search results

jlhdxw:
## Changes in this PR:

Add the feed button to search results

https://national-archives.atlassian.net/browse/FCL-373

## Screenshots of UI changes:

Desktop:

image

Mobile:

image

Mobile use to look like this without the atom feed, which isn't very tidy:

image

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_result_controls.scss
+++ b/ds_judgements_public_ui/sass/includes/_result_controls.scss
@@ -10,8 +10,12 @@
 
   form {
     display: flex;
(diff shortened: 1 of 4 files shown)


dragon-dxw (approved):

Terry-Price:
james@pancakeuprising.com
can you use the RSS feed icon that was in the design card?
I like the one wI sourced as the weight was consistent in each element.
or was there a reason for the uneven version?

PR #1544, opened 2024-10-16
update the meta  descriptions static pages

Donna-H:
## Changes in this PR:
update the meta
https://national-archives.atlassian.net/browse/FCL-304
### Before
none-before
before

### After

none-after
after

### DIFF
--- a/ds_judgements_public_ui/templates/layouts/static_content.html
+++ b/ds_judgements_public_ui/templates/layouts/static_content.html
@@ -1,4 +1,8 @@
 {% extends "layouts/base.html" %}
+{% block meta_description %}
+  <meta name="description"
+        content="{% if page_description %}{{ page_description }}{% endif %}" />
+{% endblock meta_description %}
 {% block title %}
   {% if page_title %}{{ page_title }} -{% endif %} Find Case Law
 {% endblock title %}
--- a/ds_judgements_public_ui/templates/layouts/static_content.html
+++ b/ds_judgements_public_ui/templates/layouts/static_content.html
@@ -1,8 +1,4 @@
 {% extends "layouts/base.html" %}
-{% block meta_description %}
-  <meta name="description"
-        content="{% if page_description %}{{ page_description }}{% endif %}" />
-{% endblock meta_description %}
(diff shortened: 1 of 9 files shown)


jlhdxw (approved):

ds_judgements_public_ui/templates/layouts/static_content.html
```diff
@@ -1,4 +1,8 @@
 {% extends "layouts/base.html" %}
+{% block meta_description %}
+  <meta name="description"
```
jacksonj04, line 3: It would be better if this entire meta tag was wrapped in the `if` statement, so we don't render it at all if the description is missing. At the moment it will say the description is just an empty string.

transactional_licence_form/templates/start.html
```diff
@@ -1,4 +1,9 @@
 {% extends "layouts/base.html" %}
+{% block meta_description %}
```
jacksonj04, line 2: I don't think this should be needed if the `context` is set correctly, since the base one should do the job.

jacksonj04 (changes requested):

transactional_licence_form/templates/start.html
Donna-H: @jacksonj04 these are only in the static_content not base as all the static pages have extended from static_content not base so I put it this only in this licence page as the static licence pages are not extended from static_content

dragon-dxw:
builds fine for me locally with --no-cache...

Donna-H:
> > Actually looking at this we should be able to move the entire meta description block into the `base.html` layout and then not need it anywhere else.
> 
> @jacksonj04 just done and moved it

ds_judgements_public_ui/templates/layouts/static_content.html
Donna-H: updated wrapped it @jacksonj04

dragon-dxw:
image

Looks like docker.io is having a time

dragon-dxw (approved):

jacksonj04 (approved):

PR #1547, opened 2024-10-17
Fcl 223/add stylelint

jlhdxw:
## Changes in this PR:

Add stylelint with config and pre-commit hook.

With all the recommended settings, there are a lot of issues. If you run with the `--fix` option, it can fix a lot of them. The issue is the files that have fixes in also have issues that require manual intervention. I would suggest we fix the issues as we go.

https://national-archives.atlassian.net/browse/FCL-223

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,11 +46,15 @@ jobs:
           cache-dependency-path: |
             requirements/base.txt
             requirements/local.txt
-
+      - id: file_changes
+        uses: trilom/file-changes-action@v1.2.4
+        with:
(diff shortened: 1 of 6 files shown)


dragon-dxw (approved):
Excellent work to get this set up -- but I'm loathe to make a CI hook which will cause unrelated changes. I suggest that for now we only run style lint manually, by setting `stages: [manual]` under `- hooks` and invoking with `precommit run stylelint` or similar.

That said, I'm probably not the one having to fix the SCSS files. :)

Definately not adverse to turning this fully on in the near future.

PR #1561, opened 2024-10-22
Replace structured search text pui

Donna-H:
## Changes in this PR:
Replace structured search text pui including url change
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-381
### Before
before

### After
after

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -44,6 +44,11 @@ urlpatterns = [
         StructuredSearchView.as_view(),
         name="advanced_search",
     ),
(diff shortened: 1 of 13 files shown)


Donna-H:
> We should probably add a redirect from the old URL to the new one, so existing links people have don't break.

@jacksonj04 okay added hopefully the correct way , please review

jacksonj04 (approved):
🚢

PR #1579, opened 2024-10-30
[FCL-386] Migrate to using in-app XSLT rather than MarkLogic

jacksonj04:
Now that we're able to run the latest version of the API client with the XSLT no longer relying on MarkLogic, change PUI to run XSLT transforms that way.

## Jira

FCL-386


dragon-dxw (approved):
The non-ASCII branch name might be a problem in future

PR #1582, opened 2024-10-30
[FCL-397] Add XSLT and dynamic titles to Atom results

jacksonj04:
This gives users styled content and useful orientation messages, rather than raw XML.

### DIFF
--- a/judgments/feeds.py
+++ b/judgments/feeds.py
@@ -14,6 +14,8 @@ from django.http.response import HttpResponseRedirect
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.feedgenerator import Atom1Feed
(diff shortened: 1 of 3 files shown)


judgments/feeds.py
```diff
@@ -157,16 +160,46 @@ def feed_extra_kwargs(self, obj):
...
 
     def title(self, obj):
-        if not obj["query_string"]:
-            return "Search results"
+        title_string = "Latest documents"
+
+        if obj["query_string"]:
+            title_string += f' for "{obj["query_string"]}"'
+
+        if obj["court"]:
```
dragon-dxw, line 184: there can be multiple courts

dragon-dxw (approved):

PR #1600, opened 2024-11-04
FCL-439 | add full advanced search e2e test

jlhdxw:
## Changes in this PR:

Adds a full E2E test for advanced search

https://national-archives.atlassian.net/browse/FCL-439

### DIFF
--- a/e2e_tests/test_advanced_search_page.py
+++ b/e2e_tests/test_advanced_search_page.py
@@ -1,13 +1,131 @@
+import re
+
+import pytest
 from playwright.sync_api import Page, expect
 
 from .utils.assertions import assert_is_accessible
 
 
-def test_advanced_search_page(page: Page):
+def query_input(page: Page):
+    return page.locator("#search_form")
+
+
+def submit_button(page: Page):
+    return page.get_by_role("button", name="Submit")
+
+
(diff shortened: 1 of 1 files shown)


dragon-dxw (approved):

e2e_tests/test_advanced_search_page.py
```diff
@@ -1,13 +1,131 @@
...
+
+def submit_button(page: Page):
+    return page.get_by_role("button", name="Submit")
+
+
+def update_filters_button(page: Page):
+    return page.get_by_role("button", name="Update filters")
+
+
+court_filters = [
```
jacksonj04, line 21: This feels like it could be fragile; can we derive this from the utils?

jlhdxw: I would love to do that, but the E2E docker app currently isn't set up with any of the dependencies from the main app. I had a go at getting this docker image to match the main app but failed before. I can try to do this again as a follow up PR if that works? Or if someone else wants to have a go at getting the utils available here rather than me flailing around that would be appreciated 😄

jacksonj04: Nah, we can fix it later.

jacksonj04 (approved):

PR #1610, opened 2024-11-08
Fcl 190 add facet year to date warning

dragon-dxw:
.

## Changes in this PR:

* Tell the user what year our earliest document matching their search query is.
* Fix an off-by-one error which meant the current year could never be a year facet
* Make an assertion helper raise a useful error message

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-190

### After

image

### DIFF
--- a/judgments/utils/search_utils.py
+++ b/judgments/utils/search_utils.py
@@ -16,7 +16,7 @@ def _valid_years():
     Generate a list of valid years as strings.
     """
     today = datetime.date.today()
-    valid_years = range(2003, today.year)
+    valid_years = range(2003, today.year + 1)
     return [f"{year}" for year in valid_years]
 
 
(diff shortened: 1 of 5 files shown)


judgments/utils/search_utils.py
```diff
@@ -16,7 +16,7 @@ def _valid_years():
     Generate a list of valid years as strings.
     """
     today = datetime.date.today()
-    valid_years = range(2003, today.year)
+    valid_years = range(2003, today.year + 1)
```
jacksonj04, line 19: Oh `range`, why don't you do what anyone expects? 😄

jacksonj04 (approved):
One comment which I think a rebase solves, otherwise 🚢

judgments/utils/search_utils.py
dragon-dxw: jokingly: we should add a linter rule that `range(x)` is invalid, it should either be `range(x+1)` or `range(x+0)` if you really mean it.

PR #1620, opened 2024-11-14
Remove final references to document.content_as_html

jacksonj04:
This will further remove lengthy MarkLogic-side XSLT queries in displaying documents.


judgments/views/detail/generated_pdf.py
```diff
@@ -27,13 +24,13 @@ class PdfDetailView(WeasyTemplateResponseMixin, TemplateView):
     def get_context_data(self, **kwargs) -> dict[str, Any]:
         context = super().get_context_data(**kwargs)
 
-        document_uri = DocumentURIString(kwargs.get("document_uri", ""))
+        document_uri = DocumentURIString(kwargs["document_uri"])
```
jlhdxw, line 27: Do we want this to raise a KeyError or should we throw something specific that tells us we're missing the document_uri?

jacksonj04: There should be no real way to get to this function without `document_uri` being present in the `kwargs`, and if we do then it's an application error so should raise an exception. I'm happy with a `KeyError`.

jlhdxw (approved):

PR #1621, opened 2024-11-14
Fcl 410/fix links without text

jlhdxw:
## Changes in this PR:

Adds relevant aria-labels to links where links don't have text as direct children.

https://national-archives.atlassian.net/browse/FCL-410


jacksonj04 (approved):
There's a lot of repetition here I think we could clean up, but that's a later problem.

PR #1623, opened 2024-11-14
[FCL-350] Update alt label on logo

Donna-H:
## Changes in this PR:
update alt label on logo
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-350
### Before
before

### After
after


dragon-dxw (approved):
Are we still using both these includes?

Donna-H:
> Are we still using both these includes?

@dragon-dxw thanks deleted old include and updated logo name include from beta

ds_judgements_public_ui/templates/includes/logo.html
```diff
@@ -1,12 +1,9 @@
...
-  <a href="https://www.nationalarchives.gov.uk">
-    <img class="page-header__logo {{ link_class }}"
-         src="{% static 'images/tna_logo.svg' %}"
-         alt="The National Archives"
-         height="34"
-         width="250" />
-  </a>
-</div>
+<a href="https://www.nationalarchives.gov.uk" class="govuk-header__logo {{ link_class }}" aria-label="The National Archives home page">
+  <svg xml:space="preserve" class="page-header__tna-logo" style="enable-background:new 0 0 160 160" viewBox="0 0 160 160" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
jacksonj04, line 2: Can we swap this to loading an embedded file, instead of inlining the SVG? We use this on every single page, so we may as well let browsers cache it and save the bytes!

jacksonj04 (changes requested):

ds_judgements_public_ui/templates/includes/logo.html
Donna-H: @jacksonj04 done and updated also also added to this PR from the header homepage refinement card fcl-449 to update background logo to #121212 as @Terry-Price specified in the list od requirememts - so its now removed from fcl-449 card

jacksonj04 (commented):
My only lingering uncertainty here is that we have some classes etc still in the SVG which could probably be removed.

PR #1678, opened 2024-12-10
FCL-506 | override TNA default colours

jlhdxw:
## Changes in this PR:

Override the TNA colours with our own. We will need to move this to the FCL frontend repo once we're happy.

https://national-archives.atlassian.net/browse/FCL-506


Terry-Price (approved):
thanks James, looks good to me.

PR #1688, opened 2024-12-16
[FCL-446] clear out unused CSS and switch on output compression

jlhdxw:
## Changes in this PR:

Change it so our pages have **mostly** only the CSS they need. We will go from being 90%+ css unused to 60-70% css unused.

But we also go from a CSS file of 269kb to 41kb. So while there is still a high percentage, it's low in comparison.

https://national-archives.atlassian.net/browse/FCL-446


package.json
```diff
@@ -7,9 +7,9 @@
     "doc": "docs"
   },
   "scripts": {
-    "start-sass": "npx sass --watch  --load-path=node_modules ds_judgements_public_ui/sass:ds_judgements_public_ui/static/css",
+    "start-sass": "npx sass --watch  --load-path=node_modules ds_judgements_public_ui/sass:ds_judgements_public_ui/static/css --style=compressed",
```
jacksonj04, line 10: Is it worth leaving this uncompressed for dev?

jacksonj04 (approved):
I can see us re-unifying a bunch of these in future where pages have specific styles for no good reason, but this is _awesome_.

One quick comment about compressing CSS in dev, but no strong feelings here.

PR #1691, opened 2024-12-18
[FCL-166, FCL-164] Better encapsulation of JS

jlhdxw:
## Changes in this PR:

This makes it so we only load relevant JS on relevant pages. Also includes javascript unit tests for our more complicated javascript.

Document navigation links could do with a test as well but it's got quite a lot going on.

https://national-archives.atlassian.net/browse/FCL-162
https://national-archives.atlassian.net/browse/FCL-164


jacksonj04 (approved):
LGTM - I've added in an automation for the tests as well.

PR #1697, opened 2024-12-19
PR2 remove find courts browser add button

Donna-H:
## Changes in this PR:
remove find courts browser add button
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-450
### Before
before

### After
after-pic

[after.webm](https://github.com/user-attachments/assets/f0eec78a-bb71-400a-9add-cd9e68232505)

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_footer.scss
+++ b/ds_judgements_public_ui/sass/includes/_footer.scss
@@ -1,5 +1,7 @@
 .pre-footer {
-  text-align: center;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
   background-color: colour-var("accent-background-light");
 
   h2 {
@@ -80,18 +82,5 @@
 
   @media (min-width: $grid-breakpoint-medium) {
     font-size: $typography-md-text-size;
-  }
-}
-
-.more-info-footer {
-  text-align: center;
-  background-color: brand-colour("white");
-
-  &__cta-button {
-    @include call-to-action-button-secondary;
-
-    &:visited {
-      color: brand-colour("navy");
-    }
   }
 }
--- a/ds_judgements_public_ui/sass/includes/_footer.scss
+++ b/ds_judgements_public_ui/sass/includes/_footer.scss
@@ -82,3 +82,16 @@
     font-size: $typography-md-text-size;
   }
 }
+
+.more-info-footer {
+  text-align: center;
+  background-color: brand-colour("white");
+
+  &__cta-button {
+    @include call-to-action-button-secondary;
+
+    &:visited {
+      color: brand-colour("navy");
+    }
+  }
+}
--- a/ds_judgements_public_ui/templates/includes/recent_judgments.html
+++ b/ds_judgements_public_ui/templates/includes/recent_judgments.html
@@ -2,7 +2,7 @@
   <div class="recent-judgments__container">
     <h1 id="recently-published-judgments" class="recent-judgments__header">Recently published judgments</h1>
     {% include "includes/judgment_listing.html" with judgments=recent_judgments %}
-    <section class="more-info-footer">
+    <section class="more-info-footer" aria-labelledby="more-info-footer">
       <div class="container py-2">
         <a role="button"
            aria-label="See all recent judgments"
--- a/ds_judgements_public_ui/templates/includes/recent_judgments.html
+++ b/ds_judgements_public_ui/templates/includes/recent_judgments.html
@@ -2,8 +2,15 @@
   <div class="recent-judgments__container">
     <h1 id="recently-published-judgments" class="recent-judgments__header">Recently published judgments</h1>
     {% include "includes/judgment_listing.html" with judgments=recent_judgments %}
-    <p class="judgment-listing__more-info">
-      <a href="{% url 'search' %}">See all recent judgments</a>
-    </p>
+    <section class="more-info-footer" aria-labelledby="more-info-footer-heading">
+      <div class="container py-2">
+        <a role="button"
+           aria-label="See all recent judgments"
+           draggable="false"
+           rel="noreferrer noopener"
+           class="more-info-footer__cta-button"
+           href="{% url 'search' %}">See all recent judgments</a>
+      </div>
+    </section>
   </div>
 </div>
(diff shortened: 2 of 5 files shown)


ds_judgements_public_ui/templates/includes/recent_judgments.html
```diff
@@ -2,8 +2,15 @@
   <div class="recent-judgments__container">
     <h1 id="recently-published-judgments" class="recent-judgments__header">Recently published judgments</h1>
     {% include "includes/judgment_listing.html" with judgments=recent_judgments %}
-    <p class="judgment-listing__more-info">
-      <a href="{% url 'search' %}">See all recent judgments</a>
-    </p>
+    <section class="more-info-footer" aria-labelledby="more-info-footer-heading">
```
jacksonj04, line 5: We don't have a `more-info-footer-heading` element for this `aria-labelledby` to make sense.

ds_judgements_public_ui/templates/includes/recent_judgments.html
```diff
@@ -2,8 +2,15 @@
...
+      <div class="container py-2">
+        <a role="button"
+           aria-label="See all recent judgments"
```
jacksonj04, line 8: I don't think this `aria-label` is necessary if it's exactly the same as the button's label (though @jlhdxw may have other thoughts!)

ds_judgements_public_ui/sass/includes/_footer.scss
```diff
@@ -82,3 +82,16 @@
...
+
+.more-info-footer {
+  text-align: center;
+  background-color: brand-colour("white");
+
+  &__cta-button {
+    @include call-to-action-button-secondary;
+
+    &:visited {
+      color: brand-colour("navy");
```
jacksonj04, line 94: Nice fix!

ds_judgements_public_ui/sass/includes/_footer.scss
```diff
@@ -82,3 +82,16 @@
     font-size: $typography-md-text-size;
   }
 }
+
+.more-info-footer {
+  text-align: center;
+  background-color: brand-colour("white");
```
jacksonj04, line 88: Does this element really need a background colour? Isn't it already white?

jacksonj04 (dismissed):

ds_judgements_public_ui/templates/includes/recent_judgments.html
Donna-H: removed from section - there is aria in button below

ds_judgements_public_ui/sass/includes/_footer.scss
```diff
@@ -82,3 +82,16 @@
     font-size: $typography-md-text-size;
   }
 }
+
+.more-info-footer {
```
jlhdxw, line 86: As this isn't within the footer and instead inside the `recent_judgments` component, can we move this CSS to that component and prefix with the appropriate class i.e. `recent-judgments`

ds_judgements_public_ui/templates/includes/recent_judgments.html
```diff
@@ -2,8 +2,15 @@
   <div class="recent-judgments__container">
     <h1 id="recently-published-judgments" class="recent-judgments__header">Recently published judgments</h1>
     {% include "includes/judgment_listing.html" with judgments=recent_judgments %}
-    <p class="judgment-listing__more-info">
-      <a href="{% url 'search' %}">See all recent judgments</a>
-    </p>
+    <section class="more-info-footer">
+      <div class="container py-2">
```
jlhdxw, line 6: Can we avoid using these helper classes (i.e. py-2 and container) and opt for styling in your component CSS. These styles should be applied to `more-info-footer`.

ds_judgements_public_ui/sass/includes/_footer.scss
```diff
@@ -82,3 +82,16 @@
...
+  text-align: center;
```
jlhdxw, line 87: Rather than rely on the contents of this being display inline, can we style this with flex so the contents is always styled in the middle, i.e.

```
display: flex;
flex-direction: column;
justify-content: center;
```

ds_judgements_public_ui/templates/includes/recent_judgments.html
jlhdxw: Nope we shouldn't have an aria label here. wcag specifically say not to add aria-labels/titles/alt tags if it doesn't help.

ds_judgements_public_ui/sass/includes/_footer.scss
Donna-H: @jlhdxw the prefooter below with the button has the exact same so I did the same display inline

Terry-Price (changes requested):
Hi @Donna-H Can the words change to 'View...' rather than 'See...'

ds_judgements_public_ui/sass/includes/_footer.scss
Donna-H: @jlhdxw updated both pre-footer which previously had inline  and more-info-footer with flex now

Donna-H:
> Hi @Donna-H Can the words change to 'View...' rather than 'See...'

@Terry-Price updated with view

ds_judgements_public_ui/templates/includes/recent_judgments.html
Donna-H: @jacksonj04 is this a standard class introduced seen it also added in code by you previously so we can use this container py-2" as Tim also did ?

ds_judgements_public_ui/sass/includes/_footer.scss
Donna-H: updated

ds_judgements_public_ui/templates/includes/recent_judgments.html
Donna-H: removed

ds_judgements_public_ui/templates/includes/recent_judgments.html
Donna-H: If this is fine to keep in as I have to get to nect PR 3 then sort my IT issues for rest of this afternoon @jacksonj04 @jlhdxw

Terry-Price (approved):
thanks @Donna-H seen the new text on your screen just now, approved..

PR #1701, opened 2024-12-20
Fcl 153/refactor buttons

jlhdxw:
## Changes in this PR:

Have a single source of truth for our buttons. There may be more that I haven't found.

I've aimed to have `button-primary` or `button-secondary` classes on all buttons so these can be tracked in GA nicely, but also it's super easy to see how these are styled and any overrides when looking at them in the browser.

https://national-archives.atlassian.net/browse/FCL-153


jlhdxw:
Superseded by https://github.com/nationalarchives/ds-caselaw-public-ui/pull/1714

PR #1715, opened 2025-01-06
Fix for some incorrect noindex behaviours

jacksonj04:
We are incorrectly setting noindex headers in some places we shouldn't be; update behaviours accordingly.

- HTTP redirects are now indexable, so search engines can properly resolve the destinations
- The list of courts page which is now accessible from homepage blocks is indexable (and has working breadcrumbs)


config/middleware.py
```diff
@@ -13,9 +14,16 @@ def __call__(self, request):
         # Code to be executed for each request before
         # the view (and later middleware) are called.
         response = self.get_response(request)
+
+        # If the response is a redirect, don't add the headers
```
dragon-dxw, line 18: I'd be tempted to add the word "robots" to this comment.

config/views/courts.py
```diff
@@ -15,12 +15,16 @@ class CourtsTribunalsListView(TemplateViewWithContext):
...
     template_name = "pages/courts_and_tribunals.html"
     page_title = "Judgments and decisions by court or tribunal"
+    page_allow_index = True
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["courts"] = courts.get_grouped_selectable_courts()
         context["tribunals"] = courts.get_grouped_selectable_tribunals()
         context["feedback_survey_type"] = "courts_and_tribunals"
+        context["breadcrumbs"] = [
```
dragon-dxw, line 25: This feels like it's doing something unrelated, but seems sensible.

config/views/courts.py
```diff
@@ -15,12 +15,16 @@ class CourtsTribunalsListView(TemplateViewWithContext):
 
     template_name = "pages/courts_and_tribunals.html"
     page_title = "Judgments and decisions by court or tribunal"
+    page_allow_index = True
```
dragon-dxw, line 18: How does this appear in the context as used on line 23 of the middleware? I don't think I get this.

jacksonj04: This is handled in the  `TemplateViewWithContext` superclass.

dragon-dxw (approved):

PR #1721, opened 2025-01-07
Add Terry to CODEOWNERS for html, css

dragon-dxw:
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-537


CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+*.css @Terry-Price
```
jlhdxw, line 1: I think this'll need to be scss. Also might be worth adding js

PR #1743, opened 2025-01-14
decrease ncn font size

Donna-H:
## Changes in this PR:
decrease ncn font size from 2rem to 1.8rem
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-468
### Before
before

### After
after

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_variables.scss
+++ b/ds_judgements_public_ui/sass/includes/_variables.scss
@@ -16,6 +16,7 @@ $typography-xs-text-size: 0.7rem;
 $typography-sm-text-size: 0.9rem;
 $typography-md-text-size: 1.1rem;
 $typography-lg-text-size: 1.3rem;
+$typography-2lg-text-size: 1.8rem;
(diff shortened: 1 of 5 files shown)


ds_judgements_public_ui/sass/includes/_variables.scss
```diff
@@ -16,6 +16,7 @@ $typography-xs-text-size: 0.7rem;
 $typography-sm-text-size: 0.9rem;
 $typography-md-text-size: 1.1rem;
 $typography-lg-text-size: 1.3rem;
+$typography-2lg-text-size: 1.8rem;
```
jlhdxw, line 19: We should actually be able to delete all these typography variables now as they're included in the main front-end repo.

Donna-H: Have been deleted will just await @Terry-Price to review

Terry-Price (approved):
Thanks @Donna-H This looks really good now visually.
I like the increased size contrast now.

PR #1748, opened 2025-01-15
Remove 'recent' from 'View all recent judgments' button

dragon-dxw:
As discussed in standup.


jacksonj04 (approved):

Terry-Price (commented):
thanks @dragon-dxw Awesomely quick

PR #1764, opened 2025-01-20
Fcl 586/judgments by court rework

jlhdxw:
## Changes in this PR:

This includes some refactors for the content layout to reduce duplication and allow more flexibility.

https://national-archives.atlassian.net/browse/FCL-597


ds_judgements_public_ui/sass/includes/_info_card.scss
```diff
@@ -1,24 +1,15 @@
...
-
-  @media (min-width: $grid-breakpoint-medium) {
-    flex-direction: row;
-  }
-}
-
-.info-card {
+@mixin info-card {
   display: flex;
+  flex: 33%;
```
jacksonj04, line 3: Does this need to be pinned to three-wide, or can we let `flex` solve it?

jacksonj04 (approved):

ds_judgements_public_ui/sass/includes/_info_card.scss
jlhdxw: Flex should be able to figure it out. The flex basis with 33% means the starting point for the size of the element is 33% of the parent, and then it'll adjust as necessary to fit/shrink depending on the other flex settings.

PR #1768, opened 2025-01-22
Fcl 586/judgments by court updated page

jlhdxw:
## Changes in this PR:

Visual updates to the courts and tribunals page

https://national-archives.atlassian.net/browse/FCL-586

### Before

Desktop:

staging caselaw nationalarchives gov uk_courts-and-tribunals (1)

Mobile:

staging caselaw nationalarchives gov uk_courts-and-tribunals

### After

Desktop:

localhost_3000_courts-and-tribunals

Mobile:

localhost_3000_courts-and-tribunals (1)

### DIFF
--- a/ds_judgements_public_ui/templates/includes/courts_and_tribunals_court.html
+++ b/ds_judgements_public_ui/templates/includes/courts_and_tribunals_court.html
@@ -5,16 +5,32 @@
   {% else %}
     <h2 id="{{ court.name|slugify }}" class="courts-and-tribunals__header">{{ court.grouped_name }}</h2>
   {% endif %}
-  <details class="govuk-details--compact">
-    <summary class="govuk-details__summary">
-      About the {{ court.grouped_name }}
-    </summary>
-    <div class="govuk-details__text">
-      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris nulla purus, imperdiet eget malesuada ac, mattis ac velit. Sed at urna eget nunc ultrices iaculis. Maecenas ex quam, mollis et egestas vel, sagittis non dui. Proin quis aliquam nisl. Proin vel varius nisi. Sed quis mauris nec sem rutrum hendrerit. Sed molestie pellentesque iaculis.</p>
-    </div>
-  </details>
+  {% if court.description_text_as_html %}
+    <details class="govuk-details--compact">
+      <summary class="govuk-details__summary">
+        About the {{ court.grouped_name }}
+      </summary>
+      <div class="govuk-details__text">
+        {% autoescape off %}
+          {{ court.description_text_as_html }}
+        {% endautoescape %}
+      </div>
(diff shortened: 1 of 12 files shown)


ds_judgements_public_ui/templates/includes/courts_and_tribunals_court.html
```diff
@@ -1,8 +1,36 @@
...
   <p>
-    {{ court | get_court_judgments_count }} documents from
-    {{ court.canonical_param | get_court_date_range }}
+    We currently hold <a href="{% url "search" %}?{{ type }}={{ court.canonical_param }}">{{ court | get_court_judgments_count }} documents from
+      {{ court.canonical_param | get_court_date_range }}</a>
   </p>
+  {% if court.historic_documents_support_text_as_html %}
+    <details class="govuk-details--compact">
+      <summary class="govuk-details__summary">
+        Looking for older
```
jacksonj04, line 27: This seems a bit abrupt; is it placeholder language?

jacksonj04 (approved):
One quick text change, otherwise 🚀

ds_judgements_public_ui/templates/includes/courts_and_tribunals_court.html
jlhdxw: Ah looks like the design has different content to the doc here - will update.

jacksonj04 (changes requested):
Just rebased this on `main` to get the new content, and a couple of things:

- [x] Could we lose the `WIP` commit, maybe squash it or reword?
- [x] The "Tribunals" link in the sidebox has some strange behaviour; it looks like it places the focus on the Tribunals link under the "Contents" heading in the main content, rather than actually skipping to the Tribunals section. Can we just lose "Contents" in the body since that's replicated in the sidebar? If not, the fragment link behaviour needs fixing.
- [x] The mockup has some subtle grey lines between courts, which really helps break the page up. I think they're missing in this at the moment.
- [ ] It doesn't currently happen on prod, but staging has some courts without documents. This could hypothetically happen on prod when we're adding new courts, is it worth adding a template condition for "We do not currently host any documents from this court" or similar?

Terry-Price (approved):
Visually looks good James

jlhdxw:
@jacksonj04 I've done the rebase to change the name of the commit and added the border.

WRT the behaviour of clicking the sidebar links, I couldn't replicate this. It works as expected for me - there is some javascript that handles this - did you build the assets when testing?

For the "We do not currently host any documents from this court" text, @Terry-Price do you have any thoughts on this? Currently I think it would say "We currently hold 0 documents from 2024 to 2025" which doesn't make much sense.

jacksonj04:
@jlhdxw it looks like removing the "contents" from the body has resolved the anchor glitch anyway!

jacksonj04 (approved):

PR #1773, opened 2025-01-24
fix: don't allow users to select clipboard tooltip

jlhdxw:
## Changes in this PR:

Don't allow users to select clipboard tooltip


Donna-H (approved):

Terry-Price (commented):
The font size is set at 14px, I thought we were using rems for font sizing?

PR #1791, opened 2025-01-30
fix bug previous button

Donna-H:
Fix padding on previous button in licence application

## Changes in this PR:
fix bug licence previous button
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-638
### Before
before

### After
1-after


jacksonj04 (approved):
🚀

PR #1794, opened 2025-01-30
Make older judgments visible; separate minimum sensible and allowed dates

dragon-dxw:
Pre-2001 judgments were not showing up properly on staging -- we probably could have changed the default oldest date, but realistically separating out the minimum valid (some time after 1000) and sensible (oldest judgment; defaulting to 2003-ish) more carefully was needed.

https://national-archives.atlassian.net/browse/FCL-635


judgments/views/advanced_search.py
```diff
@@ -57,14 +58,18 @@ def _do_dates_require_warnings(
...
     """
-    min_year = get_minimum_valid_year()
+    min_year = get_minimum_warning_year()
 
     if not iso_date:
         return False, None
     from_date = date.fromisoformat(iso_date)
+    # If the date is 1085, then that's the default, the user hasn't typed in
+    # a non-sensical answer, don't warn about it.
+    if from_date.year == settings.MINIMUM_ALLOWED_YEAR:
```
jacksonj04, line 71: This makes me a bit twitchy because if a user does type "1085" then we should still warn them, but solving it is a bit of a faff.

dragon-dxw: Yes; but it's such an edge case that I'm not really worried about it. And this fixes a much worse bug.

dragon-dxw: I could check if the date is 1085-01-01; otherwise I'd be have to be checking if a date was provided at all

jacksonj04 (approved):

PR #1795, opened 2025-01-31
Remove references to registering for research

dragon-dxw:
We discussed at standup that we're currently not looking for research volunteers.


ds_judgements_public_ui/templates/layouts/base.html
```diff
@@ -64,7 +64,6 @@
     {% endblock postcontent %}
 
     {% block feedback_cta %}
-      {% include "includes/help_improve_this_service.html" %}
```
jlhdxw, line 67: Shall we remove this include as well? I don't think it's used anywhere else (maybe we can remove this whole block too?)

dragon-dxw: Mostly I was leaving in the bits that let us reinstate it relatively quickly, but happy to!

PR #1797, opened 2025-01-31
FCL-637 | only return true for matching ncn if there is an ncn

jlhdxw:
## Changes in this PR:

Only match NCN when the result has an NCN

https://national-archives.atlassian.net/browse/FCL-637

### DIFF
--- a/judgments/tests/test_utils.py
+++ b/judgments/tests/test_utils.py
@@ -99,17 +99,17 @@ class TestUtils(unittest.TestCase):
     def test_is_exact_ncn_match_empty_ncn(self):
         document = JudgmentFactory.build(neutral_citation=None)
 
-        assert not is_exact_ncn_match(document, "foo")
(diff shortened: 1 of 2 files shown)


judgments/tests/test_utils.py
```diff
@@ -95,6 +96,21 @@ def test_press_summary_list_breadcrumbs(self):
             },
         ]
 
+    def test_is_exact_ncn_match_empty_ncn(self):
+        document = JudgmentFactory.build(neutral_citation=None)
+
+        assert not is_exact_ncn_match(document, "foo")
```
jacksonj04, line 102: Really picky, but could this be:
```suggestion
        assert is_exact_ncn_match(document, "foo") is False
```

and similar for the others? That way if it ever suddenly change the signature and it starts returning things like `[]` tests will explode.

jacksonj04 (approved):
LGTM 🚢

PR #1806, opened 2025-02-04
FCL-623 | fix firefox link focus issue

jlhdxw:
## Changes in this PR:

This is how you fix outline issues on firefox - chrome knows to group an inline element into 1 node and outline the lot. Firefox doesn't do that. If you explicitly give it display block, firefox sees the element as a box shape.

https://national-archives.atlassian.net/browse/FCL-623

### Before

image

### After

image


dragon-dxw (approved):

Donna-H (approved):
thanks @jlhdxw I did not realise it was in this file in list style to apply it

PR #1808, opened 2025-02-05
add survey button new link and text in pre footer

Donna-H:
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-634
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-587
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-562
### Before
before

### After
after

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_buttons.scss
+++ b/ds_judgements_public_ui/sass/includes/_buttons.scss
@@ -83,6 +83,16 @@
   @include button-secondary;
 }
 
+@mixin survey-button {
+  @include button-secondary;
+
+  padding: $space-4 $space-8;
+}
+
+.survey-button {
+  @include survey-button;
+}
+
 .button-small {
   @include button-small;
 }
--- a/ds_judgements_public_ui/sass/includes/_buttons.scss
+++ b/ds_judgements_public_ui/sass/includes/_buttons.scss
@@ -83,16 +83,6 @@
   @include button-secondary;
 }
 
-@mixin survey-button {
-  @include button-secondary;
-
-  padding: $space-4 $space-8;
-}
-
-.survey-button {
-  @include survey-button;
-}
-
 .button-small {
   @include button-small;
 }
--- a/ds_judgements_public_ui/sass/includes/_footer.scss
+++ b/ds_judgements_public_ui/sass/includes/_footer.scss
@@ -10,6 +10,13 @@
     font-size: $typography-lg-text-size;
     font-weight: $typography-normal-font-weight;
   }
+
+  &__container {
+    @include container;
+
+    margin-top: 0;
+    padding: $space-1 0 $space-8;
+  }
 }
 
 .judgments-footer {
--- a/ds_judgements_public_ui/sass/includes/_footer.scss
+++ b/ds_judgements_public_ui/sass/includes/_footer.scss
@@ -15,7 +15,7 @@
     @include container;
 
     margin-top: 0;
-    padding: $space-1 0 $space-8;
+    padding: $space-1 0 $space-8 $space-8;
   }
 }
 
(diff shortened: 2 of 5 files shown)


Terry-Price (changes requested):
hi @Donna-H 
Thanks for this, the only thing I think still needs attention is the central alignment of the text and on the survey button. When I look at the alignmnet, it;s still not quite right.
See No.2 point in the  card #587.

If this is causing to too much pain/time, put this live and remake a small card to align it later for yourself.

jacksonj04 (commented):
@jacksonj04  its the same button added back from PR where Dragon removed it ?  Nothing has been changed only text request in the Jira ticket
The old button which this is replacing carried a lot of context about where the user was when they clicked it; do we want to figure out how to include this as well? @Terry-Price  @jlhdxw maybe weigh in on this and let me know what else needs to be added as I just added back the old button and took out the word "our" as per the ticket request

Donna-H:
> hi @Donna-H Thanks for this, the only thing I think still needs attention is the central alignment of the text and on the survey button. When I look at the alignmnet, it;s still not quite right. See No.2 point in the card #587.
> 
> If this is causing to too much pain/time, put this live and remake a small card to align it later for yours

Donna-H:
> hi @Donna-H Thanks for this, the only thing I think still needs attention is the central alignment of the text and on the survey button. When I look at the alignmnet, it;s still not quite right. See No.2 point in the card #587.
> 
> If this is causing to too much pain/time, put this live and remake a small card to align it later for yourself.

@Terry-Price as confirmed with you via slack this has now been changed and the images added before to show alignment update

Terry-Price (approved):
Visually looks great- thanks @Donna-H

ds_judgements_public_ui/sass/includes/_buttons.scss
```diff
@@ -83,6 +83,26 @@
...
+  &:visited {
+    color: brand-colour("navy");
+    background-color: transparent;
+  }
+}
+
+.survey-button {
+  @include survey-button;
+}
+
```
jlhdxw, line 105: We can remove this and just use `.button-secondary`

ds_judgements_public_ui/sass/includes/_footer.scss
```diff
@@ -7,6 +7,8 @@
   h2 {
     margin-bottom: 0;
     padding-top: 0;
+    padding-left: $space-1;
```
jlhdxw, line 10: This doesn't look right when added - we shouldn't need this at all.

ds_judgements_public_ui/templates/includes/help_improve_this_service.html
```diff
@@ -0,0 +1,12 @@
+<section class="pre-footer" aria-labelledby="pre-footer-heading">
+  <div class="container py-2">
```
jlhdxw, line 2: The atom classes are deprecated - can we add the relevant styling to the `pre-footer` class?

ds_judgements_public_ui/templates/includes/help_improve_this_service.html
```diff
@@ -0,0 +1,12 @@
+<section class="pre-footer" aria-labelledby="pre-footer-heading">
```
jlhdxw, line 1: I don't think we should need this aria-labelledby - the browser should know the h2 inside this section is the appropriate header. If our axe accessibility check says otherwise I'd be interested in seeing why that is.

ds_judgements_public_ui/sass/includes/_buttons.scss
Donna-H: removed and updated

ds_judgements_public_ui/sass/includes/_footer.scss
Donna-H: removed.

ds_judgements_public_ui/templates/includes/help_improve_this_service.html
Donna-H: removed.

jlhdxw (approved):

PR #1810, opened 2025-02-06
[FCL-664] Swap out document resolution engine for new and improved identifier framework

jacksonj04:
This moves how the PUI finds documents away from the old method of "there must be an exact URI match", and to a system whereby it resolves a pre-compiled URL against a list of URIs within MarkLogic.


dragon-dxw (approved):

dragon-dxw:
There's additional work to do before this can go live (search for starters). We might need to have a think about our deployment strategy since we might wind up deploying this if we merge it.

PR #1825, opened 2025-02-13
Block LLM AI bots via robots.txt

dragon-dxw:
We discussed in the stride review that we were getting crawled a lot more; given that AI use of the data requires a licence, blocking the bots from even crawling makes sense. 

List taken from https://www.cyberciti.biz/web-developer/block-openai-bard-bing-ai-crawler-bots-using-robots-txt-file/


jacksonj04:
This feels like it needs a bigger policy decision than we can make ourselves.

jacksonj04 (approved):
LGTM, but we're going to need some kind of higher-level process for keeping this up to date.

PR #1826, opened 2025-02-13
Send signals to avoid AI crawled use of data.

dragon-dxw:
## Changes in this PR:

Add `noai` meta robots tag
Add trust.txt with similar behaviour
~~Add rel=license to footer OJL links~~

https://dxw.slack.com/archives/C02TP2L2Z0F/p1739450499538649?thread_ts=1739449217.899369&cid=C02TP2L2Z0F

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -191,6 +191,14 @@ urlpatterns = [
         "robots.txt",
         TemplateView.as_view(template_name="robots.txt", content_type="text/plain"),
     ),
+    path(
+        "trust.txt",
+        TemplateView.as_view(template_name="trust.txt", content_type="text/plain"),
+    ),
+    path(
+        ".well-known/trust.txt",
+        RedirectView.as_view(url="/trust.txt"),
+    ),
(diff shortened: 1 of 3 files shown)


config/urls.py
```diff
@@ -191,6 +191,14 @@
         "robots.txt",
         TemplateView.as_view(template_name="robots.txt", content_type="text/plain"),
     ),
+    path(
+        ".well-known/trust.txt",
+        TemplateView.as_view(template_name="trust.txt", content_type="text/plain"),
+    ),
+    path(
+        "trust.txt",
+        RedirectView.as_view(url="/.well-known/trust.txt"),
```
jacksonj04, line 200: Looking at the [spec](https://journallist.net/reference-document-for-trust-txt-specifications) I think this is actually backwards - we should redirect from `/.well-known/trust.txt` to `/trust.txt`.

jacksonj04 (changes requested):

jacksonj04 (approved):

PR #1839, opened 2025-02-20
FCL-693 add survey banner to homepage and judgments

Donna-H:
## Changes in this PR:
Add survey banner to homepage and judgments
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-693
### Before Homepage Desktop
[before-homepage-desktop.webm](https://github.com/user-attachments/assets/e3ba25a2-6f42-46d0-b141-ff31ee0183d8)

### After Homepage Desktop
[after-desktop-homepage.webm](https://github.com/user-attachments/assets/a9797131-7190-4317-aca4-99045c3a3adc)

### Before Homepage Mobile
[before-home-mobile.webm](https://github.com/user-attachments/assets/b32973d5-71b2-4fb3-8002-6824f3eaa9f5)

### After Homepage Mobile
[after-home-mob.webm](https://github.com/user-attachments/assets/0a5f934c-5229-41dc-aa6c-f7aace7b80c0)

### Before Judgment Desktop
[before-desktop-judgment.webm](https://github.com/user-attachments/assets/9bd10c9c-7bd3-4708-8f2c-0684c9ed67c6)

### After Judgment Desktop
[after-desktop-judgment.webm](https://github.com/user-attachments/assets/b6e096a4-1c03-4a43-8e37-3cca06e89288)

### Before Judgment Mobile
[before-mobile-judgment.webm](https://github.com/user-attachments/assets/f630592c-427e-4156-80da-83bf620b99ea)

### After Judgment Mobile

[after-mobile-judgment.webm](https://github.com/user-attachments/assets/6925bcad-3aac-4fc9-a413-307116564238)

### DIFF
--- a/ds_judgements_public_ui/templates/judgment/detail.html
+++ b/ds_judgements_public_ui/templates/judgment/detail.html
@@ -12,6 +12,7 @@
   Find Case Law
 {% endblock title %}
 {% block precontent %}
+  {% include "includes/survey_banner.html" %}
   {% include "includes/judgment_text_toolbar.html" %}
 {% endblock precontent %}
 {% block body_class %}document-detail{% endblock body_class %}
--- a/ds_judgements_public_ui/sass/includes/_survey_banner.scss
+++ b/ds_judgements_public_ui/sass/includes/_survey_banner.scss
@@ -1,0 +1,29 @@
+.survey-banner {
+  font-family: $font-roboto;
+  font-size: $typography-lg-text-size;
+  background-color: $color-information;
+
+  &__container {
+    @include container;
+  }
+
+  &__notice {
+    padding: $space-3;
+    text-align: center;
+
+    a {
+      font-weight: $typography-bold-font-weight;
+      color: colour-var("link");
+
+      &:focus {
+        color: colour-var("link");
+        background-color: none;
+      }
+    }
+  }
+
+  &__message {
+    display: inline;
+    margin: 0;
+  }
+}
--- a/ds_judgements_public_ui/sass/main.scss
+++ b/ds_judgements_public_ui/sass/main.scss
@@ -130,3 +130,6 @@ $govuk-focus-colour: colour-var("focus-outline");
 // Dev
 @import "includes/environment_banner";
 @import "includes/style_guide";
+
+// Survey
+@import "includes/survey_banner";
--- a/ds_judgements_public_ui/templates/includes/survey_banner.html
+++ b/ds_judgements_public_ui/templates/includes/survey_banner.html
@@ -1,0 +1,7 @@
+<div class="survey-banner">
+  <div class="survey-banner__container">
+    <div class="survey-banner__notice">
+      <a target="_blank" href="{{ feedback_survey_link }}">Take part in research to improve the Find Case Law Service</a>
+    </div>
+  </div>
+</div>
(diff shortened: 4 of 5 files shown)


jacksonj04 (commented):
Two drive-by comments:

- Should "service" have a lowercase "s"?
- Can we feature flag this (the right thing in Waffle is called a [switch](https://waffle.readthedocs.io/en/stable/types/switch.html)) so we don't need a code push to turn it off again?

Donna-H:
> Two drive-by comments:
> 
> * Should "service" have a lowercase "s"?
> * Can we feature flag this (the right thing in Waffle is called a [switch](https://waffle.readthedocs.io/en/stable/types/switch.html)) so we don't need a code push to turn it off again?

@jacksonj04 https://jpd2z5.axshare.com/?id=f5enx0&g=14&p=survey-band-short-term_judgment-desktop__1 https://jpd2z5.axshare.com/?id=hvxrjq&g=14&p=survey-band-short-term_homepage-mobile__1 both have uppercase and he reviewed this all and I tweaked some  padding he requested so rest is as his designs so maybe Nicki can be asked in stand up if designs need changing.  OH yes I did say I would put flags in but realised as its include I could just delete the code easily when its no longer required.  I will read up about switch then as I have used flags but not as a switch I assume it deletes the code for us when its switched off that is why we are adding the flag? So then I do not need to go back in to code  delete the include when its not needed.

Donna-H:
> > Two drive-by comments:
> > 
> > * Should "service" have a lowercase "s"?
> > * Can we feature flag this (the right thing in Waffle is called a [switch](https://waffle.readthedocs.io/en/stable/types/switch.html)) so we don't need a code push to turn it off again?
> 
> @jacksonj04 https://jpd2z5.axshare.com/?id=f5enx0&g=14&p=survey-band-short-term_judgment-desktop__1 https://jpd2z5.axshare.com/?id=hvxrjq&g=14&p=survey-band-short-term_homepage-mobile__1 both have uppercase and he reviewed this all and I tweaked some padding he requested so rest is as his designs so maybe Nicki can be asked in stand up if designs need changing. OH yes I did say I would put flags in but realised as its include I could just delete the code easily when its no longer required. I will read up about switch then as I have used flags but not as a switch I assume it deletes the code for us when its switched off that is why we are adding the flag? So then I do not need to go back in to code delete the include when its not needed.

@jacksonj04 https://mozilla.github.io/addons-server/topics/development/waffle.html#why-switches-and-not-flags it mentions in clean up we remove from the code which I can easily remove the banner include from code. As I am reading up now on switch and flag differences, I decided not to initially use a flag as previously intended so please explain why it would be worth adding switch.  As I can easily remove the banner include when user research no longer require it there.

ds_judgements_public_ui/templates/judgment/detail.html
```diff
@@ -12,6 +12,7 @@
   Find Case Law
 {% endblock title %}
 {% block precontent %}
+  {% include "includes/survey_banner.html" %}
```
anthonyhashemi, line 15: I was going to ask if there is a higher level template that is shared between judgment/detail and pages.home that we could put the banner in so we could avoid including it twice, but looking at the screenshots, we have them in different parts of the page and so it would require some sort of refactor to the base template that probably isn't worth it/ may not make sense.

anthonyhashemi (approved):
Approving this assuming there is no good reason to put it behind Feature flag /switch (I currently agree with you in that we can just remove the banner include when we are done with the survey)

The only reason I could see for a flag/switch would be if we wanted to turn it on/off for a certain period/time but I can't think of a scenario for that

PR #1881, opened 2025-03-11
Try to preserve playwright test traces

dragon-dxw:
Playwright allows us to preserve the test traces so we can open them up in the [Playwright Trace Viewer](https://trace.playwright.dev/). This zips up the failing test traces, which is **not** a valid Trace file -- unzip it locally and drag one of the sub-zip files into the Trace Viewer to view.

Not 100% sure where the best place to see the artifacts is -- there is a link in the voluminous console log.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,13 +148,13 @@ It's because the editor UI project is still running, you'll need to reopen that 
 Now go back to the Public UI project and use the same command `fab stop`.
 Now you can restart the project up again with `fab run`.
 
-### 8. Other development tips
+## Other development tips
 
 For day to day development, running `fab run` should provide you with all you need.
 
 Other useful commands are:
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -219,7 +219,9 @@ def test_my_page(page: Page):
 
 If the page is accessible, the console won't output anything other than the usual test output. If there are accessibility issues, there will be output that explains what the issues are and also helpful links explaining how to fix them.
 
-#### Viewing code coverage
+####
```
jacksonj04, line 222: I think this has crept in.

jacksonj04 (changes requested):
Could we add a quick paragraph on the fact we capture output from e2e if you need it, and where to find the link?

jacksonj04 (approved):

PR #1886, opened 2025-03-13
[FCL-645] Update site footer

Donna-H:
## Changes in this PR:
Update the footer
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-645
### Before Desktop
before-desktop

### After Desktop
after-desktop

### Before Mobile
before-mobile

### After Mobile
after-mob

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_footer.scss
+++ b/ds_judgements_public_ui/sass/includes/_footer.scss
@@ -20,9 +20,13 @@
 }
 
 .judgments-footer {
-  font-size: $typography-sm-text-size;
+  font-size: $typography-md-text-size;
   color: colour-var("contrast-font-base");
   background-color: colour-var("contrast-background");
+
+  &__container {
+    @include container;
+  }
 
   &__links-container {
     @media (min-width: $grid-breakpoint-medium) {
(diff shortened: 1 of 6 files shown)


Terry-Price (approved):
Thanks @Donna-H - great work

ds_judgements_public_ui/sass/includes/_footer.scss
```diff
@@ -63,6 +120,98 @@
     }
   }
 
+  &__h3-link {
+    font-weight: bold;
+  }
+
+  &__about-links {
```
jacksonj04, line 127: My immediate thought here is that the columns in the footer probably should be - at least in terms of visual/layout - the same class. Needing a specific class for each one with hand-carved widths instead of letting `flex` handle it makes me think that something isn't quite right. This introduces a lot of duplication which shouldn't be necessary.

jacksonj04 (changes requested):
Some concerns about the amount of duplication in CSS which feels a bit smelly to me; we'll have a proper look once we've got the data stuff out of the way!

ds_judgements_public_ui/sass/includes/_footer.scss
Donna-H: @jacksonj04 the footer update is based on what we have in our current footer using flex and all using BEM footer class judgments-footer and had to be specific based on Terry design had to be specific @Terry-Price remember in our meeting that I  mentioned the duplication maybe raised see @jacksonj04 comment .

Donna-H: @jacksonj04 updated removed classes from the columns in the links-container. The other section with 'Legal' and 'other Websites' will require keeping classes because the list is styled to be horizontal then also have aligned it to fit design

ds_judgements_public_ui/templates/includes/footer.html
```diff
@@ -1,67 +1,96 @@
...
+               height="50"
+               width="32" />
+        </div>
+        <div>
+          All content is available under the <a class="site-footer__link"
+                                                href="{% url 'open_justice_licence' %}"
+                                                rel="license">Open Justice Licence</a>.
+        </div>
+        <br/>
+        For computational analysis, which is not allowed under the OJL <a href="{% url "transactional-licence-form" %}">apply for a license</a>.
```
jacksonj04, line 68: _Super_ nit-picky point for @Terry-Price here, should this be:

```suggestion
        For computational analysis, which is not allowed under the OJL, <a href="{% url "transactional-licence-form" %}">apply for a license</a>.
```

to add a missing comma and properly separate the subordinate clause?

jacksonj04 (approved):

jacksonj04:
This will probably need a rebase once #1892 has merged to `main` so that we get the Playwright update to address the e2e failure.

PR #1896, opened 2025-03-20
FCL 745 update start text searching judgments

Donna-H:
## Changes in this PR:
FCL 745 update start text in sticky header - searching judgments
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-745
### Before
before

### After
after


Terry-Price (approved):
all good - thanks @Donna-H

PR #1912, opened 2025-04-02
Homepage accessibility updates

Donna-H:
## Changes in this PR:
Add aria current to navigation links, remove aria labels on the homepage cards links, update image alt tags to empty

https://national-archives.atlassian.net/browse/FCL-746
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-761
### Before - Add aria current to navigation links
before-aria

### After -Add aria current to navigation links
aria-after

### Before - Remove aria labels on the homepage cards links
before-label-aria

### After - Remove aria labels on the homepage cards links
after-label-aria

### Before -Update image alt tags to empty

before-image-alt

### After -Update image alt tags to empty
after-image-alt

### DIFF
--- a/ds_judgements_public_ui/templates/includes/header.html
+++ b/ds_judgements_public_ui/templates/includes/header.html
@@ -14,17 +14,17 @@
       </button>
       <ul id="navigation" class="govuk-header__navigation-list">
         <li class="{% navigation_item_class 'home' %}">
-          <a class="govuk-header__link" title="Home" href="{% url 'home' %}">
+          <a class="govuk-header__link" title="Home" href="{% url 'home' %}"{% if request.resolver_match.view_name == "home" %} aria-current="true"{% endif %}>
             Home
           </a>
         </li>
         <li class="{% navigation_item_class 'about_this_service' %}">
-          <a class="govuk-header__link"  title="About" href="{% url 'about_this_service' %}">
+          <a class="govuk-header__link"  title="About" href="{% url 'about_this_service' %}"{% if request.resolver_match.view_name == "about_this_service" %} aria-current="true"{% endif %}>
             About
           </a>
         </li>
         <li class="{% navigation_item_class 'help_and_guidance' %}">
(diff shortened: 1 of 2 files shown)


ds_judgements_public_ui/templates/includes/header.html
```diff
@@ -14,17 +14,17 @@
       </button>
       <ul id="navigation" class="govuk-header__navigation-list">
         <li class="{% navigation_item_class 'home' %}">
-          <a class="govuk-header__link" title="Home" href="{% url 'home' %}">
+          <a class="govuk-header__link" title="Home" href="{% url 'home' %}" aria-current="true">
```
jacksonj04, line 17: Unless I'm misreading this, this change will add `aria-current="true"` to all of the header nav items whether or not they are actually the current page or not.

To do this properly we actually need to add something to the page `context` object as well, so we can make these `aria` attributes conditional.

jacksonj04 (changes requested):
Card changes look good (just an errant space we could tidy up), but the `aria-current` link changes I think need a new bit of page `context` so they can be made conditional.

ds_judgements_public_ui/templates/includes/info_cards_homepage.html
```diff
@@ -3,32 +3,32 @@
 <div class="homepage-cards__outer-container">
   <div class="homepage-cards__inner-container">
     <h2>Learn more about Find Case Law</h2>
-    <a href="{% url 'courts_and_tribunals' %}" aria-label="Browse by court or tribunal"  class="homepage-cards__info-cards-home">
+    <a href="{% url 'courts_and_tribunals' %}"  class="homepage-cards__info-cards-home">
```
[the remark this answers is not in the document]
Donna-H: @jacksonj04  okay updated

ds_judgements_public_ui/templates/includes/header.html
Donna-H: @jacksonj04 okay updated thanks for the help.

jacksonj04 (approved):

PR #1944, opened 2025-04-22
[FCL-811] Remove duplicated declaration of canonical URL in document headers

jacksonj04:
We're declaring a canonical URL twice when we don't need to; clean it up.


dragon-dxw (approved):
(the other one is in base.html)

PR #1950, opened 2025-04-23
[FCL-826] Redirect press-summary/1 to press-summary

dragon-dxw:
## Changes in this PR:

Supreme Court reported they use /press-summary/1 to point to press summaries -- support this again.

### After


jacksonj04 (approved):
This works for the time being, but we will potentially (as with all things) need to revisit it in future!

PR #1955, opened 2025-04-24
[FCL-763] add description to change button in review summary list 

Donna-H:
## Changes in this PR:
The component used in gov here https://design-system.service.gov.uk/components/summary-list/ does not have 'change' as a button **as we do in our code** therefore the no 'visually hidden text 'span class was omitted.  
The button uses template language to dynamically retrieve the list key/label.  So aria-label attribute has been used to completely over-ride the button text content, whilst duplicating the visible text and adding the key/label content. Therefore will now be descriptive of their purpose and function as they will now
reference the question that the user will change. This means that screen reader users are now made fully aware.
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-763
### Before
before

### After
after

### DIFF
--- a/transactional_licence_form/templates/review.html
+++ b/transactional_licence_form/templates/review.html
@@ -17,6 +17,7 @@
                 <button name="wizard_goto_step"
                         type="submit"
                         class="button-link"
+                        aria-label="Change question {{ key|get_field_name:all_field_names }}"
                         value="{{ form_key }}">Change</button>
               </dd>
             </div>


anthonyhashemi (approved):
Approving but this made me wonder, why aren't the accebility tests in playwright not catching this, and this led me to find that even on the run in this PR there are some accessiblity warnings being raised but they arent causing the action to error out and so its happening silently. See https://github.com/nationalarchives/ds-caselaw-public-ui/actions/runs/14647788299/job/41334512007


I think fixing that so it does actually cause a broken check is important and is a great opportunity for you to get familiar with that bit of CI / the test suite in either the next sprint or the one after

anthonyhashemi:
Having read the code used for the accessibility testing, it seems to just raise errors for critical issues and just check lower issues. If its not that many to fix, it could be worth having it strict now because I think we have fixed so much. It only seems to be flagging one warning. Will bring up at backlog review/ planning

PR #1989, opened 2025-05-01
FCL-754 | fix issues with sticky judgment header

jlhdxw:
## Changes in this PR:

Fixes the judgment sticky header on various breakpoints.

One of the main issues was we were using `100vw` for the width, which doesn't take into account of scrollbars. `width:100%` works perfectly fine instead and takes into consideration if the user has scrollbars visible or not.

Also removed some redundant styles that were being directly overridden below.

https://national-archives.atlassian.net/browse/FCL-754

### Before

image

image

### After

image

image

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_document_navigation_links.scss
+++ b/ds_judgements_public_ui/sass/includes/_document_navigation_links.scss
@@ -2,7 +2,7 @@
   position: fixed;
   top: 0;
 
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):

Terry-Price:
@Floppy looks good, I was in training so missed this, I thought I might have had a chance to take a look

PR #1993, opened 2025-05-02
[FCL-764] fix cookie issues 

Donna-H:
## Changes in this PR:
Fix cookie issues -fix header menu navigation position as it is inside the cookie banner in mobile 320px then update'skip to content for keyboard only
Moved the skip link back out of the header so its is fisrt element in body tag see https://www.a11y-collective.com/blog/skip-to-main-content/#:~:text=Placement%3A%20The%20'Skip%20to%20main,accessible%20before%20any%20navigation%20menus.

Screenshot from 2025-05-02 13-48-31

The skip link was was initial moved into header in this ticekt https://national-archives.atlassian.net/browse/FCL-414 and the PR 
https://github.com/nationalarchives/ds-caselaw-public-ui/pull/1617

 but an exception to the rule is the skip link see https://dequeuniversity.com/rules/axe/4.4/region?application=axeAPI
Screenshot from 2025-05-02 13-25-34

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-764
### Before
before

before-2

### After
after

Screenshot from 2025-05-02 13-39-32

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_links.scss
+++ b/ds_judgements_public_ui/sass/includes/_links.scss
@@ -1,19 +1,27 @@
 #skip-to-main-content {
   position: absolute;
-  top: -3rem;
+  z-index: 999;
+  top: 0.313rem;
+  left: 0.313rem;
+
+  padding: $space-1;
+  border: 0.19rem solid colour-var("accent-link");
+
+  color: colour-var("accent-link");
+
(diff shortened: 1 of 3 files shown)


Terry-Price (approved):
@Donna-H - thank you, glad that it's hidden now.

PR #1997, opened 2025-05-02
Use identifier for XML filename not marklogic uri

dragon-dxw:
When downloading the XML it would download with the filename of the marklogic URI. Instead download with a public identifier.

Before:
image


judgments/tests/tests.py
```diff
@@ -176,6 +176,7 @@ def test_xml(self, mock_get_document_by_uri):
         mock_get_document_by_uri.assert_called_with("ml-eat/2023/1")
         self.assertContains(response, "This is a document.")
         self.assertEqual(response.headers.get("X-Robots-Tag"), "noindex,nofollow,noai")
+        self.assertEqual(response.headers.get("Content-Disposition"), "attachment; filename=tna.tn4t35ts.xml")
```
jacksonj04, line 179: Have we tested this on various platforms to make sure having two dots in filenames doesn't cause explosions?

dragon-dxw:
Behaviour looks good on staging. Closed.

PR #2005, opened 2025-05-06
FCL-178 | ensure we validate the word count for the public statement …

jlhdxw:
…field

## Changes in this PR:

Validates the public statement field on submit to ensure the user doesn't go over 150.

Also note `maxlength` isn't a valid field on a textarea, and it wouldn't work anyway as we're limiting on words not characters.

Wording might need adjusting?

https://national-archives.atlassian.net/browse/FCL-178

## Screenshots of UI changes:

image

### DIFF
--- a/transactional_licence_form/utils.py
+++ b/transactional_licence_form/utils.py
@@ -5,6 +5,7 @@ from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
 
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.template.loader import render_to_string
 from django.utils.safestring import mark_safe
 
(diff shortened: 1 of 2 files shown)


dragon-dxw:
If we're aiming for 150 words and that isn't an absolute maximum, we should probably allow at least 300, maybe 450 max?
Especially since they won't know how many words they've typed until they submit.

transactional_licence_form/utils.py
```diff
@@ -87,3 +88,9 @@ def sanitize_and_format_response_as_xml(form_data):
         else:
             sanitized_fields[key] = sanitize_value(value)
     return render_to_string(EMAIL_TEMPLATE_PATH, sanitized_fields)
+
+
+def validate_max_words(value, max_words=150):
```
dragon-dxw, line 93: can we pop an `int` on max_words

dragon-dxw (approved):

PR #2012, opened 2025-05-08
Fcl 737/move breadcrumbs out of header

jlhdxw:
## Changes in this PR:

Some adjustments to the breadcrumbs:

 - Move it out of the header so it's semantically correct
 - Remove the colon and padding from the prefix
 - Fix the chevron on mobile so it's aligned correctly

https://national-archives.atlassian.net/browse/FCL-737

### Before

image

### After

image


Terry-Price (approved):
I'm loving this now, thanks @Floppy

PR #2013, opened 2025-05-08
FCL-792 | adjust homepage card text lineheight

jlhdxw:
## Changes in this PR:

Adjust the line-height of the cards text component. As the requested line height wasn't one in our list of variables, I needed to add another and adjust the existing variable usage.

https://national-archives.atlassian.net/browse/FCL-792

### Before

image

### After

image


Terry-Price (approved):
thanks @Floppy

PR #2030, opened 2025-05-13
MultipleResolutionsError reports what the resolutions are

dragon-dxw:
## Changes in this PR:
Noticed that MultipleResolutionsError doesn't tell you the offending document names.

https://app.rollbar.com/a/dxw/fix/item/tna-caselaw-public-ui/511

### After


anthonyhashemi (approved):
lovely stuff

PR #2052, opened 2025-05-21
[FCL-862] Swap Atom XSLT to use correct link tag instead of id

jacksonj04:
Using the value of the `<id>` element instead of the appropriate `<link>` tag means links to individual documents in the Atom feed for users viewing it in the browser do not resolve correctly.


anthonyhashemi (approved):
approved. Wondering do we have a test for this xsl at some level?

PR #2059, opened 2025-05-22
[FCL-865] Add redirects from /id/ URIs to documents

jacksonj04:
URIs at `/id/` represent the concrete concept of a work. When accessed, they should redirect users to an expression.

## Jira

FCL-865


judgments/tests/test_id_dispatch.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+
+        response = self.client.get("/id/d-a1b2c3")
+
+        mock_get_document_by_uri.assert_called_once_with("d-a1b2c3")
+
+        self.assertEqual(response.status_code, 303)
+        self.assertEqual(response.headers.get("Location"), "/uksc/2025/123")
+
+    @patch("judgments.resolvers.id_dispatch.get_published_document_by_uri")
```
dragon-dxw, line 23: I'm not sure this test makes sense -- we should be testing that we get a 404 if no valid identifier match exists, rather than if something is returned.

jacksonj04: `get_document_by_uri()` will do the job of raising a 404 exception if the document itself doesn't exist. This is testing that if the document does exist but for some reason doesn't have a public URL (which shouldn't happen, but let's be defensive here) that we return a 404 as we have nowhere to point the user.

judgments/tests/test_id_dispatch.py
```diff
@@ -0,0 +1,32 @@
...
+
+class TestIdDispatchEngine(TestCaseWithMockAPI):
+    @patch("judgments.resolvers.id_dispatch.get_published_document_by_uri")
+    def test_id_dispatch_for_document(self, mock_get_document_by_uri):
+        """Check that IdDispatchEngine tries to retrieve the correct document, and redirects to the right place using the right code."""
+        mock_get_document_by_uri.return_value = JudgmentFactory.build(
+            is_published=True, identifiers=[NeutralCitationNumber(value="[2025] UKSC 123")]
+        )
+
+        response = self.client.get("/id/d-a1b2c3")
```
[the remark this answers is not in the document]
jacksonj04: Yes; a document's URI on the database forms its canonical work URI. This is just a mechanism for getting from that URI to a representation.

PR #2081, opened 2025-05-28
FCL-877 | add snapshot testing

jlhdxw:
## Changes in this PR:

Adds screenshot snapshot testing. Saves the expected screenshot and the actual - we could ignore the actual ones specifically if we want to as it get overwritten on every e2e test. Not sure it's worth it though?

I had to remove the "check for large files" check as it doesn't allow any of the screenshots to be added to the repo.

https://national-archives.atlassian.net/browse/FCL-877


anthonyhashemi:
The page(s) with dynamic content such as the homepage will always break if this is running against prod, and will break when run across different envs (prod vs stg), so I think we need to rethink whether those can be incorporated here. For those we cold have a followup to do component visual regression tests, or have an env with static data (not possible currentry and I doont see it being worth spinning up a new env just for that)

e2e_tests/utils/assertions.py
```diff
@@ -64,3 +67,34 @@ def format_violation(violation):
         violation_str += "\n"
 
     return violation_str
+
+
+def compare_snapshot(actual_path, expected_path):
+    actual_image = Image.open(actual_path).convert("RGB")
+    expected_image = Image.open(expected_path).convert("RGB")
```
dragon-dxw, line 74: Concerned that we store two copies in git -- surely we just need one?

e2e_tests/utils/assertions.py
```diff
@@ -64,3 +67,34 @@ def format_violation(violation):
         violation_str += "\n"
 
     return violation_str
+
+
+def compare_snapshot(actual_path, expected_path):
+    actual_image = Image.open(actual_path).convert("RGB")
```
dragon-dxw, line 73: The images weigh 31 megabyte per commit which changes the styling -- that means we'll hit a gigabyte after 32 commits, so this might not be a sustainable approach.

e2e_tests/utils/assertions.py
```diff
@@ -64,3 +67,34 @@ def format_violation(violation):
...
+    expected_image = Image.open(expected_path).convert("RGB")
+
+    diff = ImageChops.difference(actual_image, expected_image)
+    diff_np = np.array(diff)
+    num_diff_pixels = np.sum(np.any(diff_np != 0, axis=-1))
+
+    return num_diff_pixels <= 0
```
dragon-dxw, line 80: The strictness of this check might cause issues across different hardware/browser versions

e2e_tests/utils/assertions.py
jlhdxw: Have removed the "actual" as that will change each time.

dragon-dxw: Github will be keeping the revision history of the modified pngs, which won't diff cleanly since they're not text based, so essentially, yes, this will store all snapshots indefinately.

anthonyhashemi (approved):
Looks good on my local too!

PR #2084, opened 2025-05-28
[FCL-884] Remove feedback banner from document pages

jacksonj04:
The blue feedback banner should not be present on document pages; remove it.


Terry-Price (approved):
Thanks @jacksonj04 - nice one.

PR #2110, opened 2025-06-06
FCL-912 | update font weight NCN results banner

Donna-H:
## Changes in this PR:
Update font weight NCN results banner
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-912
### Before
before

### After
after


Terry-Price (approved):
Thnaks @Donna-H this is so much more visible now.

PR #2119, opened 2025-06-09
FCL-911 | change highlight style

jlhdxw:
## Changes in this PR:

Changes the styling of the `mark` tag.

https://national-archives.atlassian.net/browse/FCL-911

### Before

image

### After

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -324,8 +324,7 @@ jobs:
         run: docker compose exec -d django python manage.py runserver 0.0.0.0:3000
 
       - name: Build playwright runner container
-        run: docker compose build e2e_tests
-      - run: docker compose up -d e2e_tests
+        run: docker compose up -d e2e_tests
 
       - name: Run playwright e2e tests
         run: docker compose exec e2e_tests pytest --base-url http://django:3000 --tracing retain-on-failure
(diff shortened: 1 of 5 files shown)


jacksonj04 (approved):

ds_judgements_public_ui/sass/includes/_typography.scss
```diff
@@ -52,13 +52,11 @@ code {
 
 mark {
   scroll-margin-top: 50vh;
-  border-bottom: 3px solid colour-var("accent-brand");
   font-weight: $typography-bold-font-weight;
-  background-color: transparent;
+  background: color-mix(in srgb, colour-var("accent-brand") 50%, transparent);
```
dragon-dxw, line 56: **incredibly** nitpicky because I don't think it matters with transparency since it isn't a colour: is srgb the right choice?

> Only use sRGB if you need to match the behavior of a specific device or software that uses sRGB. The sRGB color space is neither linear-light nor perceptually uniform, and produces poorer results such as overly dark or grayish mixes. -- https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix

jlhdxw: I think in our case it doesn't matter as it's just transparency.

Terry-Price (approved):
Thanks @jlhdxw this is looking smart, let;s see if when it's tested if the colour strength is enough.

PR #2121, opened 2025-06-10
FCL-919 | add home link to footer

jlhdxw:
## Changes in this PR:

Links the "Find Case Law" text in the footer to the homepage

https://national-archives.atlassian.net/browse/FCL-919

### Before

image

### After

image


Terry-Price (changes requested):
Hi @jlhdxw Please can I check, this footer link will not appear on the homepage, (just plain text). this will avoid reloading the same page if they click it on the homepage.
Can we do this if it isn't already like that.

PR #2122, opened 2025-06-10
FCL 906 Refine labels search results tools font and padding

Donna-H:
## Changes in this PR:
Refine labels search results tools font family, size and padding
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-906
### Before
before

### After

after-text


Terry-Price (changes requested):
All good @Donna-H except I am unsure if you have a dedicated card for the button text change, it should say 'Apply'.

Terry-Price (approved):
Thanks @Donna-H all good.

PR #2124, opened 2025-06-11
Fcl 924/add atom feed tooltip

jlhdxw:
### Before

image

### After

image

image

image

image


Terry-Price (changes requested):
Hi @jlhdxw the desktop looks really nice, can the mobile info panel be the same width as the form fields. so it only takes up two lines and so is wider.

Terry-Price (approved):
Thanks @jlhdxw - looks really good now wider. lets see if the text size is okay. I think I had it set at 0.9rem. looks small on mobile?

PR #2127, opened 2025-06-12
Remove e2e_build step

dragon-dxw:
We theorise that the separate `docker compose build e2e_tests` should be unnecessary since `docker compose up -d e2e_tests` should cover it.


anthonyhashemi (approved):
We have no idea why the build step used to work and then suddently started giving
`no such service e2e_tests`
but seems like ` docker compose up -d e2e_tests` works fine. 

Dont need to build the image separately, compose up will do it if necessary.

Maybe there was a reason we needed to build separately. lets see if it impacts but doubt it

PR #2131, opened 2025-06-13
[FCL-909] Update search bar on search results to match homepage

Donna-H:
## Changes in this PR:
Update search bar on search results to match homepage
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-909
### Before
before

### After
after


jacksonj04 (commented):
Visually this looks fine, but it feels like a ripe candidate for some deduplication work so that they both include the same element and pull from the same CSS. Less to maintain in future that way!

Terry-Price (approved):
Thanks @Donna-H

PR #2148, opened 2025-06-18
[fcl-928] update search bar on Advanced search results to match homepage

Donna-H:
## Changes in this PR:
Update 'search bar' on Advanced search results to match homepage
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-928
### Before
before

### After
after

### DIFF
--- a/ds_judgements_public_ui/sass/includes/_structured_search.scss
+++ b/ds_judgements_public_ui/sass/includes/_structured_search.scss
@@ -39,6 +39,19 @@
     text-align: left;
   }
 
+  &__container-advanced {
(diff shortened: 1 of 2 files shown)


jlhdxw:
This would be a good opportunity to consolidate the 2 components into 1 reusable component across the homepage and advanced search.

I've had a look at the 2 components and they similar enough that it should be straightforward to have one "search_form.html" component that we pass some params to.

Terry-Price (approved):
Thanks @Donna-H looks so much better

PR #2150, opened 2025-06-19
[fcl-935] fix bug align doc navigation

Donna-H:
## Changes in this PR:
Fix bug to align doc navigation
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-935
### Before
before

### After
after


Terry-Price (commented):
Thanks @Donna-H

PR #2153, opened 2025-06-19
FCL-845 | don't expose internal id on pdf download

jlhdxw:
## Changes in this PR:

Hides the ID on the html and ensures we give the user a good filename when they download it.

https://national-archives.atlassian.net/browse/FCL-845

### DIFF
--- a/judgments/views/detail/best_pdf.py
+++ b/judgments/views/detail/best_pdf.py
@@ -10,7 +10,7 @@ from judgments.models.document_pdf import DocumentPdf
 
 def best_pdf(request, document_uri):
     """
-    Response for the legacy data.pdf endpoint, used by data reusers
+    Response for the data.pdf endpoint, used by data reusers
 
     If there's a DOCX-derived PDF in the S3 bucket, return that.
     Otherwise fall back and redirect to the weasyprint version."""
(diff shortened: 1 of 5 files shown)


dragon-dxw (approved):

judgments/views/detail/best_pdf.py
```diff
@@ -15,13 +15,21 @@ def best_pdf(request, document_uri):
     If there's a DOCX-derived PDF in the S3 bucket, return that.
```
dragon-dxw, line 15: I'm not sure this is a "legacy data.pdf" any more.

judgments/views/detail/best_pdf.py
```diff
@@ -15,13 +15,21 @@ def best_pdf(request, document_uri):
...
 
-    if response.status_code != 404:
-        logging.warn(f"Unexpected {response.status_code} error on {document_uri} whilst trying to get_best_pdf")
+    external_response = requests.get(pdf.generate_uri())
+    logging.debug("Response %s", external_response.status_code)
+
+    if external_response.status_code == 200:
+        response = HttpResponse(external_response.content, content_type="application/pdf")
+
+        response["Content-Disposition"] = f"attachment; filename={document_uri}.pdf"
```
dragon-dxw, line 25: I'm not sure this is the right type of URI to be using -- could we have a test for the header?

jlhdxw: Have added a test for the whole view as it was missing.

PR #2157, opened 2025-06-20
FCL-925 | remove pipes and inline matching text on search

jlhdxw:
## Changes in this PR:

To make it easier to understand the snippets on search, we're removing the pipes and not showing them inline

https://national-archives.atlassian.net/browse/FCL-925

### Before

image

### After

image


dragon-dxw (approved):
Code good -- but do we now potentially have too much snippet?

PR #2163, opened 2025-06-23
FCL-964: atom feed id tag value is fclid

dragon-dxw:
## Changes in this PR:

The <id> tag in the atom feed will be `https://caselaw.nationalarchives.gov.uk/id/fcl.bcdfghjk` not `d-a1b2c3`. This may be a small breaking change for consumers.

Requires https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/1105 to ensure the tests pass (ensures search result factories have proper Identifiers not dicts, and that there's an FCL ID)

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136/backlog?assignee=unassigned%2C60bdecb0c90cb2006806ae0f&selectedIssue=FCL-964

### DIFF
--- a/judgments/feeds.py
+++ b/judgments/feeds.py
@@ -6,6 +6,7 @@ from caselawclient.client_helpers.search_helpers import (
     search_judgments_and_parse_response,
 )
 from caselawclient.models.identifiers import Identifier
+from caselawclient.models.identifiers.fclid import FindCaseLawIdentifier
 from caselawclient.responses.search_result import SearchResult
 from caselawclient.search_parameters import SearchParameters
 from django.contrib.syndication.views import Feed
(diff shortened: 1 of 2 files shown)


dragon-dxw:
I'm not sure where best to document this (as per Jira)

judgments/feeds.py
```diff
@@ -158,7 +159,17 @@ def item_title(self, item):
         return item.name
 
     def item_guid(self, item) -> str:
-        return "https://caselaw.nationalarchives.gov.uk/id/" + item.uri
+        fclid = item.identifiers.of_type(FindCaseLawIdentifier)
```
jacksonj04, line 162: Thought: Since we're confident that FCLIDs (and indeed NCNs, but that's another problem...) belong to the _work_ not the _expression_, is it worth shimming this in the API so we can do something like `Document.work.id`?

judgments/feeds.py
```diff
@@ -158,7 +159,17 @@ def item_title(self, item):
...
+
+        if not fclid:
+            msg = f"No FindCaseLawIdentifier for search result {item}"
+            raise RuntimeError(msg)
+
+        if len(fclid) > 1:
+            msg = f"Multiple FindCaseLawIdentifiers for search result {item}"
+            raise RuntimeError(msg)
+
+        return "https://caselaw.nationalarchives.gov.uk/id/" + fclid[0].url_slug
```
jacksonj04, line 172: I don't think this should be the slug, since the whole point of slugs is to provide a shorthand version. Maybe we should decide on some ID namespaces (eg `/id/work/` or `/id/w-`)?

jacksonj04 (changes requested):
Some comments to talk through. The underlying principle seems sound though!

PR #2201, opened 2025-07-04
Fcl 953/add hidden ncn labels

jlhdxw:
## Changes in this PR:

Adds a visually hidden/reset visually hidden mixin and class so we can show/hide things visually but show for machine readers where the label is relevant.

https://national-archives.atlassian.net/browse/FCL-830

## Screenshots of UI changes:

(only visible for screen readers/robots 🤖 )

image

image


Donna-H (approved):

Terry-Price (commented):
@jlhdxw Should we use `clip-path`?
I think `clip: rect` is depreciated now.

PR #2212, opened 2025-07-07
FCL-1005: Dynamic URIs in data.xml

dragon-dxw:
Needs v38.0.1 of the API client.

### After


anthonyhashemi:
Looks ok but ci complaining that the apiclient version hasnt been released

dragon-dxw:
Needs feature flagging.

PR #2219, opened 2025-07-10
[fcl-957]add a distinct label to the search bar

Donna-H:
## Changes in this PR:
Add a distinct label to the search bar 
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-957
### Before
before-home
before-advance

### After
after-home
after-advance


jlhdxw (approved):

Terry-Price (commented):
Thanks for doing this @Donna-H

PR #2221, opened 2025-07-11
FCL-988: Explicitly state assets_base in feed

dragon-dxw:
Before there wasn't a way of getting the base assets URI. Now there's tna:assets 

`<tna:assets href="https://assets.caselaw.nationalarchives.gov.uk/ewhc/ch/2024/505/"/>`

docs PR: https://github.com/nationalarchives/ds-find-caselaw-docs/pull/205

https://national-archives.atlassian.net/browse/FCL-988

### After

### DIFF
--- a/judgments/feeds.py
+++ b/judgments/feeds.py
@@ -106,6 +106,9 @@ class JudgmentAtomFeed(Atom1Feed):
                     "href": f"https://assets.caselaw.nationalarchives.gov.uk/{uri}/{path_underscore}.pdf",
                 },
             )
+            handler.addQuickElement(
+                "tna:assets_base", "", {"href": f"https://assets.caselaw.nationalarchives.gov.uk/{uri}/"}
+            )
 
     def add_root_elements(self, handler):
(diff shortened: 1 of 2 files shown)


judgments/feeds.py
```diff
@@ -106,6 +106,9 @@ def add_item_elements(self, handler, item) -> None:
                     "href": f"https://assets.caselaw.nationalarchives.gov.uk/{uri}/{path_underscore}.pdf",
                 },
             )
+            handler.addQuickElement(
+                "tna:assets", "", {"href": f"https://assets.caselaw.nationalarchives.gov.uk/{uri}/"}
```
jacksonj04, line 110: Would `assets_base` be clearer?

dragon-dxw:
Now pondering if the assets-base would make more sense in the XML...

dragon-dxw:
> Can we document this decision to adding this field to the API with this name

Done. https://github.com/nationalarchives/ds-find-caselaw-docs/pull/205

PR #2252, opened 2025-07-24
FCL 1056: Set up memray

dragon-dxw:
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136

## Screenshots of UI changes:
image

image

new commands:

`fab memray` -- launch an instrumented version of the app: use `fab live` to get a live view (in a different window)
`fab flamegraph` -- run an instrumented version of the app and output a flamegraph of max memory at end

### After

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -112,6 +112,58 @@ def run(c): ...
 
 
 @task
+def memray(c):
+    """Launch the server with memray tracking for live connections with `live`"""
+    start(c, "django")
+    background_exec("npm run watch", "assets")
+    collectstatic(c)
+    try:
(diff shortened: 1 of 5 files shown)


fabfile.py
```diff
@@ -111,6 +111,58 @@ def runquick(c):
 def run(c): ...
 
 
+@task
+def memray(c):
+    """Launch the server with memray tracking for live connectiosn with `live`"""
```
anthonyhashemi, line 116:
```suggestion
    """Launch the server with memray tracking for live connections with `live`"""
```

anthonyhashemi (approved):
Ah this is really nice, gave it a spin locally 👌

PR #2281, opened 2025-08-06
[SPIKE] FCL-1097 | WIP - jinja spike

jlhdxw:
## Changes in this PR:

Showing what adding jinja templates could look like.  The about page loads, but there's some stuff broken in the footer. 

https://national-archives.atlassian.net/browse/FCL-1097

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -184,7 +184,18 @@ TEMPLATES = [
                 "judgments.context_processors.environment",
             ],
         },
-    }
+    },
+    {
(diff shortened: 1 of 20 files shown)


dragon-dxw (commented):
❤️

anthonyhashemi:
Looks good! I think we are generally happy as a team to slowly move to jinja2 rather than DTL as it will let us leverage the govuk and tna jinja macros etc but yeah let's present this to the team and then maybe we can sit down and write up a small epic to track the work.

PR #2332, opened 2025-08-20
Add new feature flag for search query highlighting

jacksonj04:
Search query highlighting is the primary cause of long-running MarkLogic queries, and we suspect it's leading to unexpected load. Move the entire feature behind a flag (`document_search_query_highlighting`), allowing us to do staged roll-out (and easier roll-back) in future.


dragon-dxw (commented):
We could also put the search index link behind the same feature flag so `?query` isn't passed to the judgment detail page at all, but this isn't necessary.

PR #2337, opened 2025-08-22
[fcl-1095-1094] Update licence form layout and e2etests

Donna-H:
## Changes in this PR:
Update licence form layout and e2etests
https://national-archives.atlassian.net/browse/FCL-1095
https://national-archives.atlassian.net/browse/FCL-1094
### Before

[before.webm](https://github.com/user-attachments/assets/54be1ec7-9939-4798-b5e6-e20a86c4d0e6)

### After

[after.webm](https://github.com/user-attachments/assets/265a5e59-09bc-4f34-89b2-75d77032a329)


e2e_tests/test_transactional_licence_form.py
```diff
@@ -17,8 +17,18 @@ def test_transactional_licence_form(page: Page):
     assert_is_accessible(page)
     assert_matches_snapshot(page, "re_use_find_case_law_records_page")
     page.get_by_text("I want to perform computational analysis").click()
+
+    assert_is_accessible(page)
```
anthonyhashemi, line 21: we noticed the visual regression tests (and accessibility tests) were missing for these pages. 

This approach of grouping accessbility testing and visual regression testing alongside the e2e user journey test does have the benefit of not needing to navigate to these pages multiple times but it widens the focus of the test and makes it possible that we miss pages we wanted to test for accesiblity or visual regression as we had preiobusly done here.

I feel like if we had a dedicated test which tested all urls as defined in a django url definition files automatically then things would be easier to catch, and if there are exceptions we could exclude them.

Thoughts @jlhdxw ?

anthonyhashemi (changes requested):
see comment

transactional_licence_form/templates/layouts/form_step.html
```diff
@@ -40,7 +40,7 @@ <h3>Licence application steps</h3>
     {% block pre_title %}
     {% endblock pre_title %}
     {% block step_title %}
-      <h1 class="transactional-licence-form__step-title">{{ wizard.form.title }}</h1>
+      <h1 class="transactional-licence-form__step-title">{{ wizard.form.navigation_current }}</h1>
```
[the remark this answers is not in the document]
Donna-H: @anthonyhashemi I have reverted the branch back to earlier commit and kept the title but refactored the rest of navigation then James can review this too.

anthonyhashemi (approved):
Great work, and thanks for putting in the effort to refactor, code will be much easier to maintain now 👌

e2e_tests/test_transactional_licence_form.py
jlhdxw: I think these static pages related to the transactional license form could easily be separate tests rather than having them here.

The issue with testing each of the transactional license form pages individually is we need to run through all the previous pages to fill out the form to get to them, so it would increase the run time by quite a bit. 

I'd opt for (not in this PR) moving that part out of this test and having the form and static pages as separate tests.

PR #2346, opened 2025-08-29
Reduce the number of requests before a worker is restarted

rjw1:
We are still seeing a memory leak, so let's reduce the number of
requests again to see if that will stop the instances running out of
memory and then causing downtime when the OOM killer decides to do its
thing.


anthonyhashemi:
CI should be fixed with a rebase/merge from main

PR #2383, opened 2025-09-10
FCL-1276 | add jinja templating

jlhdxw:
## Changes in this PR:

Adds jinja templating. Also adds a page where we can add jinja components as we build them.

The base.jinja has the very basic stuff needed to render a page - doesn't include the header/footer etc as that'll come later.

https://national-archives.atlassian.net/browse/FCL-1276

## Screenshots of UI changes:

Page for adding templates currently looks like this (could swap this to storybook at a later date):

image

### DIFF
--- a/ds_judgements_public_ui/templates/layouts/base.jinja
+++ b/ds_judgements_public_ui/templates/layouts/base.jinja
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html lang="en-gb">
+  <head>
+    <meta charset="utf-8" />
+    {% block meta_description %}
+      {% if page_description %}
+        <meta name="description" content="{{ page_description }}" />
+      {% endif %}
+    {% endblock meta_description %}
+
+    {% block robots %}
+      {% if not page_allow_index %}
+        <meta name="robots" content="noindex,nofollow,noai" />
+      {% endif %}
+    {% endblock robots %}
+
+    <meta http-equiv="x-ua-compatible" content="ie=edge" />
+
+    {% if page_canonical_url %}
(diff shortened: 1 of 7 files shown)


judgments/jinja.py
```diff
@@ -0,0 +1,22 @@
+from django.contrib.staticfiles.storage import staticfiles_storage
+from django.urls import reverse
+from jinja2 import Environment, select_autoescape
+
+
+def environment(**options):
```
anthonyhashemi, line 6: could you explain the choices made here?

anthonyhashemi (approved):

ds_judgements_public_ui/templates/layouts/base.jinja
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
jlhdxw: Essentially yes - the only difference is when you call templatetags they're just function calls in jinja.

judgments/jinja.py
jlhdxw: A lot of this is boilerplate, but there's some specific bits I had to add to satisfy flake8:

This was added to explicitly autoescape in the templates - flake8 flagged this up.
```python
    options.pop("autoescape", None)
    env = Environment(
        autoescape=select_autoescape(
            enabled_extensions=("jinja"),
            default_for_string=True,
            default=True,
        ),
        **options,
    )
```

This is how you add any functions that you want to be available in the template engine. The main ones django gives you out of the box are static and url. Any new ones we want to add, we will need to add them here.
```python
    env.globals.update(
        {
            "static": staticfiles_storage.url,
            "url": reverse,
        }
    )
```

PR #2394, opened 2025-09-12
FCL-1195 | pass browser window size through to feedback link

jlhdxw:
## Changes in this PR:

Allows sending the browser dimensions through to smart survey as url params. We may want to add other things, so this PR is a draft for now.

https://national-archives.atlassian.net/browse/FCL-1195

### DIFF
--- a/ds_judgements_public_ui/static/js/src/feedback_link.js
+++ b/ds_judgements_public_ui/static/js/src/feedback_link.js
@@ -8,6 +8,25 @@ export const setupFeedbackLink = function () {
     const width = window.innerWidth || document.documentElement.clientWidth;
     const height = window.innerHeight || document.documentElement.clientHeight;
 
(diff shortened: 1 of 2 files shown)


ds_judgements_public_ui/static/js/src/feedback_link.js
```diff
@@ -0,0 +1,36 @@
...
+
+    if (uaData && uaData.platform) {
+        url.searchParams.set("device-platform", uaData.platform);
+        url.searchParams.set(
+            "device-class",
+            uaData.mobile ? "mobile" : "desktop",
+        );
+    } else {
+        const deviceClass =
+            width <= 768 ? "mobile" : width <= 1024 ? "tablet" : "desktop";
```
dragon-dxw, line 21: Dubious about the accuracy of these bins, but as long as they're useful

PR #2413, opened 2025-09-19
[FCL-1217] create navigation component

Donna-H:
## Changes in this PR:
Create navigation component
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-1217
### After


jlhdxw (changes requested):

ds_judgements_public_ui/templates/components/navigation.jinja
```diff
@@ -0,0 +1,11 @@
+{% macro nav_content(title=None) %}
```
jlhdxw, line 1: We can the component `contents` here too

PR #2442, opened 2025-09-29
Allow HEAD request on search (by removing the explicit check on HTTP verb)

dragon-dxw:
https://dxw.slack.com/archives/C02TP2L2Z0F/p1758201302819729?thread_ts=1758200165.696399&cid=C02TP2L2Z0F
https://national-archives.atlassian.net/browse/FCL-1305

~~It might seem odd that .GET works just fine on HEAD, but apparently so. (It is GET lite, I suppose)~~
We remove the test as it isn't doing anything


jacksonj04 (commented):
I'm curious to know what the unintended side effect of removing this check entirely would be. I can't figure out what would break if we did.

jacksonj04 (approved):

PR #3015, opened 2026-04-10
Fix production django docke permissions, test in ci and improve dev docker compose setup

anthonyhashemi:
## Changes in this PR:

[fix: Fix more prod django docker write permissions](https://github.com/nationalarchives/ds-caselaw-public-ui/commit/ee6bbeb385871883bc1233c34e886593d21bf1ff)

[ci: add production smoke test and modernise docker-compose](https://github.com/nationalarchives/ds-caselaw-public-ui/commit/9cfc661be4a8cbea20a4a18a9bac57b19593e8b8) 

- Replace build-only CI step with compose-based production smoke test
  that builds the image, starts it, and verifies it serves via healthcheck
- Add docker-compose.prod-test.yml overlay targeting production stage
  with gunicorn, no volume mounts, and django user — matching ECS deployment
- Add MarkLogic service with healthcheck to docker-compose.yml
- Add postgres healthcheck and service_healthy depends_on conditions
- Replace hardcoded env vars and env_file with ${VAR:-default} syntax
- Update postgres image digest pin
- Update README documenting compose files, local vs production stages,
  MarkLogic setup, and Docker network requirement
  
  
**Previous error was:**
Screenshot 2026-04-10 at 18 54 59

https://national-archives.atlassian.net/browse/FCLC-624

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,7 @@ repos:
       - id: check-merge-conflict
       - id: check-xml
       - id: check-yaml
+        args: [--unsafe] # Allow !reset tag for clearing inherited volumes in prod test config. See https://github.com/pre-commit/pre-commit-hooks/issues/1090 issue comment for context.
       - id: end-of-file-fixer
       - id: forbid-submodules
       - id: mixed-line-ending
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,50 +9,100 @@ services:
       args:
         BUILD_ENVIRONMENT: local
     depends_on:
-      - postgres
+      postgres:
+        condition: service_healthy
+      marklogic:
+        condition: service_healthy
+    init: true
     volumes:
       - .:/app:z
       - ../ds-caselaw-custom-api-client:/apiclient
-    init: true
     environment:
-      DATABASE_URL: "postgres://postgres:admin@postgres:5432/ds_judgements_public_ui"
-      DJANGO_SETTINGS_MODULE: "config.settings.local"
-      SECRET_KEY: local_dev_secret_key
-      SECURE_SSL_REDIRECT: "false"
-    env_file:
-      - ./.env
+      DATABASE_URL: "postgres://${POSTGRES_USER:-postgres}:${POSTGRES_PASSWORD:-admin}@postgres:5432/${POSTGRES_DB:-ds_judgements_public_ui}"
+      DJANGO_SETTINGS_MODULE: config.settings.local
+      SECRET_KEY: ${SECRET_KEY:-local_dev_secret_key}
+      DJANGO_SECRET_KEY: ${DJANGO_SECRET_KEY:-local_dev_secret_key}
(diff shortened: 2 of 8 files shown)


docker-compose.yml
```diff
@@ -9,51 +9,101 @@ services:
...
-      PGPASSWORD: admin
+      POSTGRES_DB: ${POSTGRES_DB:-ds_judgements_public_ui}
+      POSTGRES_USER: ${POSTGRES_USER:-postgres}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-admin}
       POSTGRES_HOST_AUTH_METHOD: trust
     volumes:
       - ./database_backups:/backups:z
 
+  marklogic:
+    image: progressofficial/marklogic-db:latest
```
dragon-dxw: is it worth pinning?

.pre-commit-config.yaml
```diff
@@ -9,6 +9,7 @@ repos:
       - id: check-merge-conflict
       - id: check-xml
       - id: check-yaml
+        args: [--unsafe]
```
dragon-dxw: > --unsafe - Instead of loading the files, simply parse them for syntax. A syntax-only check enables extensions and unsafe constructs which would otherwise be forbidden. Using this option removes all guarantees of portability to other yaml implementations. Implies --allow-multiple-documents
https://stackoverflow.com/a/79673807

Why do we need this?

docker-compose.yml
anthonyhashemi: yes!

.pre-commit-config.yaml
anthonyhashemi: https://github.com/pre-commit/pre-commit-hooks/issues/1090 I can add a comment linking to this

dragon-dxw (approved):

docker-compose.yml
```diff
@@ -9,51 +9,101 @@ services:
...
       - ../ds-caselaw-custom-api-client:/apiclient
-    init: true
     environment:
-      DATABASE_URL: "postgres://postgres:admin@postgres:5432/ds_judgements_public_ui"
-      DJANGO_SETTINGS_MODULE: "config.settings.local"
-      SECRET_KEY: local_dev_secret_key
-      SECURE_SSL_REDIRECT: "false"
-    env_file:
-      - ./.env
+      DATABASE_URL: "postgres://${POSTGRES_USER:-postgres}:${POSTGRES_PASSWORD:-admin}@postgres:5432/${POSTGRES_DB:-ds_judgements_public_ui}"
```
dynamictulip, line 21: It would be better to use Docker Secrets rather than env vars to manage most of this info to prevent accidental exposure via logs etc
https://docs.docker.com/compose/how-tos/use-secrets/

dynamictulip: (by most info I mean the non-faked things like rollbar token)

anthonyhashemi: deffo! followup ticket created

PR #3073, opened 2026-04-22
Pin third-party actions to commit SHAs

kurtismash:
This PR pins third-party GitHub Actions to full commit SHAs for supply-chain security.

This has been done automatically by [pinact](https://github.com/suzuki-shunsuke/pinact). When reviewing please confirm that the SHAs are correct, zizmor will alert if not.

As a maintainer, please merge this PR once approved.

### DIFF
--- a/.github/workflows/publish-storybook.yml
+++ b/.github/workflows/publish-storybook.yml
@@ -34,12 +34,12 @@ jobs:
 
     steps:
       - name: Checkout Code Repository
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):

jacksonj04:
@kurtismash you might also be interested in https://docs.renovatebot.com/presets-default/#pindependencies as a Renovate option, which I've just added to our default configuration.

kurtismash:
> @kurtismash you might also be interested in https://docs.renovatebot.com/presets-default/#pindependencies as a Renovate option, which I've just added to our default configuration.

Renovate is the next thing on my list. I'm considering replicating Case Law's automated patching to a wider scope. More so for security patching when CVEs are found, but many teams are struggling with the Dependabot PR load.

PR #3117, opened 2026-05-08
FCLC-1879 | update single court page

jlhdxw:
## Changes in this PR:

Updates the single court page to show extra details. Added a snapshot test and VCR cassette.

https://national-archives.atlassian.net/browse/FCLC-1879
https://national-archives.atlassian.net/browse/FCLC-1880
https://national-archives.atlassian.net/browse/FCLC-1882


config/views/courts.py
```diff
@@ -76,16 +85,28 @@ def court(self):
         except CourtNotFoundException:
             raise Http404("Court not found")
 
+    def _get_search_response(self, search_parameters):
+        try:
+            return search_judgments_and_parse_response(api_client, search_parameters)
+        except (MarklogicResourceNotFoundError, RequestException):
```
dragon-dxw: Should we be blindly continuing in the face of these errors?

dragon-dxw (approved):
Does it make sense to push search up above Recent Judgments?
The whitespace around the headers looks wrong, it's narrower at the top than the bottom

config/views/courts.py
jacksonj04: I think for me, if we can get the court details but there's a transient failure to get recent judgments I'd prefer to degrade gracefully and show what we can. Recent judgments isn't the primary purpose.

We should probably bubble the error up for logging so we know about it though!

config/views/courts.py
```diff
@@ -76,16 +85,28 @@ def court(self):
...
+            return []
+
     def get_context_data(self, **kwargs):
 
         court = self.court
 
         context = super().get_context_data(**kwargs)
 
+        search_parameters: SearchParameters = SearchParameters(
+            court=kwargs["param"], page=1, order="-date", page_size=5
```
jacksonj04: I'd be tempted to swap this for `court.param`, so we interpret the URL once and then use the `court` everywhere downstream. Not a blocker though.

jacksonj04 (approved):
I agree with @dragon-dxw that the vertical spacing on the headers feels off.

ds_judgements_public_ui/templates/pages/court_or_tribunal.jinja
```diff
@@ -85,7 +85,7 @@
   {% endcall %}
   {% call card(title="Questions about judgments from this court", title_tag="h2") %}
     <p>
-      Please note that neither Find Case Law or The National Archives are responsible for the content of any court judgments provided to us by The Supreme Court.
+      Please note that neither Find Case Law or The National Archives are responsible for the content of any court judgments provided to us by the {{ court.name }}.
```
jacksonj04, line 88: Good catch!

requirements/base.txt
```diff
@@ -21,5 +21,5 @@ Jinja2>=3.0
 govuk-frontend-jinja>=3.4.0
 
 # Type stubs
-mypy-boto3-s3==1.43.5
```
jacksonj04, line 24: These can now be reverted as we're fixing the problem in APIC.

jacksonj04 (approved):

PR #3136, opened 2026-05-12
FCLC-1883 | Add court descriptions as jinja templates

jlhdxw:
## Changes in this PR:

Adds the descriptions from the utils into this repo and converts them to jinja.

The plan is other content will be put into these files for each courts as macros.

https://national-archives.atlassian.net/browse/FCLC-1883


jacksonj04 (changes requested):
We should probably exclude the `content/courts` directory from code coverage, or maybe even all `.jinja` files!

Other than that, this looks solid to me. We probably want to decompose these a bit further in future, stuff like the "read more" link can be dragged out into the structured data and then we can just do every paragraph there programmatically.

jacksonj04 (approved):

PR #3137, opened 2026-05-12
FCLC-1925 | add hidden field for court search

jlhdxw:
## Changes in this PR:

Adds a hidden field to the court specific search on the court page with the court param.

https://national-archives.atlassian.net/browse/FCLC-1925


jacksonj04 (approved):
Double checked to confirm that `court` also works for tribunals (it does).

Ship it!

PR #3197, opened 2026-05-28
Remove MarkLogic setup in favour of using repo

dancs-dev:
## Changes in this PR:

- Removes the local MarkLogic setup.
- Updates README with options for connecting to the MarkLogic repo

https://national-archives.atlassian.net/browse/FCLC-1951

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,6 @@ building of the project's containers:
 
 - **`django`** — The application container, built from a multi-stage Dockerfile (local dev stage by default)
 - **`postgres`** — PostgreSQL database
-- **`marklogic`** — Local MarkLogic instance (included in compose by default)
 - **`e2e_tests`** — Playwright end-to-end tests (opt-in via `--profile e2e_tests`)
 
 ### Compose files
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -65,22 +64,37 @@ cp .env.example .env
...
+If you need to connect to a live MarkLogic instance, there are two main options:
+
+1. Connect to the shared staging MarkLogic database.
+2. Set up a local MarkLogic server.
+
+**Setting up a local MarkLogic server:**
+
+Clone and follow the README of the [ds-caselaw-marklogic](https://github.com/nationalarchives/ds-caselaw-marklogic) repository to set up a local MarkLogic server.
+
+You can then set `MARKLOGIC_HOST=marklogic` in your `.env` to point the public interface at the local MarkLogic instance.
```
dragon-dxw, line 96: I've not tested this, but `host.docker.internal` was what I thought the marklogic repo used.

It's entirely possible there's two names and this is a better one.

dragon-dxw (approved):

README.md
dancs-dev: I believe this should work by default because [`docker-compose.yaml` in the MarkLogic repo uses `marklogic` as the container name](https://github.com/nationalarchives/ds-caselaw-marklogic/blob/main/docker-compose.yml#L3).

dragon-dxw:
Can you rebase these commits into one as well, the first is misleading as we went down a different path in the end. (I see this is done :) )

README.md
dragon-dxw: Awesome. I'm now pretty happy this is acceptable too :)

PR #3200, opened 2026-05-28
FCLC-1958 | use website link from court utils data

jlhdxw:
## Changes in this PR:

Use the website link from the court utils data. 

https://national-archives.atlassian.net/browse/FCLC-1958


jacksonj04 (approved):
Would be nice if we could move is_court_ended into the `Court` object from utils, but I'm happy for that to go in another ticket.

PR #3217, opened 2026-06-04
FCLC-881 | add correct config to allow for asset reload

jlhdxw:
## Changes in this PR:

This makes it so assets are compiled with a hash now, and jinja will reload when run locally.

The `READ_DOT_ENV_FILE = env.bool("DJANGO_READ_DOT_ENV_FILE", default=False)` meant we never loaded the .env file at this point, which meant DEBUG was never being set correctly locally. If we want to use this option, you can set it in the docker file or when the command is run.


config/settings/base.py
```diff
@@ -12,7 +12,7 @@
 APPS_DIR = ROOT_DIR / "ds_judgements_public_ui"
 env = environ.Env()
 
-READ_DOT_ENV_FILE = env.bool("DJANGO_READ_DOT_ENV_FILE", default=False)
+READ_DOT_ENV_FILE = env.bool("DJANGO_READ_DOT_ENV_FILE", default=True)
```
dragon-dxw, line 15: We need to ensure ths if this is correctly set to False in staging/prod.

dragon-dxw (approved):

config/settings/base.py
jlhdxw: Not needed as there is no .env file on staging/production and the next line doesn't do anything/doesn't error when there is no file.

PR #3255, opened 2026-06-11
FCLC-881 | remove asset test setup that was hiding errors

jlhdxw:
## Changes in this PR:

This was hiding asset errors in test.

https://national-archives.atlassian.net/browse/FCLC-881

### DIFF
--- a/config/settings/test.py
+++ b/config/settings/test.py
@@ -3,7 +3,6 @@ With these settings, tests run faster.
 """
 
 from .base import *  # noqa
-from . import base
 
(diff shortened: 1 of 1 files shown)


dragon-dxw (approved):

jlhdxw:
Closing this. When the tests are run on CI, we haven't done a collectstatic call, so the tests fail that do anything with static assets. We don't want to do a collectstatic call just for running pytest on CI, so I'm closing this PR and leaving it as is (assets not required for running pytest).

PR #3304, opened 2026-06-25
bump utils 4.7

dragon-dxw:
### After


dancs-dev (approved):
LGTM assuming e2e tests pass

dragon-dxw (commented):
I don't love the approach we're taking here, but it's consistent.

PR #3375, opened 2026-07-09
FCLC-2102 | handle when we don't have content for a court

jlhdxw:
## Changes in this PR:

Handles when we haven't explicitly got content for a court

https://national-archives.atlassian.net/browse/FCLC-2102

## Screenshots of UI changes:

An example of how this might look:

localhost_3000_courts-and-tribunals_ewcop_t2


dragon-dxw (approved):

dragon-dxw:
Code looks good but I'd question if we should be redirecting to the 'actual court' instead in this case?

PR #3426, opened 2026-07-17
FCLC-2158 | remove hidden input where we have a control

jlhdxw:
## Changes in this PR:

Ensure we don't have the hidden inputs for the order and per page controls so we don't double up the query param in the url when searching

https://national-archives.atlassian.net/browse/FCLC-2158


jacksonj04 (approved):
LGTM!

PR #3454, opened 2026-07-21
FCLC-2169 | clear up filter params to ensure only params set by users…

jlhdxw:
… are being applied

## Changes in this PR:

Clearup for the filter params. Ensures we don't send through blank values unnecessarily, and clears out page and order when you clear your filters to ensure you don't get positioned randomly in results.

Moved the search results controls into a macro and made it support the footer/normal versions and removed the deprecated includes.

https://national-archives.atlassian.net/browse/FCLC-2169


jacksonj04 (approved):
🚢

PR #3494, opened 2026-07-27
[FCLC-2057] Push GTM dataLayer page metadata for GA4 dimensions

jacksonj04:
- Push `page_type`, `court_code`, `court_type`, and `document_noun` into `window.dataLayer` before GTM loads so GA4 can register them as event-scoped custom dimensions
- Wire values on document, court browse/landing, search results, and static pages
- `document_noun` is `judgment`/`press summary`); `court_type` is `court`/`tribunal`. These should auto-extend in future.

## Jira

FCLC-2057


ds_judgements_public_ui/templates/includes/gtm_datalayer.jinja
```diff
@@ -0,0 +1,8 @@
+{# Page metadata for GTM / GA4. Only emit when a view has set gtm_data_layer. #}
+{% if gtm_data_layer %}
+  {{ gtm_data_layer | json_script("gtm-data-layer") }}
+  <script>
+    window.dataLayer = window.dataLayer || [];
+    window.dataLayer.push(JSON.parse(document.getElementById("gtm-data-layer").textContent));
```
jlhdxw, line 6: What happens when `document.getElementById("gtm-data-layer")` isn't found? Does this still work?

I'm not sure if it's possible for `{{ gtm_data_layer | json_script("gtm-data-layer") }}` to fail if `gtm_data_layer` for some reason has bad data?

judgments/utils/gtm_datalayer.py
```diff
@@ -0,0 +1,62 @@
...
+        return court_analytics(all_courts.get_by_param(CourtParam(param)))
+    except CourtNotFoundException:
+        return court_analytics_from_code(param.upper())
+
+
+def court_analytics_from_code(code: Union[str, CourtCode]) -> dict[str, str]:
+    if not code:
+        return {}
+    try:
+        return court_analytics(all_courts.get_by_code(CourtCode(str(code))))
```
jlhdxw: I'm worried this function will later get called multiple times and cause unexpected api calls - I wonder if we fetch the court in the view, then just use court_analytics function

judgments/views/browse.py
```diff
@@ -116,4 +117,12 @@ def get_context_data(self, **kwargs):
         context["feedback_survey_type"] = "browse"
         context["feedback_survey_court"] = court_query
 
+        if court_query:
+            context["gtm_data_layer"] = build_gtm_data_layer(
+                page_type="browse",
```
jlhdxw: I wonder if we should have a list of these page types somewhere - these should be defined by whoever is doing the analytics and it should be a set list. If these change, it messes up the analytics later.

ds_judgements_public_ui/templates/includes/gtm_datalayer.jinja
jacksonj04: I think if `gtm_data_layer` is invalid JSON then it'll serve a 500 instead - if it's _wrong_ JSON we should be testing elsewhere!

jlhdxw (approved):
LGTM

judgments/views/detail/detail_html.py
```diff
@@ -90,6 +96,17 @@ def build_context(request, document_uri):
     context["page_title"] = document.body.name  # TODO: Remove this from context
     context["pdf_uri"] = pdf.uri  # TODO: Remove this from context
 
+    try:
+        court = all_courts.get_by_code(document.body.court_and_jurisdiction_identifier_string)
+    except CourtNotFoundException:
+        court = None
+
```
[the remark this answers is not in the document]
jacksonj04: Documents don’t expose params, only codes. `court_and_jurisdiction_identifier_string` is a real API client property (it's a `CourtCode` specifically); we just haven’t needed it elsewhere in this repo before.

`body.court` alone isn’t enough either, because this isn't actually a `Court`.

PR #3514, opened 2026-07-30
Fclc 2205/add subjects to licence support email links

jlhdxw:
## Changes in this PR:

Adds subjects to licence support email links as well as tracking classes for analytics. 

I removed some views/templates that are no longer accessible as they had the raw email address in them.

https://national-archives.atlassian.net/browse/FCLC-2205


jacksonj04 (approved):
🚀

PR #3528, opened 2026-08-03
Present dates in Europe/London; keep storage UTC

jacksonj04:
- Set `TIME_ZONE = "Europe/London"` while keeping `USE_TZ = True` so DB/API interchange stays UTC
- Add `judgments.utils.timezones` helpers (`london_today`, `utc_now`, `as_utc_datetime`)
- Use London calendar day for year facets / court ended / recalculate end-year clamp
- Parse sitemap `lastmod` and Atom feed timestamps via `as_utc_datetime` (works with client 49 ISO strings with offsets)
- Format aware datetimes in Jinja `formatdate` in Europe/London; date-only values unchanged
- Clear remaining ruff DTZ findings that blocked the ruff Renovate PR

### DIFF
--- a/judgments/utils/timezones.py
+++ b/judgments/utils/timezones.py
@@ -1,0 +1,31 @@
+"""Timezone helpers: UTC for storage/interchange, Europe/London for presentation."""
+
+from __future__ import annotations
+
+from datetime import date, datetime, timezone
+from zoneinfo import ZoneInfo
+
+LONDON = ZoneInfo("Europe/London")
+UTC = timezone.utc
+
+
+def utc_now() -> datetime:
+    return datetime.now(tz=UTC)
+
+
+def london_today() -> date:
+    """UK calendar date (Europe/London), including BST when in effect."""
+    return datetime.now(tz=LONDON).date()
+
+
+def as_utc_datetime(value: datetime | str) -> datetime:
+    """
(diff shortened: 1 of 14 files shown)


judgments/utils/timezones.py
```diff
@@ -0,0 +1,31 @@
+"""Timezone helpers: UTC for storage/interchange, Europe/London for presentation."""
```
dancs-dev, line 1: **suggestion (non-blocking):** As a general suggestion, it might be worth making more use of [`django.utils.timezone`](https://docs.djangoproject.com/en/6.0/ref/utils/#module-django.utils.timezone).

jlhdxw (approved):

judgments/utils/timezones.py
```diff
@@ -0,0 +1,31 @@
...
+
+from __future__ import annotations
+
+from datetime import date, datetime, timezone
+from zoneinfo import ZoneInfo
+
+LONDON = ZoneInfo("Europe/London")
```
dancs-dev, line 8: **nitpick (non-blocking):** could this be `LOCAL`, pointing to `settings.TIME_ZONE`? Unlikely to matter now, but would make things easier if this project is ever used in other countries/timezones.

dancs-dev (approved):

PR #3587, opened 2026-08-12
feat: add about page for UKFTT-IAC

dancs-dev:
## Changes in this PR:

- Add about page for the new UKFTT-IAC tribunal

Subtask FCLC-2289 of FCLC-2252

## Screenshots of UI changes:

See e2e test screenshots.


ds_judgements_public_ui/templates/content/courts/ukftt/iac.jinja
```diff
@@ -0,0 +1,88 @@
+{% from "components/link.jinja" import link %}
```
anthonyhashemi, line 1: Not necessarily blocking for this PR, but looking at this, change, there seems to be a lot of similar code to the other court pages. We should be able to make a component (using a jinja macro) which takes in:

- `understanding`
- `transfer_method`
-`description`
-`about`

parameters to which we can pass those html blocks

anthonyhashemi (approved):

ds_judgements_public_ui/templates/content/courts/ukftt/iac.jinja
dancs-dev: I've created a Jira ticket to address this: https://national-archives.atlassian.net/browse/FCLC-2318

PR #3588, opened 2026-08-12
Fclc 2181/backfill jinja template tags

jlhdxw:
## Changes in this PR:

Refactoring how we set up the global jinja functions and filters. Backfilled a number of tests around those functions we've added recently where we didn't have them. 

https://national-archives.atlassian.net/browse/FCLC-2181


anthonyhashemi (approved):
sensible refactor and tests

PR #3595, opened 2026-08-13
Split document URI and identifier slug path converters

jacksonj04:
We're using a single `DocumentUriConverter` to handle both slug-based document URLs at `/`, and URI-based document URIs at `/id/`. Unfortunately, these two things aren't the same.

To fix this we now have a `DocumentUriConverter` for `/id/` which strictly matches what we consider to be a URI (without allowing `.`), and a new `DocumentIdentifierSlugConverter` for `/` which allows the full range of slugified identifier values.

This means that people trying to access invalid document URIs via `/id/` by including format extensions like `data.xml` will correctly receive a HTTP 404, instead of a HTTP 500 and us getting an annoying error message in Rollbar.

### DIFF
--- a/judgments/converters.py
+++ b/judgments/converters.py
@@ -71,6 +71,20 @@ class FileFormatConverter:
 
 
 class DocumentUriConverter:
+    """Path converter for document URIs under `/id/` (no full stops)."""
+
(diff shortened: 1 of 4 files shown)


dancs-dev (approved):
Looks good to me!

PR #3615, opened 2026-08-20
FCLC-2322 | spike only return legally significant judgments from search

jlhdxw:
## Changes in this PR:

We wanted to get confidence getting legally significant judgments with the recent "court_of_record" change was straight forward.

Result: yes, it is.

**This is just a spike, so don't merge as is - needs tests/UI etc**


jlhdxw:
Closing the spike as it proved how simple it is to do this.

PR #3619, opened 2026-08-21
FCLC-2279 | adjust cookie banner urls

jlhdxw:
## Changes in this PR:

Adjusts the cookie banner URLs

https://national-archives.atlassian.net/browse/FCLC-2279

### DIFF
--- a/ds_judgements_public_ui/static/js/src/cookies.js
+++ b/ds_judgements_public_ui/static/js/src/cookies.js
@@ -2,7 +2,7 @@ import { CookieBanner } from "@nationalarchives/frontend/nationalarchives/compon
 import Cookies from "@nationalarchives/cookies";
 
 document.addEventListener("DOMContentLoaded", () => {
-    const cookies = new Cookies();
+    new Cookies();
 
     const cookieBannerElement = document.querySelector(
         '[data-module="tna-cookie-banner"]',
--- a/ds_judgements_public_ui/templates/layouts/base.jinja
+++ b/ds_judgements_public_ui/templates/layouts/base.jinja
@@ -12,7 +12,7 @@
 {% from "components/footer.jinja" import footer, footer_row, footer_item, footer_link, footer_item_spacer with context %}
 {% from "components/cookie-banner/macro.html" import tnaCookieBanner %}
 <!DOCTYPE html>
-<html lang="en-gb">
+<html lang="en-gb" data-tna-cookies-domain="{{ cookie_domain }}">
   <head>
     <meta charset="utf-8" />
     {% block meta_description %}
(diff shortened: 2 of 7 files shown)


ds_judgements_public_ui/static/js/src/cookies.js
```diff
@@ -2,7 +2,7 @@ import { CookieBanner } from "@nationalarchives/frontend/nationalarchives/compon
 import Cookies from "@nationalarchives/cookies";
 
 document.addEventListener("DOMContentLoaded", () => {
-    const cookies = new Cookies();
+    new Cookies();
```
anthonyhashemi, line 5: constant was unused I guess?

ds_judgements_public_ui/templates/layouts/base.jinja
```diff
@@ -12,7 +12,7 @@
 {% from "components/footer.jinja" import footer, footer_row, footer_item, footer_link, footer_item_spacer with context %}
 {% from "components/cookie-banner/macro.html" import tnaCookieBanner %}
 <!DOCTYPE html>
-<html lang="en-gb">
+<html lang="en-gb" data-tna-cookies-domain="{{ cookie_domain }}">
```
anthonyhashemi, line 15: what is this new attribute for?

judgments/context_processors.py
```diff
@@ -2,10 +2,23 @@
 from urllib.parse import unquote
 
 from django.core.exceptions import SuspiciousOperation
+from django.http.request import split_domain_port
 
 from config.settings.base import env
 
 
+def cookie_domain_from_host(host):
```
anthonyhashemi, line 10: I had a concern that cookie_domain_from_host is a touch brittle, depending on host structure and is a little complicated at first glance due to breaking down domains into constituent parts, whilst acknolwegding different domain formats in particular for development purposes.

Alternatively, we could just set the cookie domain as a config value.

However, if these are the only hostnames we’ll ever serve from, then this seems reasonable and avoids unnecessary config.

anthonyhashemi (approved):

ds_judgements_public_ui/static/js/src/cookies.js
jlhdxw: Yep I realised while I was looking at the code - no need to set it to a variable.

ds_judgements_public_ui/templates/layouts/base.jinja
jlhdxw: This sets the cookie domain. Usually the library just figures it out based off your host, but it doesn't remove the subdomain by default (I thought it would) so we need to add this specifically to tell it our cookie domain is ".nationalarchives.gov.uk" rather than "caselaw.nationalarchives.gov.uk" otherwise the cookie settings on the main TNA website don't apply to caselaw.

judgments/context_processors.py
jlhdxw: We could do it with a config which would be pretty straightforward. Only issue would be locally if people use custom local domains for example they'd need to set that up, which isn't much of an issue.

It's unlikely tna domain will change, so setting this as a config is pretty low risk. I'm not sure how I go about doing that - I know the code side/docker side, but I'm not sure where our secrets/envs are held (github? AWS SM?) I can add a ticket to the backlog but I don't think I could action it fully

anthonyhashemi: Yeah I think it's fine, just was a mild concern initially but I agree that we are unlikely to change the domain

ds_judgements_public_ui/templates/layouts/base.jinja
anthonyhashemi: I see!

PR #3648, opened 2026-08-27
Fclc 2281/remove unavailable pagination options

jlhdxw:
## Changes in this PR:

Don't show previous/next controls on pagination if there is no previous/next page

https://national-archives.atlassian.net/browse/FCLC-2281


anthonyhashemi (commented):
Could be worth a test of the view itself on the last page to show that pagination is rendered correctly. Don't think an extra snapshot would be needed though

anthonyhashemi (approved):

PR #3649, opened 2026-08-27
FCLC-2306 | reduce squashing of pagination on lower breakpoints

jlhdxw:
## Changes in this PR:

Fixes the unnecessary squishing of the pagination on the lower breakpoints that is default with govuk pagination.

https://national-archives.atlassian.net/browse/FCLC-2306


ds_judgements_public_ui/sass/components/_pagination.scss
```diff
@@ -1,7 +1,13 @@
 .govuk-pagination {
```
anthonyhashemi, line 1: Not necessarily for this PR as it is a pattern we have been using across the board but I think we should discuss the way we are extending/modifying GDS components.


GDS suggest we should avoid overriding their classes since although it can work in the short term, future changes from GDS could result our UI breaking.

When making small modifications to components they suggest making use of:
our own [prefix](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#use-a-unique-prefix-for-component-names).

e.g. `fcl-pagination`

https://design-system.service.gov.uk/get-started/extending-and-modifying-components/

In AYR, we have been slowly moving away from the pattern shown here to the GDS recommendation

anthonyhashemi (approved):

PR #3747, opened 2026-09-11
Fclc 2297/improve sort per page controls

jlhdxw:
## Changes in this PR:

Adjusts the results sorting and per page controls so they change the results on change. Falls back to the original "with button" version when javascript is disabled (see screenshot below)

Previously the E2E tests weren't actually submitting the sorting/per page changes. Now it does it on click, it does so I needed to re-record some cassettes. 

https://national-archives.atlassian.net/browse/FCLC-2297

## Screenshots of UI changes:

Desktop:

image

Mobile:

image

image


dynamictulip (approved):
Looks good but one question - as we've removed "sort by" from the drop down text, do we need to add it anywhere else for screen readers?

PR #3748, opened 2026-09-11
FCLC-2393 | fix embedded search input width bug

jlhdxw:
## Changes in this PR:

When you view the embedded search input (for example on a single court page) on a breakpoint between mobile and desktop, it caused the page to have a horizontal scroll. This fixes that.

https://national-archives.atlassian.net/browse/FCLC-2393

### Before

image

### After

image

### DIFF
--- a/ds_judgements_public_ui/sass/components/_search_input.scss
+++ b/ds_judgements_public_ui/sass/components/_search_input.scss
@@ -88,7 +88,11 @@
     }
   }
 
-  @media (min-width: $grid-breakpoint-medium) {
+  @media (min-width: $grid-breakpoint-large) {
+    width: 38rem;
+  }
+
+  @media (min-width: $grid-breakpoint-extra-large) {
     width: 47rem;
   }
 }


ds_judgements_public_ui/sass/components/_search_input.scss
```diff
@@ -88,7 +88,11 @@
     }
   }
 
-  @media (min-width: $grid-breakpoint-medium) {
+  @media (min-width: $grid-breakpoint-large) {
+    width: 38rem;
```
anthonyhashemi, line 92: This fixes the immediate overflow issue, but it feels breakpoint-specific. Do we have the option for something alternative, e.g. by relying on container options like width: % or max-width, or parent flex layouts? Maybe it is not possible but just wondering

anthonyhashemi (approved):

ds_judgements_public_ui/sass/components/_search_input.scss
```diff
@@ -88,7 +88,11 @@
...
+  }
```
anthonyhashemi, line 93: Can we add some tests for these breakpoints - I guess just a snapshot with an appropriate viewport?

jlhdxw: Yep we should be able to add a specific breakpoint for this component. Ill get that added

ds_judgements_public_ui/sass/components/_search_input.scss
jlhdxw: Potentially - these fixed width's are because we have it set to a random width on the homepage, rather than having it set to a grid. We could adjust how this is rendered on the homepage where the container sets the width of the component, rather than the component itself.

I'll have a look at that and see if it's doable.

PR #3787, opened 2026-09-16
Fclc 2408/ensure cleanup vite process

jlhdxw:
## Changes in this PR:

I encountered an issue where django crashed and when I tried to run it again, the vite process was already running so a new one couldn't run due to port conflict.

This change makes it so we ensure cleanup of that sub process.

Also docker always stays running when you kill the process, which is annoying when you're swapping between PUI and EUI development, so the extra change here makes sure the docker containers are stopped when you stop fab run.

https://national-archives.atlassian.net/browse/FCLC-2408


dancs-dev (approved):
Looks good to me! The Docker containers running after fab run had been stopped has also been annoying me when working on EUI, so nice to see this also fixed.

PR #3807, opened 2026-09-18
Include judgments with dummy dates in search results

dancs-dev:
## Changes in this PR:

- Include judgments with dummy dates (01-01-1000) in search results
- Continue excluding judgments with dummy dates from the atom feed

[FCLC-2370](https://national-archives.atlassian.net/browse/FCLC-2370)

[FCLC-2370]: https://national-archives.atlassian.net/browse/FCLC-2370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/settings/base.py
```diff
@@ -320,4 +321,6 @@
 # This is the minimum year the site currently says it handles
 MINIMUM_WARNING_YEAR = 2003
 # Constant enforcing the minimum possible year a user can search for
-MINIMUM_ALLOWED_YEAR = 1085
+MINIMUM_ALLOWED_YEAR = 1000
+# Used to exclude judgments with dummy dates from the atom feed
+DUMMY_DATE = datetime.date(1000, 1, 1)
```
jlhdxw, line 326: I haven't been part of the conversation around this, so I don't know what we're using "dummy dates" for - would it either make sense to include in the comment here, or update the variable to reference what kind of judgments are being entered with these dates? Is it we sometimes add judgments with these dates and then update them later? Or is it for testing purposes and we don't want test data to show up on the atom feed, or something else?

judgments/tests/test_forms.py
```diff
@@ -39,10 +39,10 @@ def test_advanced_search_form_date_limit(self):
         data = {
             "from_date_0": "10",
             "from_date_1": "12",
-            "from_date_2": "1000",
+            "from_date_2": "999",
```
dynamictulip: this used to be 20 years behind the from date but now it is 10 months ahead of it. Was this intentional?

config/settings/base.py
dynamictulip: Dummy dates are the default date when a judgment has no clear date associated with it, either because the parser couldn't find it or because there just isn't one in the document. I agree that a comment defining this would be useful

dynamictulip (approved):
Looks good but for very minor tweaks

config/settings/base.py
dancs-dev: Thanks both - have clarified this in the comment

judgments/tests/test_forms.py
dancs-dev: Well spotted, have updated this

